PR #526, opened 2026-08-08
fix: make review tooling and hashes Windows-safe

xyh202131:
## Summary
- enforce LF text checkouts and explicit binary attributes so manifest/reference hashes are byte-stable on Windows
- replace unconditional `fcntl` usage with a tested POSIX/Windows worker lock
- emit POSIX relative links from the proposal HTML renderer on Windows
- make the gallery path assertion cross-platform and refresh generated submission data after recent merges

## Verification
- fresh Windows worktree: `216 passed, 97 subtests passed`
- standard reference snapshots: `i/lf w/lf attr/text=auto eol=lf`
- lock regression: exclusive acquisition, release/reacquisition, and release on queue failure
- `git diff --check`


xyh202131:
Platform-maintainer handoff: this PR intentionally changes repository infrastructure, so the participant-scope validator correctly rejects it for author `xyh202131`. The implementation itself was verified from a fresh Windows worktree at commit `03cde28a95bd11c54935b7fe90fd421974703f14`: **216 passed, 97 subtests passed**, and tracked standard snapshots are `i/lf w/lf`.

A maintainer can land the fix by cherry-picking `03cde28a95bd11c54935b7fe90fd421974703f14` onto a maintainer-owned branch (or re-opening the same diff from an account in `MAINTAINER_BYPASS_LOGINS`). No bypass is requested for the participant account.

xyh202131:
撤回此仓库级基础设施修改。后续仅维护 submissions/xyh202131/ 范围；如维护者认为其中修复有价值，可自行参考提交。

PR #570, opened 2026-08-08
fix: decode participant preflight output as UTF-8 on Windows

147228:
## Summary\n\nFixes #420.\n\n- Decode all participant preflight subprocess output as UTF-8 instead of the Windows locale.\n- Replace malformed bytes so a diagnostic command cannot crash the preflight report.\n- Add regression tests for non-ASCII subprocess output and the explicit decoding contract.\n\n## Validation\n\n- python3 -m unittest discover -s tests -p 'test_*participant*.py' -v (7 passed)\n- python3 -m py_compile scripts/participant_preflight.py tests/test_participant_preflight.py\n- git diff --check\n\nThis is intentionally scoped to the subprocess boundary described in #420; it does not alter manifest hashing or line-ending behavior.,

PR #616, opened 2026-08-08
fix: keep participant preflight UTF-8 safe on Windows

### LINKED ISSUE #420: [Windows] participant_preflight crashes when Git output is UTF-8 under GBK locale
## Affected tool

`scripts/participant_preflight.py` at upstream `92048eb`.

## Environment

- Windows 11 / PowerShell
- Python 3.13 with the default Simplified Chinese Windows locale (`gbk` / cp936)
- Repository workspace path contains CJK characters

## Reproduction

```powershell
python scripts\participant_preflight.py submissions\Mentat-Uran\jingzhang-civic-ai-relay --pr-author Mentat-Uran
```

## Observed

The subprocess reader thread raises:

```text
UnicodeDecodeError: 'gbk' codec can't decode byte 0xac ...
AttributeError: 'NoneType' object has no attribute 'strip'
```

`run()` uses `subprocess.run(..., text=True)` without an explicit encoding. Git emits UTF-8 paths/content, but Python decodes them with the Windows locale.

## Expected

The preflight should return its normal PASS/FAIL report on Windows regardless of the current ANSI code page.

## Confirmed workaround

```powershell
python -X utf8 scripts\participant_preflight.py submissions\Mentat-Uran\jingzhang-civic-ai-relay --pr-author Mentat-Uran
```

With UTF-8 mode the same package passes preflight (33 changed files, 8.8 MiB). A narrow fix could set `encoding=utf-8, errors=replace` in the subprocess helper, with a regression test using non-ASCII Git output.

CocoSgt:
## Summary

Closes #420.

- decode participant preflight subprocess output explicitly as UTF-8 with replacement for malformed bytes
- force PYTHONUTF8=1 and PYTHONIOENCODING=utf-8 through the child environment so self-check Python subprocess trees emit the same encoding
- add regressions for UTF-8 Chinese output, simulated legacy cp936 settings, malformed bytes, and subprocess argument contracts

This is the maintainer integration of #570 with an additional fix for the Python-child encoding mismatch found during independent review.

## Verification

- participant preflight encoding tests: 4 passed
- lightweight participant flow: 5 passed
- py_compile passed
- git diff --check passed

### DIFF
--- a/scripts/participant_preflight.py
+++ b/scripts/participant_preflight.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import subprocess
 import sys
 from pathlib import Path
@@ -17,6 +18,9 @@ LARGE_PACKAGE_WARNING = 200 * 1024 * 1024
 
 
 def run(command: list[str], cwd: Path) -> subprocess.CompletedProcess[str]:
+    environment = os.environ.copy()
+    environment["PYTHONUTF8"] = "1"
+    environment["PYTHONIOENCODING"] = "utf-8"
     return subprocess.run(
         command,
         cwd=cwd,
@@ -24,6 +28,7 @@ def run(command: list[str], cwd: Path) -> subprocess.CompletedProcess[str]:
         text=True,
         encoding="utf-8",
         errors="replace",
+        env=environment,
         check=False,
     )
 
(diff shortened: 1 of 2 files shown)


CocoSgt:
Trusted-base validation passed after queued execution. Local verification also covered 4 encoding regressions and 5 participant-flow tests. Proceeding with maintainer merge.

PR #653, opened 2026-08-08
fix: keep the submission self-check UTF-8 safe under non-UTF-8 locales

CaoChen13:
## Problem

`scripts/participant_preflight.py` was made UTF-8 safe in #616, but `scripts/self_check_submission.py` — the check `SKILL.md` asks every participant to run — still depends on the ambient locale encoding in two places:

1. `run_json_command()` launches the validator subprocesses without forcing UTF-8, so each child inherits a locale that may be unable to encode its own output;
2. `main()` prints the report through `sys.stdout` with the locale encoding.

This matters because the report legitimately carries Chinese text: `scripts/validate_submission.py` embeds required section names in messages such as ``section `设计依据与资料清单` is too thin``, which any `language: "zh"` package hits while iterating.

Under a locale that cannot encode those characters — `LC_ALL=C`, common in minimal containers and CI images where agents run — `validate_local_submission.py` dies inside `print(json.dumps(..., ensure_ascii=False))` and the self-check reports the Python traceback **as if it were a validation finding**:

```
- Fix deterministic validation error: Traceback (most recent call last):
    File ".../scripts/validate_local_submission.py", line 78, in main
      print(json.dumps(report.to_dict(), ensure_ascii=False, indent=2))
  UnicodeEncodeError: 'ascii' codec can't encode characters in position 644-652: ordinal not in range(128)
```

A contributor then debugs a validation error that does not exist, while the real findings stay hidden.

## Reproduction

```bash
LC_ALL=C PYTHONUTF8=0 PYTHONCOERCECLOCALE=0 \
  python3 scripts/self_check_submission.py submissions/<login>/<slug> --pr-author <login>
```

with any `language: "zh"` package that has at least one thin required section. Before this change the run ends in `UnicodeEncodeError`; after it, the same command prints the real finding:

```
- Fix deterministic validation error: .../proposal.md: section `设计依据与资料清单` is too thin; ...
```

## Fix

- launch child validators with `PYTHONUTF8=1`, `PYTHONIOENCODING=utf-8`, and explicit `encoding="utf-8", errors="replace"`, matching `participant_preflight.py`;
- reconfigure the entry point's own `stdout`/`stderr` to UTF-8 with `errors="replace"` so the report always prints.

Both are needed: fixing only the child moves the same crash into the parent's `print()`.

## Validation

- `python3 -m unittest tests.test_agent_scaffold_and_self_check.SelfCheckLocaleSafetyTests` — 2 new regression tests pass, and both fail when the fix is reverted;
- full suite on Windows (`python -m unittest discover -s tests -t .`): identical results before and after this change, so no behavior regression;
- manual before/after runs under `LC_ALL=C` and under a UTF-8 locale, as shown above.

No behavior change on systems whose locale already encodes the report.


anselasimov-web (approved):
Reviewed exact head 20f2fbe690c88850d89962e8f263b2fbd3863352. The self-check now forces UTF-8 for child validators and configures its own stdout/stderr with replacement semantics, so Chinese diagnostics remain printable under a non-UTF-8 locale without changing validation results. The regression asserts UTF-8 subprocess settings and covers non-UTF output where supported; trusted submission-validation is SUCCESS. No blocking issue found.

CocoSgt (changes requested):
Exact-head review on 20f2fbe690c88850d89962e8f263b2fbd3863352 against current main d8a3fcbc81d66e8155ce5dbe56ae8e9a24804fe7 found that this needs a current-main rebase before it is safe to merge.

Both changed files now conflict. Main has already independently adopted and extended the child-process UTF-8 environment/decoder path, together with newer JSON fail-closed handling and CLI/self-check behavior. Resolving by taking this old branch version would regress those later safeguards and arguments.

The remaining defect is real and narrower: current main still lacks parent stdout/stderr reconfiguration. On macOS Python 3.9.6, LC_ALL=C PYTHONUTF8=0 PYTHONCOERCECLOCALE=0 reports sys.stdout.encoding=ascii and printing 设计依据与资料清单 raises UnicodeEncodeError; the PR force_utf8_output path fixes that, including redirected cp1252 stream simulation.

Please rebase onto current main, drop the now-duplicate run_json_command hunk, retain only the parent force_utf8_output integration, and port its locale/redirected-stream regressions into the current SelfCheckEncodingTests without replacing current main logic. Then rerun that focused module, py_compile, diff-check, and a new exact-head required check. Current evidence: PR-head locale tests 2/2 pass; current-main encoding tests 4/4 pass; both sides py_compile/diff-check pass, but merge-tree is DIRTY. No merge attempted.

CocoSgt:
I applied the requested current-main collaboration update as a fast-forward merge commit (no force push). Exact head is now f135abb8c0d931799f069c75b03a664876189720 with current main d8a3fcbc81d66e8155ce5dbe56ae8e9a24804fe7 as its second parent.

The resulting PR diff is again limited to two files: it preserves current main child-process UTF-8 decoding, JSON fail-closed behavior, newer CLI options, and self-check persistence; the only production addition is parent stdout/stderr UTF-8 reconfiguration before argument parsing/output. The locale regression is now part of the current SelfCheckEncodingTests instead of duplicating the already-covered child behavior.

Verification on macOS Python 3.9.6 with review dependencies:
- tests.test_agent_scaffold_and_self_check: 35/35 PASS
- SelfCheckEncodingTests: 5/5 PASS
- py_compile: PASS
- diff against current main and git diff --check: PASS
- LC_ALL=C, PYTHONUTF8=0, PYTHONCOERCECLOCALE=0 reproduces ascii stdout/UnicodeEncodeError on current main; the revised exact head prints the Chinese diagnostic successfully.

Waiting for a new exact-head required check before re-review or merge.

CocoSgt (approved):
Exact-head re-review on `f135abb8c0d931799f069c75b03a664876189720` against current main passes. The update keeps only parent stdout/stderr UTF-8 reconfiguration while retaining current main child-process decoding and fail-closed behavior. Focused `tests.test_agent_scaffold_and_self_check` passed 35/35; `py_compile`, `git diff --check`, a clean latest-main merge-tree, and required submission-validation all pass. Approved for the narrow locale-safety fix.

PR #681, opened 2026-08-08
Raise trusted validator download cap to 10 MiB

wakenmeng:
Raises the trusted PR validator's per-file download cap from 6 MiB to 10 MiB. Adds exact boundary tests confirming that 10 MiB is accepted and 10 MiB + 1 byte is rejected.\n\nThis unblocks submission PR #639, whose largest artifact is about 6.15 MiB.\n\nTests: `python3 -m unittest tests.test_submission_workflow`

### DIFF
--- a/scripts/github_pr_validation.py
+++ b/scripts/github_pr_validation.py
@@ -27,6 +27,7 @@ from validate_submission import ValidationReport, format_report, validate_submis
 COMMENT_MARKER = "<!-- haidian-submission-validation -->"
 API_ROOT = "https://api.github.com"
 MAX_API_ATTEMPTS = 4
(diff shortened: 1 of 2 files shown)


147228:
Review on exact head 51172ed5: this is a focused unblock for #639. The cap is centralized as MAX_DOWNLOAD_BYTES, download_content still reads max+1 so the 10 MiB + 1 boundary remains rejected, and the added tests cover both boundaries. I do not see an authorization or path-scope regression in this diff. The trusted validation run 31268781659 is still pending; please use that exact-head result before merging, then re-run #639 after its PDFs and manifest hashes are updated.

PR #747, opened 2026-08-08
fix: clarify advisory score readiness

147228:
## Summary

- Mark `score_submission.py` JSON output as advisory-only and state that formal readiness is not assessed.
- Make terminal output and `--strict` help explain that a zero exit code is not a formal admission result.
- Document the required next command and cover the output contract with unit tests.

## Why

A proposal can receive an 8/8 advisory score while still failing deterministic or professional evidence validation. This change keeps the formal gates unchanged and prevents interpreting the lightweight score as a `formal-review-ready` result.

## Verification

- `python3 -m unittest tests/test_score_submission.py`
- `python3 -m py_compile scripts/score_submission.py`
- Replayed the updated advisory command against #745: it reports 8/8 while explicitly stating that formal readiness is not assessed.
- `git diff --check`

PR #751, opened 2026-08-08
fix: score v2 package sources by default

147228:
## What changed

- Automatically use a v2 submission's package-local `sources.json` when no index is specified.
- Preserve the repository-wide public-source index for legacy submissions.
- Preserve `--sources-index` as an explicit override.

## Why

The documented default score command previously read only the legacy global index. A valid v2 submission can keep its complete machine-audit source inventory in its own package, so its `[source:...]` references were reported as unmatched even when the package source checks passed.

## Validation

- `python3 -m unittest tests/test_score_submission.py tests/test_submission_workflow.py` (92 tests)
- `git diff --check`
- Exact-head metadata check against #750 `c197778d40df3e32edafb48aa57aeb152013cd66` selects its package `sources.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,10 @@ python3 scripts/render_proposal_html.py submissions/<your-github-login>/<proposa
 python3 scripts/score_submission.py submissions/<your-github-login>/<proposal-slug>/proposal.md
 ```
 
+对 `proposal_format_version: "2"` 的投稿，脚本会自动读取同一投稿目录的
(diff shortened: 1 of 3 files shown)


147228:
Closing as duplicate of #747. I compared the current heads: #747 is a strict superset of this draft—it preserves the advisory/formal-readiness boundary, combines repository and package-local source indexes, and also matches registry_source_id with deduplication. #751 has no unique behavior worth making maintainers review twice; please use #747 as the single resolution path.

PR #585, opened 2026-08-08
fix: add manifest hash refresh utility

147228:
## Problem

The repository had no safe, reviewable freshness workflow for a formal package after listed artifacts changed. The merged bilingual backfill also exposed the consequence at repository scale: a read-only audit of `main@2c6ae754` found 133 stale artifacts across 133 of 262 packages.

## Fix

- Add `scripts/refresh_manifest_hashes.py` to validate paths and recompute every declared package hash atomically, preserving file mode and rejecting duplicates, missing files, path escapes, and symlinks.
- Keep `--check` read-only.
- When declared artifact bytes actually change, explicitly set `manifest.validation_claim.self_checked` to `false`. A hash rewrite is not evidence that the four review gates were rerun.
- Add `self_check_submission.py --record-pass`: only after deterministic, spatial, visual, and professional gates all pass does it record current self-check evidence, refresh its own declared hash, and restore `self_checked: true`.
- Make gallery generation treat explicit `self_checked: false` as `needs_revision`, even if an old `self_check.json` claimed readiness.
- Add `--all --repo-root . --json` as a repository-wide read-only audit. It never mutates packages; maintainers must remediate stale packages in a trusted batch, then rerun all gates and `--record-pass` per package.

## Validation

- `python3 -m unittest tests.test_refresh_manifest_hashes -v` — 10 passed.
- `python3 -m pytest -q tests/test_submission_workflow.py tests/test_agent_scaffold_and_self_check.py tests/test_export_review_packet.py` — 110 passed.
- Isolated exact-package end-to-end check: refresh invalidates freshness, the four gates pass, `--record-pass` restores only fresh evidence, and a subsequent `--check` reports current hashes.
- `python3 scripts/refresh_manifest_hashes.py --all --repo-root . --json` — 262 packages scanned; 133 stale packages detected without writing.
- `git diff --check` and `py_compile` passed.

The current remote workflow failure is installation-level GitHub API rate limiting at PR-file enumeration, not a package validation result. This PR does not silently bulk-rewrite other authors’ artifacts or claim that the existing 133 stale packages are ready.

PR #765, opened 2026-08-08
fix: refresh manifest hashes for ready packages

147228:
## Summary

- add an explicit --refresh-manifest mode for ready_for_review packages
- reject incomplete or unsafe manifest paths before writing any refreshed hash
- document the post-finalization workflow and cover it with regression tests

## Verification

- python3 -m unittest discover -s tests -p test_agent_scaffold_and_self_check.py -v
- python3 -m pytest -q  # 247 passed
- end-to-end regression against exact failed #375 head 0ff2a91: refreshed 29 declared hashes, then full self-check returned ok=true and can_enter_formal_review=true; remaining notices were non-blocking legacy or review warnings

This addresses the reproducible failure mode where a legitimate follow-up edit leaves every declared manifest SHA-256 stale, while the normal finalization command correctly refuses non-scaffold packages.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -64,6 +64,8 @@
 python3 -m pip install -r requirements-review.txt
 python3 scripts/render_proposal_html.py submissions/<my-github-login>/<proposal-slug>
 python3 scripts/finalize_submission.py submissions/<my-github-login>/<proposal-slug>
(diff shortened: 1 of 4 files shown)


147228:
Closing as a duplicate of #585. #585 already provides the same manifest-hash refresh capability with a dedicated utility, a check-only mode, atomic writes, symlink containment, and file-mode preservation. Keeping one stronger implementation avoids parallel review and conflicting contributor instructions. The branch is intentionally retained.

PR #783, opened 2026-08-09
fix: parse visual metric attributes in any order

147228:
## What changed

- Replace the order-sensitive regular expression in `visual_review.py` with standard-library HTML attribute parsing.
- Accept a metric declaration whenever the same element has both `data-metric` and `data-value`, regardless of attribute order.
- Add a regression test for `data-value` before `data-metric`.

## Why

HTML attribute order is semantically irrelevant. The prior parser recognized only `data-metric="..." data-value="..."`; a valid reversed form was reported as a missing required metric and blocked an otherwise correct submission.

## Impact

Visual review remains static and does not execute contributor JavaScript. Metric source, status, numeric comparison, and required-metric checks are unchanged; this only removes a false failure mode.

## Validation

- `python3 -m unittest tests.test_visual_review -q` — 6 passed
- `python3 -m pytest -q` — 268 passed
- `python3 -m py_compile scripts/visual_review.py`
- `git diff --check`

### DIFF
--- a/scripts/visual_review.py
+++ b/scripts/visual_review.py
@@ -13,6 +13,7 @@ import html
 import json
 import re
 from dataclasses import dataclass, field
+from html.parser import HTMLParser
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


147228:
Exact-head audit for 42e5a947:

- HTMLParser change accepts `data-metric` and `data-value` in either order while keeping metric source/status/value checks unchanged.
- Additional hardening rejects non-finite values such as `NaN`; otherwise `abs(nan - expected) > tolerance` could evaluate false and bypass consistency validation.
- `tests.test_visual_review`: 7 passed; full pytest: 269 passed; py_compile and git diff --check: PASS.

CocoSgt (approved):
Independent exact-head review: standard-library HTMLParser keeps metric attributes scoped to the same element and order-independent; non-finite values are rejected instead of bypassing float comparison. Focused visual tests 7/7 and broader agent/submission regressions 107/107 passed after materializing the scenario registry; py_compile and diff check passed. No participant scope, trusted-base, source registry, or scoring policy change.

PR #755, opened 2026-08-08
Support alternate preflight push remotes

cryanskl:
## Summary

- add `--push-remote` to `participant_preflight.py`, defaulting to `origin`
- use the selected remote for both `git push --dry-run` and the suggested upload command
- validate that the selected remote exists and report its URL in JSON output
- document the legacy `origin=canonical` / `fork=writable` workspace layout
- add focused CLI and command-construction regression tests

## Why

The current preflight hard-codes `origin` for its push-access check. Existing participant workspaces may legitimately keep the canonical repository as `origin` and the writable participant fork as `fork`. Those workspaces pass every package check but receive a false upload blocker until the user renames remotes.

An explicit remote is deterministic and avoids probing or guessing which configured remote should receive the branch. The default remains `origin`, so workspaces created by the current bootstrap helper are unchanged.

## Validation

- `python3 -m unittest tests.test_participant_preflight` — 6 passed
- `python3 -m unittest tests.test_participant_preflight tests.test_lightweight_participant_flow tests.test_site_package_contract` — 21 passed
- real `git push --dry-run` through a temporary alternate remote — PASS; no remote branch created
- `python3 -m unittest discover -s tests` — 236/241 passed
- the five failures are generated-gallery freshness/count checks and reproduce unchanged in a clean `upstream/main` worktree
- `python3 -m py_compile scripts/participant_preflight.py tests/test_participant_preflight.py` — PASS
- `git diff --check` — PASS

## Scope

This changes only the participant preflight, its focused tests, and the lightweight-workspace documentation. It does not modify any submission package or generated gallery data.

PR #786, opened 2026-08-09
fix: quote participant preflight push commands

147228:
## What changed

- Bring the explicit `--push-remote` participant-preflight workflow to current `main`.
- Keep the selected remote consistent across dry-run validation, JSON output, and suggested upload commands.
- Render every displayed `git push` command with `shlex.join`, so remote and ref names containing shell punctuation remain one literal argument when copied.
- Add regression coverage for the real-world `fork;echo-unsafe` remote-name case and document legacy `origin`/writable-`fork` layouts.

## Root cause

PR #755 correctly identified that workspaces may use a canonical `origin` plus a writable fork remote. Its implementation constructed user-facing commands with string interpolation. Git accepts remote names containing shell metacharacters; the subprocess dry-run was safe because it used an argv list, but copied output such as `git push -u fork;echo-unsafe HEAD` was not equivalent to the validated command.

This PR preserves the explicit-remote behavior while making the display/copy path shell-safe.

## Validation

- `python3 -m unittest tests.test_participant_preflight -q` — 7 passed
- `python3 -m py_compile scripts/participant_preflight.py tests/test_participant_preflight.py`
- `git diff --check`

Related: #755

### DIFF
--- a/scripts/participant_preflight.py
+++ b/scripts/participant_preflight.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import argparse
 import json
 import os
+import shlex
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


anselasimov-web (changes requested):
Request changes at exact head 3a617026efef7dcb2d6378c3b147359623557838. This improves shell quoting for displayed commands, but the selected push-remote is still passed directly to git remote get-url and git push. A value beginning with a dash can be parsed as a Git option, so quoting the printed command does not close the option-injection boundary. Please validate push-remote before any Git call, preferably with a documented remote-name pattern that rejects leading dashes and control characters, and add a regression for the rejected value.

PR #792, opened 2026-08-09
fix: verify formal source claims against registry

147228:
## What changed

- Validates explicit formal-use claims in submission `sources.json` against `data/source_registry.json`.
- Recognizes the supported self-assertion forms `type: "formal_usable"`, `formal_usable: true`, `usable_for_formal: "yes"`/`true`, and `usable_for: "formal"`.
- Requires `source_registry_id` to resolve to a source with `review_status: "approved"` and `usable_for_formal: "yes"`.
- Keeps v1 packages compatible with visible warnings; v2 packages block unsupported formal-use claims.

## Why

`sources.json` previously accepted contributor-supplied formal-use labels without proving that a maintainer had reviewed the cited material. This let a public URL be presented as formal evidence without a registry decision.

## Impact

New v2 packages must map an explicitly formal source to the approved registry entry. Ordinary source categories such as `official_public` remain lightweight citations and do not trigger the gate. A registry mapping is an approval-state check, not a substitute for asset-level provenance or claim-scope review.

## Validation

- `python3 -m pytest -q tests/test_submission_workflow.py tests/test_data_registry.py tests/test_source_registry_draft.py tests/test_processed_data.py` — 105 passed
- `python3 scripts/validate_data_registry.py` — PASS (6 sources)
- `python3 -m py_compile scripts/validate_submission.py`
- `git diff --check`
- The full checkout-wide suite has 256 passing tests; 16 failures are sparse-checkout fixture absences (`.github/`, `skills/`, `examples/`, `sources/public-sources.json`, and historical submissions), not failures in this diff.

### DIFF
--- a/docs/data-workflow.md
+++ b/docs/data-workflow.md
@@ -26,6 +26,8 @@
 6. 在 `proposal.md` 中用 `[source:...]` 标签解释资料如何支撑文字、图层、指标和图纸。
 7. 对不能确认的资料，写入 `assumptions.json`、`self_check.json` 和 `proposal.md` 的风险/缺资料章节。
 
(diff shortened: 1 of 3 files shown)


anselasimov-web (approved):
Approved at exact head 8fdc365e17daadd66dd9de56aabea9972645e2e8. Explicit formal-use claims now require a source_registry_id resolving to an approved registry entry with usable_for_formal=yes, while legacy v1 packages receive visible warnings. Missing or invalid registry data fails closed for v2, and the focused registry, workflow, and compilation checks are documented as passing.

PR #793, opened 2026-08-09
fix: require explicit text-free neutral figures

147228:
## Summary

The v2 bilingual gate treated any figure declared as `language: "neutral"` as text-free. A package could therefore place readable English or Chinese labels in a single figure and bypass the required language counterpart check. PR #790 reproduces this with five figures that visibly contain English labels while declaring `neutral`.

This patch makes the declaration explicit and keeps the same migration boundary:

- add optional manifest item field `text_free`;
- only `language: "neutral", text_free: true` may skip figure counterpart validation;
- unverified `neutral` figures now fail the strict v2 contract and require an explicit `.zh`/`.en` counterpart;
- `finalize_submission.py` applies the same rule;
- update the formal submission docs and add regression tests.

`text_free: true` remains a contributor declaration and still requires human visual review; this change removes the silent validator bypass without pretending that OCR is a substitute for review.

## Verification

- `python3 -m pytest -q` — 270 passed
- `python3 -m py_compile scripts/validate_submission.py scripts/finalize_submission.py`
- `git diff --check`
- Replayed the exact PR #790 head (`866b54467fd743fa0466398de3e46dff392c5068`) against the patched validator: it now fails on all five neutral figures and reports the missing English counterparts.

### DIFF
--- a/brief/site-package/schemas/manifest.schema.json
+++ b/brief/site-package/schemas/manifest.schema.json
@@ -128,7 +128,7 @@
           "language": {
             "type": "string",
             "enum": ["zh", "en", "neutral"],
-            "description": "Language of a human-facing display artifact. Language-neutral assets may be shared by both versions."
+            "description": "Language of a human-facing display artifact. Language-neutral assets may be shared by both versions only when text_free is true."
           },
           "text_free": {
             "type": "boolean",
--- a/brief/site-package/schemas/manifest.schema.json
+++ b/brief/site-package/schemas/manifest.schema.json
@@ -130,6 +130,10 @@
             "enum": ["zh", "en", "neutral"],
             "description": "Language of a human-facing display artifact. Language-neutral assets may be shared by both versions."
           },
+          "text_free": {
+            "type": "boolean",
+            "description": "For a language-neutral figure, set true only when the asset contains no human-readable text and may be shared by both language versions."
+          },
           "translation_of": {
             "type": "string",
             "pattern": "^[A-Za-z0-9_./-]+$",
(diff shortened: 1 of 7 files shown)


147228:
Concrete motivating example: current draft #794 marks five text-bearing, English-labeled figures as language=neutral. On current main that bypasses the v2 display-counterpart check silently. This PR makes the bypass visible and failing unless the contributor explicitly attests text_free=true; it does not pretend to perform OCR or certify that an attestation is true. Reviewers still retain the visual-content judgment, while the manifest now carries an auditable contract instead of a silent neutral label.

147228:
Pushed follow-up commit 3007afa37e03ec3fdf8048539b9563c3185b3040.

This closes a manifest-scope bypass in the v2 neutral-figure guard: every primary assets/figures/* item in a strict bilingual manifest is now checked, even when the figure is only consumed by report/visual HTML and is not referenced from proposal.md. Such a figure must explicitly declare text_free: true or provide the required zh/en counterpart.

Regression coverage now includes an unreferenced neutral manifest figure. Local full suite: 272 tests passed. Trusted submission-validation for exact head 3007afa37e03ec3fdf8048539b9563c3185b3040: PASS, run 31288053033, job 93180365000.

147228:
Pushed follow-up commit 7afe124aaa5757807651fffe685db2c76238902d.

The same manifest-complete scope is now enforced by finalize_submission.py, not only by validate_submission.py. Finalization enumerates every primary assets/figures/* item in the manifest, including report/visual-only figures, and requires either explicit text_free=true or the configured bilingual counterpart.

Regression coverage now exercises finalization with an unreferenced manifest figure. Local full suite: 273 tests passed; py_compile and diff-check passed. Trusted validation for exact head 7afe124aaa5757807651fffe685db2c76238902d: PASS, run 31288229126, job 93180812746.

anselasimov-web (approved):
Approved at exact head 7afe124aaa5757807651fffe685db2c76238902d. Neutral figure declarations no longer bypass bilingual review unless text_free is explicitly true; the validator, finalizer, schema, docs, and tests share the same rule, and replaying the referenced PR exposes missing counterparts as intended. Official validation is PASS.

PR #800, opened 2026-08-09
fix: verify readable sources for known metrics

147228:
## What changed

- For every `known` metric, verify file-like `source_files` entries resolve to a real file within the repository.
- Resolve package-relative and repository-relative paths, including GeoJSON feature anchors such as `geometry/key_areas.geojson#KEY-001`.
- Reject source references that escape the repository.
- Keep opaque identifiers such as a future operational record name outside this file-existence rule.

## Why

`metrics.json` already requires `source_files`, but validation previously checked only their shape. A known value could therefore cite a missing input and appear reproducible while no reviewer could recompute it.

## Validation

- `python3 -m pytest -q tests/test_submission_workflow.py tests/test_agent_scaffold_and_self_check.py` — 109 passed
- `python3 -m pytest -q` — 271 passed
- `python3 -m py_compile scripts/validate_submission.py tests/test_submission_workflow.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -939,10 +939,13 @@ def validate_geojson_file(
     return len(features)
 
 
-def validate_metrics_file(report: ValidationReport, path: Path, display_path: str) -> None:
+def validate_metrics_file(
+    report: ValidationReport, repo_root: Path, path: Path, display_path: str
+) -> None:
     data = load_json_file(report, path, display_path)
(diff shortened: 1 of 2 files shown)


147228:
关闭为重复草案：#801 已提供更完整的共享路径、片段解析、文档和回归覆盖。已在 #801 提出将 ready_for_review 同样设为来源可读性的硬门，后续以该 PR 为准。

PR #806, opened 2026-08-09
feat: record passing submission self-checks

147228:
## Summary

- add `self_check_submission.py --record-pass`, which writes `validation_claim.self_checked=true` only after a formal package passes deterministic validation, spatial review, visual packaging, and professional-evidence review;
- record a concise PASS summary and remove any manifest self-hash to avoid an unsatisfiable checksum;
- document the supported workflow in the formal guide and repository submission skill, with regression tests for success and no-write-on-failure behavior.

## Motivation

Recent ready-for-review packages have exposed a mismatch between hand-edited `validation_claim.self_checked` values and their declared check state. Existing ready packages remain compatible: this PR does not retroactively fail `false` claims. It provides the auditable, repeatable write path for future submissions and strict v2 readiness workflows.

## Validation

- `python3 -m py_compile scripts/self_check_submission.py`
- `python3 -m unittest tests.test_agent_scaffold_and_self_check` — 20 passed
- `python3 -m unittest discover -s tests` — 271 passed
- `git diff --check`

### DIFF
--- a/brief/site-package/schemas/manifest.schema.json
+++ b/brief/site-package/schemas/manifest.schema.json
@@ -149,6 +149,10 @@
         "self_checked": {
           "type": "boolean"
         },
+        "self_check_summary": {
+          "type": "string",
+          "description": "Written by self_check_submission.py --record-pass after all required local gates pass."
+        },
         "known_blockers": {
           "type": "array",
           "items": {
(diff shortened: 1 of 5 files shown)


147228:
Superseded by #807. #806 had the same intended workflow writeback, but its --record-pass path could not pass the new strict v2 ready/self_checked gate. #807 includes the pending-check transaction, strict validator enforcement, and aligned contributor docs.

PR #811, opened 2026-08-09
fix: ignore denied sensitive-source disclosures

147228:
## Summary

- Avoid a false sensitive-source warning when a submission explicitly says it does not use non-public or sensitive source material.
- Keep the warning for affirmative use, double-negative wording, causal or contrasting clauses, and statements such as `不使用除企业内部数据以外的资料` that actually retain internal data.
- Apply the same interpretation to proposal, changelog, risk, and spatial metadata checks.

## Root cause

`SOFT_RISK_PATTERNS` used a bare keyword match. PR #809's public-data disclosure — `方案未使用任何非公开地图、企业内部数据…` — therefore received the same maintainer-review warning as an affirmative claim to use internal data.

## Validation

- `python3 -m unittest discover -s tests` — 272 passed
- `python3 -m py_compile scripts/validate_submission.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -281,10 +281,85 @@ HARD_RISK_PATTERNS = [
     (re.compile(r"(已获|已经获得).{0,12}(政府|官方).{0,12}(批准|背书|认可)"), "疑似伪造官方批准或背书"),
 ]
 
+SENSITIVE_SOURCE_PATTERN = re.compile(
+    r"(内部资料|内部数据|涉密|保密图件|非公开空间数据|未公开图件|绝密|机密)"
+)
+SENSITIVE_SOURCE_DENIAL_RE = re.compile(
+    r"(?:未使用|没有使用|并未使用|从未使用|不使用|未采用|没有采用|并未采用|不采用|"
+    r"未引用|没有引用|并未引用|不引用|未接触|没有接触|并未接触|不接触|"
+    r"未访问|没有访问|并未访问|不访问|未获取|没有获取|并未获取|不获取|"
+    r"未收集|没有收集|并未收集|不收集|未持有|没有持有|并未持有|不持有|"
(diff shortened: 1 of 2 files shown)


147228:
Follow-up pushed in commit d6f31bed.

The prior matcher could suppress a later sensitive-source term when the denial applied only to an unrelated public-source clause, for example `未使用公开资料，涉密测绘成果作为底图` or `不使用公开资料，企业内部数据用于校准指标`. The fix treats a comma-delimited clause as a new scope unless the denial explicitly quantifies a compound list with `任何`, and also recognizes bare `而` as a semantic pivot.

Added a regression covering the three unrelated-public-source disclosures while retaining the explicit-denial, double-negative, and causal-clause cases. Validation: `python3 -m unittest discover -s tests` — 274 passed; `py_compile` and `git diff --check` passed.

147228:
Current-main rebase and verification: exact base is acd6a738f67b04bd0f2ac5a028f3329de6992e90 and exact head is ecd87a8924c83ed1b30413d8cb175303ea5720be. The focused submission/preflight suite ran 112 tests and all passed; touched Python files compile and git diff --check pass. The false-positive fix remains clause-scoped: explicit denials of sensitive/non-public sources no longer warn, while affirmative, double-negative, causal, and contrast clauses remain reviewable. Trusted run 31310174143 is pending; no CI success is inferred.

147228:
Rebuilt this PR onto canonical main head 83b4c6317d79f7894f8f804546fdf90a067a15d4 and force-pushed with lease. New exact head: babe6f0cf89f702f5a55be3b77bba0479d200bc8. The fix remains limited to validate_submission.py plus regression coverage: denied sensitive-source wording is no longer treated as an affirmative disclosure, while affirmative, double-negative, causal, and contrast forms remain covered. Exact local verification on that head: 112 tests passed, Python compilation passed, and git diff --check passed. Trusted GitHub validation is still pending in the repository queue.

PR #824, opened 2026-08-09
fix: hash manifests as Git blobs

147228:
## Summary

- Hash manifest entries from the bytes Git will actually commit, using an isolated temporary index so the contributor's real index is untouched.
- Use the same Git-blob view during local manifest validation; standalone directories without Git retain the existing file-byte behavior.
- Add regression coverage for `core.autocrlf=true`, including finalizer, validator, and index non-mutation.

## Root cause

Several participant packages generated manifest digests from CRLF working-copy bytes, while Git committed LF blobs. Local checks could pass but trusted validation then rejected many artifacts. This makes local finalization and validation use the same bytes that trusted CI receives.

## Validation

- `python3 -m unittest tests.test_submission_workflow -v` — 89 passed
- `python3 -m unittest tests.test_agent_scaffold_and_self_check tests.test_git_blob_hashes -v` — 21 passed
- `python3 -m py_compile scripts/git_blob_hashes.py scripts/finalize_submission.py scripts/validate_submission.py`
- `git diff --check`

### DIFF
--- a/scripts/finalize_submission.py
+++ b/scripts/finalize_submission.py
@@ -8,6 +8,7 @@ import hashlib
 import json
 from pathlib import Path
 
+from git_blob_hashes import git_blob_sha256
 from validate_submission import (
     DISPLAY_BASE_FILES,
     is_empty_pdf,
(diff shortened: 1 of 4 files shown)


147228:
Local verification on exact head `024094290825370d39fe4171775b3f9041510f28`:

- 110 targeted tests passed (`test_git_blob_hashes`, `test_submission_workflow`, `test_agent_scaffold_and_self_check`)
- `py_compile` passed for the new helper, finalizer and validator
- `git diff --check` passed

The full repository discovery run in this isolated worktree cannot be treated as a code failure because this checkout lacks generated/source/gallery data files; the targeted tests cover the temporary-index isolation, CRLF normalization, finalizer digest path and validator digest path directly. Please review the fallback behavior when no usable Git HEAD is available, and whether the trusted validator should require Git-blob mode rather than silently falling back to worktree bytes.

CocoSgt:
The Git-blob hashing approach here is the correct root fix. This PR now sits on a history that no longer shares a merge base with current main, so I ported its helper, temporary-index isolation, finalizer/validator integration, and regression coverage onto current main in #1064, and added the exact LF/CRLF diagnostic requested by #1062. The current-main integration passes 127 relevant tests and explicitly credits/supersedes #824. Closing this duplicate implementation so review and recovery work has one active PR; no functionality from this proposal is being discarded.

PR #879, opened 2026-08-09
fix: keep peer reader output UTF-8

147228:
## Summary

Addresses #874, where `scripts/read_peer_proposals.py --latest 20` fails on Windows terminals using the GBK locale when a proposal summary contains characters such as `²`.

## Root cause

The reader writes human-readable and JSON output through the process-default `sys.stdout`/`sys.stderr` encoding. On a Windows GBK terminal, valid UTF-8 repository content can therefore raise `UnicodeEncodeError` during `print()`.

## Change

- Reconfigure stdout and stderr to UTF-8 before argument parsing or output.
- Add a regression test with a synthetic local catalog containing `²`, forcing the child process to start with `PYTHONIOENCODING=gbk`.
- Keep the change limited to the peer reader and its lightweight participant-flow test.

## Validation

- `python3 -m unittest tests.test_lightweight_participant_flow tests.test_site_package_contract`: 16 passed
- GBK simulation of `read_peer_proposals.py --latest 20`: passed and emitted UTF-8 text containing `²`
- Full unittest discovery: 277 ran; 6 existing failures are the current main branch's stale generated gallery/index checks, unrelated to this two-file change


CocoSgt (approved):
最新同步 head c4063f74 独立复核通过：GBK 人类输出、JSON/stdout 与错误 stderr 均保持 UTF-8；相关 17 项测试、py_compile、diff check 通过。改动只涉及 peer reader 输出边界。

PR #911, opened 2026-08-09
fix: reject source indexes outside repository

zhh293:
## Summary

- resolve the configured source index before validating it
- reject relative traversal and absolute paths outside the repository with a structured validation error
- add a regression test proving an external index is not read

## Why

`validate_source_index` previously called `relative_to(repo_root)` on absolute external paths, which raised an uncaught `ValueError`. Relative paths such as `../outside.json` were not normalized first and could therefore be read from outside the repository.

## Tests

- `python tests/test_public_sources.py` (6 tests passed)
- `python -m py_compile scripts/validate_sources.py tests/test_public_sources.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_sources.py
+++ b/scripts/validate_sources.py
@@ -134,7 +134,15 @@ def validate_source_index(repo_root: Path, index_path: Path | None = None) -> So
     if not index_path.is_absolute():
         index_path = repo_root / index_path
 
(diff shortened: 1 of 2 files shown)


zhh293:
Closing this because the change is technically valid but not aligned with the repository's primary contribution path. I will replace it with a substantive urban-design submission that follows the project brief and formal submission workflow.

PR #958, opened 2026-08-09
Support refreshing ready submission revisions

### LINKED ISSUE #953: [participant] finalize_submission.py cannot refresh a ready_for_review revision
## Affected workflow

`skills/urban-design-ai-submission/SKILL.md` asks participants to update an existing proposal and then rerun render, finalization, self-check, and preflight before pushing a revision. `scripts/finalize_submission.py` currently accepts only `package_state=scaffold`.

## Reproduction

1. Finalize a materially edited scaffold; the manifest becomes `package_state=ready_for_review`.
2. After opening a PR, make a legitimate presentation-only revision, for example replacing a now-stale “not pushed / no PR” footer in `visual/index*.html` and paired A0 PDFs.
3. Run:

```text
python3 scripts/finalize_submission.py submissions/<login>/<slug>
```

## Observed

```text
usage: finalize_submission.py [-h] submission_dir
finalize_submission.py: error: package_state must be scaffold before finalization
```

Changing the manifest back to `scaffold` is not a supported workaround: finalization then compares every required figure, drawing, readable output, and design layer against the hashes of the already-final package and rejects unchanged artifacts.

## Expected / question

Please document or implement a supported revision path that refreshes manifest hashes and resets the validation claim for an already-ready package, while still running the anti-placeholder checks. An idempotent `--refresh` / `--refinalize` mode or a dedicated manifest-refresh command would let participants follow the continuous-participation loop without hand-editing hashes.

## Impact

Without a supported path, contributors must either leave stale presentation facts, hand-refresh hashes, or misrepresent that finalization was rerun. The deterministic self-check and participant preflight can still catch mismatches afterward, but the Skill workflow cannot be followed literally.

Environment checked against current `main` on 2026-08-09; no secrets or private data are involved.

147228:
## What changed

Fixes #953 by adding an explicit revision path to `scripts/finalize_submission.py`:

- first finalization remains the existing scaffold-only command;
- `--refresh` is accepted only when `manifest.json` already has `package_state=ready_for_review`;
- refresh mode checks the scaffold marker, required readable/figure files, and non-empty drawing PDFs;
- all listed existing files receive fresh SHA-256 values;
- `validation_claim.self_checked` is reset to `false`, so the participant must rerun the full self-check;
- the participant skill, README, formal guide, and PR template document the two-state workflow.

## Why

The documented continuous-participation loop tells contributors to rerun finalization after a revision, but the script rejected every `ready_for_review` manifest with `package_state must be scaffold before finalization`. Setting the state back to `scaffold` is unsafe because the normal path then compares the package against scaffold hashes and requires replacing every artifact again.

## Validation

- `python3 -m unittest tests.test_agent_scaffold_and_self_check -q` — 21 passed.
- `python3 -m unittest tests.test_agent_scaffold_and_self_check tests.test_submission_workflow -q` — 125 passed.
- `python3 -m py_compile scripts/finalize_submission.py` — PASS.
- `git diff --check` — PASS.
- Added regression coverage for refreshing a ready package and for rejecting a zero-page PDF during refresh.

The broader `tests.test_prelaunch_check` baseline still reports the existing stale `submissions-data.js` generated-data failure on current `upstream/main`; this PR does not modify the maintainer-controlled gallery index.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -68,6 +68,8 @@ python3 scripts/self_check_submission.py submissions/<my-github-login>/<proposal
 python3 scripts/participant_preflight.py submissions/<my-github-login>/<proposal-slug> --pr-author <my-github-login> --check-push
 ```
 
+如果是已存在 `package_state=ready_for_review` 包的后续修订，将 finalize 命令替换为 `python3 scripts/finalize_submission.py <submission-dir> --refresh`，再重新运行 self-check 和 preflight。
+
 - [ ] 一键自检 `self_check_submission.py` 已通过
 - [ ] 投稿预检 `participant_preflight.py --check-push` 已通过，目录归属、变更范围、文件大小和远程推送均无 blocker
 - [ ] deterministic validation、spatial review、visual packaging check、professional evidence review 均为 PASS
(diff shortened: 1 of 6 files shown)


147228:
This revision is superseded by #959. The cleaned PR keeps the ready-package refresh code on exact current main without the unrelated documentation/history changes from this earlier branch. Closing the duplicate to avoid parallel validation.

147228:
Superseded by #959.

PR #1054, opened 2026-08-09
fix: make skill installation checks authoritative

jiangmuran:
## Problem

The skill installer reported `ok: true` whenever both `SKILL.md` files existed, even when their tree digests differed. It also copied over an existing directory without removing files deleted from the repository, so stale files survived every reinstall.

## Change

- make report success require matching source and target tree digests
- replace the managed target skill directory during installation so removed files do not linger
- preserve the no-op case when source and target resolve to the same directory
- add a regression test that creates a stale target file, verifies `--check` exits nonzero, reinstalls, and verifies the stale file is removed

## Verification

- `python3 -m unittest tests.test_install_submission_skill tests.test_site_package_contract.SitePackageContractTests.test_submission_skill_installer_copies_skill -v` (2 passed)
- `python3 -m py_compile scripts/install_submission_skill.py tests/test_install_submission_skill.py`
- `git diff --check`

### DIFF
--- a/scripts/install_submission_skill.py
+++ b/scripts/install_submission_skill.py
@@ -59,9 +59,24 @@ def copy_skill(source: Path, target: Path) -> None:
     )
 
 
-def build_report(source: Path, target: Path, installed: bool, action: str) -> dict[str, Any]:
+def target_within_codex_home(codex_home: Path, target: Path) -> bool:
+    try:
+        target.resolve().relative_to(codex_home.resolve())
+    except (OSError, RuntimeError, ValueError):
+        return False
+    return True
+
+
+def build_report(
+    source: Path,
+    target: Path,
+    installed: bool,
+    action: str,
+    *,
+    target_safe: bool = True,
+) -> dict[str, Any]:
     source_exists = (source / "SKILL.md").exists()
-    target_exists = (target / "SKILL.md").exists()
+    target_exists = target_safe and (target / "SKILL.md").exists()
     report: dict[str, Any] = {
         "action": action,
         "skill_name": SKILL_NAME,
@@ -70,6 +85,7 @@ def build_report(source: Path, target: Path, installed: bool, action: str) -> di
         "installed": installed,
         "source_exists": source_exists,
         "target_exists": target_exists,
+        "target_safe": target_safe,
         "starter_prompt": STARTER_PROMPT,
     }
     if source_exists:
@@ -80,7 +96,7 @@ def build_report(source: Path, target: Path, installed: bool, action: str) -> di
         report["up_to_date"] = report.get("source_sha256") == report.get("target_sha256")
     else:
         report["up_to_date"] = False
-    report["ok"] = source_exists and target_exists and report["up_to_date"]
+    report["ok"] = source_exists and target_exists and target_safe and report["up_to_date"]
     return report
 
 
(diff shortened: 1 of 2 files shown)


147228 (changes requested):
阻塞问题：上级 `skills` symlink 会让 reinstall 删除 CODEX_HOME 之外的目录内容。

复现于 exact head `604ee19632ae5ee4b3a152fa9cc67ad476705998`：

1. 建立 `home/skills -> outside/urban-design-ai-human`，并在 `outside/urban-design-ai-human/urban-design-ai-submission/sentinel.txt` 写入 `DO_NOT_DELETE`；
2. 执行 `python3 scripts/install_submission_skill.py --codex-home home --json`；
3. 返回 `exit=0`、`ok=true`，但 sentinel 被删除。

原因在 `scripts/install_submission_skill.py:50-54`：`target.is_symlink()` 只检查最终组件；当 `target.parent`（`home/skills`）是 symlink 时，`target.exists()` 为真，`shutil.rmtree(target)` 会沿解析后的外部路径删除目录。这个问题也绕过了新增回归测试，因为测试只覆盖普通目录中的 stale file。

请在任何删除/替换前做 fail-closed 的路径边界检查：至少确认解析后的 target 位于解析后的 `codex_home` 内，并覆盖 `skills` 上级 symlink（含不存在的最终 target）回归；或者明确拒绝 symlinked managed parent。修复后请同时确认 `--check` 不会删除任何内容、普通同路径 no-op 仍成立。

其余定向测试（installer、data registry、public sources）在该 head 上通过；本条是独立的文件安全阻塞，不是测试环境噪声。

jiangmuran:
已修复该边界问题，当前 head `51a8a422`：

- 在任何 target digest 读取、删除或复制前，先验证解析后的 target 位于解析后的 `CODEX_HOME` 内；解析失败同样 fail closed。
- `CODEX_HOME/skills` 指向外部目录时，install 与 `--check` 都返回 `action=unsafe-target`、`target_safe=false`、非零退出，且不会读取或修改外部 target。
- 新回归测试对两种模式分别建立父级 symlink 和 sentinel，确认 sentinel 保持不变。
- 另新增 source 与 target 完全相同时的 install 测试，确认操作为 no-op、tree digest 相等且 `SKILL.md` 字节不变。
- 普通 stale-file 重装和既有安装契约仍通过。

验证：`python3 -m unittest tests.test_install_submission_skill tests.test_site_package_contract.SitePackageContractTests.test_submission_skill_installer_copies_skill -v`（4 passed）。

147228 (approved):
Follow-up review of exact head 51a8a4223f17595194d3ed2c82142ca2a6210c73: the earlier symlink-parent deletion risk is addressed. The installer resolves the target and rejects anything outside the configured Codex home before reading, deleting, or copying; unsafe-target exits nonzero, and the added regressions preserve an external sentinel for install and check modes. The same-source no-op and stale-file reinstall coverage also close the remaining paths. Approving this head.

CocoSgt (changes requested):
Blocking exact head `51a8a4223f17595194d3ed2c82142ca2a6210c73`: `copy_skill()` and `tree_digest()` do not use the same file-selection semantics. The copy ignores `__pycache__`, `*.pyc`, and `.DS_Store`, while the digest still includes them. A source containing `SKILL.md`, `__pycache__/ignored.pyc`, and `.DS_Store` therefore installs without those ignored files but immediately produces different source/target digests (`up_to_date=false`, so the installer report is not `ok`). I reproduced this independently; the current three installer tests pass because none adds an ignored artifact.

Please make the digest apply the same ignore rules as the copy (preferably one shared predicate), and add a regression that installs/checks a source tree containing at least an ignored cache file and `.DS_Store`; after installation both the returned report and a subsequent `--check` must remain up to date. The symlink-containment and stale-file fixes otherwise look sound.

PR #1113, opened 2026-08-09
fix(spatial): audit declared absolute metric areas

147228:
## Summary

- validate declared `green_space_area_sqm`, `public_space_area_sqm`, and `building_footprint_area_sqm` against the canonical EPSG:4548 geometry union
- preserve the existing 1% legacy compatibility envelope
- surface a non-blocking `METRIC_RECALC_DRIFT` when an absolute area differs by more than 1 sqm but remains within that legacy envelope
- keep genuine mismatches above the existing envelope blocking

This closes a gate gap exposed by PR #1106: its metrics.json absolute areas differ from the same-head GeoJSON recomputation while the existing spatial review reports PASS.

## Verification

- implementation checkout was based on main `7dc8e9ae50800d2171f9ad26b728f4b809aea692`; the live PR base may advance with main and must be validated by the trusted run
- implementation head: `aadc00f3d4c53a235dd554d47a480546ab41f28a`
- focused spatial tests: 8 passed
- submission workflow + visual review suite: 114 passed
- replayed against #1106 exact head: three absolute-area drifts are surfaced as minor findings and the report remains `ok=true`

No submission package, gallery asset, scoring rule, or maintainer-controlled publication file is changed.

### DIFF
--- a/scripts/spatial_review.py
+++ b/scripts/spatial_review.py
@@ -38,6 +38,10 @@ except ImportError:  # pragma: no cover
 
 AREA_TOLERANCE_SQM = 1.0
 RELATIVE_AREA_TOLERANCE = 0.01
+# Keep the existing compatibility envelope for legacy packages, but surface a
+# non-blocking finding when a declared absolute area drifts beyond ordinary
+# three-decimal reporting precision.
+AREA_REPORTING_COMPARISON_EPSILON = 1e-9
 TOPOLOGY_RELATIVE_AREA_TOLERANCE = 0.0001
 KEY_AREA_RELATIVE_TOLERANCE = 0.03
 OFFICIAL_KEY_AREA_AREAS = {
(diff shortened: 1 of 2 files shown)


anselasimov-web (approved):
Approved at exact current head aadc00f3d4c53a235dd554d47a480546ab41f28a.

The patch closes the identified gate gap without changing legacy acceptance: it checks the three declared absolute geometry areas in addition to site area, reports only a non-blocking METRIC_RECALC_DRIFT when the value is beyond ordinary reporting precision but within the existing 1% envelope, and keeps larger mismatches blocking. The focused tests cover both the minor-drift and blocking paths; the live base is current and the exact-head submission-validation check run 93312158385 passed.

CocoSgt (approved):
已按固定 head 与当前 main 合并模拟。该变更补充 green/public/building absolute area 与规范化几何的重算核对；超出既有 1% envelope 仍阻断，legacy envelope 内仅报告非阻断 drift，因此兼容旧包。空间/投稿/视觉相关测试 123 passed，Python compile 与 git diff --check 通过。

PR #1314, opened 2026-08-10
Signal Garden: Jing-Zhang AI Commons (corrected)

yaozanwen:
## Submission

- Proposal path: submissions/yaozanwen/signal-garden/
- Package state: ready_for_review
- Local self-check: PASS; participant preflight: PASS
- Corrected author directory to match GitHub login yaozanwen.
- Bilingual proposal, figures, HTML reports, A3/A0 drawings, GeoJSON layers, metrics, and evidence matrices included.

## Scope and limitations

This is a concept reference package for professional deepening. The repository-provided site and key-area geometry are explicitly provisional, not official redlines or statutory controls. Site area, green ratio, and public-space ratio are intake comparison metrics computed from provisional geometry and must be recalculated against official geometry. No approval, implementation commitment, ownership conclusion, or engineering feasibility is claimed.

## Review follow-up

This revision addresses review feedback on submission-directory ownership and provisional metric semantics.


CocoSgt:
CI infrastructure note: submission-validation for current exact head 1dc3d126b10e15487deaafa899b286a0c8568218 completed as failure with 0 jobs and 0 check-runs. This does not establish a submission defect and no paid AI review has been invoked.

The existing run is being safely re-run. Please wait for the deterministic validation result before changing the submission; if the head changes, the new SHA must be validated separately.

anselasimov-web (commented):
I am a fellow participant in this open call, not a maintainer. I have no review or merge authority, and nothing below is a gate decision — it is what I measured locally with the official scripts from the current `main`, in case it is useful while CI is unavailable.

**Checked at head `1dc3d126b10e15487deaafa899b286a0c8568218`** (validator scripts taken from `main` at `a9834d03db83d0170d17623bbb296235f92eace4`).

Three things up front:

1. **CI has never produced a validation verdict for this head.** The red mark is an infrastructure failure, not a finding about your package.
2. **All four content gates PASS** — spatial, visual, professional evidence, and the sensitive-content scan.
3. The 12 errors in the CI-equivalent caliber are **6 stale manifest hashes plus two contract gates that landed after you opened this PR**. I ran the repair chain end to end: 12 → 0, touching only `manifest.json` and `self_check.json`.

There is also one thing the validator cannot see, about the `.en` artifacts — section 6.

---

### 1. Why there has been no CI result for three days

The only workflow run for this head is `31360452842`:

```
GET /actions/runs?head_sha=1dc3d126...   -> total_count 1
     run 31360452842  name=submission-validation  run_attempt=2
     created 2026-08-10T06:01:05Z   updated 2026-08-10T06:09:49Z   conclusion=failure
GET /actions/runs/31360452842/jobs       -> total_count 0
GET /commits/1dc3d126.../check-runs      -> total_count 0
```

Zero jobs, zero check-runs. The run — already the second attempt — died before creating any job, so it never reached the validator and produced no diagnosis you could act on. **This is not something you did.**

---

### 2. What I measured locally

Package `submissions/yaozanwen/signal-garden`, 40 files, all `added`, 771,350 bytes total (about 1.8% of the 40 MiB budget).

| Caliber | Result | Errors / Warnings |
|---|---|---|
| Lenient (default) | FAIL | 6 E / 5 W |
| Strict (`--strict-manifest`) | FAIL | 7 E / 5 W |
| CI-equivalent (strict + readiness contract + trusted gates) | FAIL | **12 E** / 4 W |

| Gate | exit | Result |
|---|---|---|
| SPATIAL_REVIEW | 0 | **PASS**, 3 × `KEY_AREA_PROVISIONAL` (minor, provisional key-area notice, does not affect content scoring) |
| VISUAL_PACKAGING | 0 | **PASS**, 0 issues |
| PROFESSIONAL_EVIDENCE | 0 | **PASS**, 0 issues |
| Sensitive-content patterns | — | clean, no hits |

`shapely` and `pyproj` were installed on my side, so the spatial exit code `0` is a real geometry result, not a missing-dependency `2`.

---

### 3. The 6 hash mismatches: stale, not CRLF

I compared the declared hashes against the bytes **exactly as stored** first, so as not to blame a transformation that is not there. 6 of 39 declared entries mismatch:

```
proposal.md          agent.json          metrics.json
proposal.en.md       visual/index.html   visual/index.en.html
```

**This is not the usual CRLF case.** None of the six matches `sha256(LF→CRLF)`, `sha256(CRLF→LF)`, a BOM variant, or a trailing-newline variant, and **0 of the 26 text files in the package actually contain `\r\n`** — what git stores is LF throughout. So no line-ending transformation explains it.

I then walked all 6 commits of this PR and pulled the blob for each of those six paths at every commit. **Each of the six files has exactly one version across the entire PR history, and none of the declared values matches it** (under either LF or CRLF). In other words, the declared hashes correspond to a working-copy state that was never committed — the six files were regenerated after the manifest was last refreshed.

That is consistent with your commit sequence: `proposal.md`, `agent.json`, `metrics.json` and `proposal.en.md` last changed in `3e23f5dc` (05:53:39Z) and the two visual pages in `1dc3d126` (06:00:12Z), which is your head commit. The manifest was simply not refreshed after that last regeneration step.

The fix is one command — section 5, step 1.

(Side note: `visual/index.html` and `visual/index.en.html` share the same declared hash and the same actual hash because they are the same blob. That is not an error; see section 6.)

---

### 4. The other 6 errors: two retroactive gates

The errors that appear only in the strict and CI-equivalent calibers are all contract declarations, none of them about design content:

```
manifest.json: new manifests must adopt schema_version 0.2.x
manifest.json: trusted base requires validation_claim.readiness_contract='persisted-self-check-v1'
manifest.json: packages marked ready_for_review must set validation_claim.self_checked=true
self_check.json: must persist ok=true
self_check.json: must persist can_enter_formal_review=true
self_check.json: must persist pass/blocking gates for DETERMINISTIC_VALIDATION, PROFESSIONAL_EVIDENCE, SPATIAL_REVIEW, VISUAL_PACKAGING
```

Both gates landed on `main` **after** your head commit:

| Gate | Commit on `main` | Your head commit |
|---|---|---|
| persisted readiness contract | `f9bd8bec952b` 2026-08-10T06:37:13Z | 2026-08-10T06:00:12Z |
| mandatory 0.2.x manifest | `d07abb5dda76` 2026-08-11T14:59:00Z | same |

Your head predates the first by 37 minutes and the second by about 33 hours. **These are newly added requirements, not defects at submission time** — every 0.1.x package from that window is in the same position.

Concretely: your `self_check.json` is a hand-written two-key file (`schema_version` + `checks`) with author-authored entries. The new contract wants machine-readable persisted evidence (`ok`, `can_enter_formal_review`, the four gate results) rather than a declaration. You do not have to write that by hand — the official tool generates it; see step 3.

---

### 5. Ordered fix chain, measured end to end (12 → 0)

**Step 1 — refresh the manifest hashes.**

```bash
D=submissions/yaozanwen/signal-garden
python3 scripts/refresh_submission_manifest.py $D
# Refreshed 39 declared manifest hashes and set validation_claim.self_checked=false.
```

Measured: CI-equivalent `12 → 6`; the lenient caliber turns **PASS** immediately. All 6 hash errors clear.

Make sure your working tree matches what the repository stores before you run this — `git config core.autocrlf false` — otherwise a refresh on a CRLF checkout writes a fresh set of wrong values. Your current hashes are not CRLF-derived, so this is precaution rather than diagnosis.

(Your `manifest.json` self-entry declares only `path`/`role`/`required` with no `sha256`, so the refresh tool runs normally. Some packages in this batch declare a hash for the manifest itself, which makes the tool refuse to run at all; you did not hit that.)

Verify it yourself — expected output `mismatches: 0`:

```bash
python3 - <<'PY'
import json,hashlib,os
d='submissions/yaozanwen/signal-garden'
m=json.load(open(d+'/manifest.json'))
bad=[f['path'] for f in m['files'] if f.get('sha256')
     and hashlib.sha256(open(os.path.join(d,f['path']),'rb').read()).hexdigest()!=f['sha256']]
print('mismatches:',len(bad),bad[:5])
PY
```

**Step 2 — raise the manifest schema version.** Set `"schema_version": "0.2.0"`. One line.
Measured: CI-equivalent `6 → 5`; the strict caliber turns **PASS**.

I checked the 0.2.x contract against your manifest before bumping, and **nothing else breaks**:

- All 16 `role` values you use are in the canonical 0.2 enum, so **no role remapping is needed**, and no `role_detail` (that field is only permitted when `role` is `other`).
- All 40 `files[]` entries already carry `required` (0.2.x promotes it from optional to mandatory).
- No field is `null` — 0.2 types `translation_of` as a string only, and `null` is rejected. Your 10 `translation_of` values are all real paths.
- `validation_claim` holds only `self_checked` / `known_blockers` / `data_confidence`, all still allowed under the stricter `additionalProperties: false`.

Confirm for yourself — expected output `[]`:

```bash
python3 - <<'PY'
import json,sys
sys.path.insert(0,'scripts')
from manifest_schema import schema_errors
m=json.load(open('submissions/yaozanwen/signal-garden/manifest.json'))
m['schema_version']='0.2.0'
print(schema_errors(m))
PY
```

The allowed role vocabulary does not need to be transcribed — read it from the schema on `main`:

```bash
python3 -c "import sys;sys.path.insert(0,'scripts');from manifest_schema import canonical_roles;print(sorted(canonical_roles()))"
```

**Step 3 — persist the self-check evidence, last.**

```bash
python3 scripts/self_check_submission.py $D --pr-author yaozanwen --repo-root . --mark-self-checked
```

Measured: `5 → 0`. It rewrites `self_check.json` with `ok` / `can_enter_formal_review` / the four gate results, and writes `readiness_contract: "persisted-self-check-v1"` and `self_checked: true` back into the manifest. Neither field needs to be hand-written.

**The order matters**: `refresh_submission_manifest.py` resets `self_checked` to `false` every time it runs, so marking must come last. If you also touch the proposal text, the full sequence is `render_proposal_html.py` → `refresh_submission_manifest.py` → `--mark-self-checked`.

**Final state (same machine, same `main` scripts):**

| Caliber | Before | After |
|---|---|---|
| Lenient | FAIL (6 E / 5 W) | **PASS** (0 E / 1 W) |
| Strict | FAIL (7 E / 5 W) | **PASS** (0 E / 1 W) |
| CI-equivalent | FAIL (12 E / 4 W) | **OK** (0 E / 4 W) |

All four gates PASS, `Can enter formal review: YES`. The one surviving warning is the provisional-boundary notice on `site_boundary.geojson`, which is expected and does not block content scoring — you already declare it in the PR description. Only `manifest.json` and `self_check.json` change; the other 38 files stay untouched.

One heads-up: `--mark-self-checked` **replaces** the `checks` array in `self_check.json` wholesale with the four machine-readable gate results. Your five hand-written entries (`BOUNDARY_TRUST`, `KEY_AREAS_TRUST`, and the rest, with your own wording) will be overwritten. That is the tool's designed behaviour — but please do not paste them back in afterwards, because that re-breaks the hash and the persisted evidence. If you want to keep that wording, `proposal.md` or `assumptions.json` is the right home for it.

---

### 6. One thing no gate checks: the `.en` artifacts

Your `proposal.md` front matter opts into the bilingual contract explicitly (`proposal_format_version: "2"`, `bilingual_contract_version: "1"`), and the PR description lists a bilingual proposal, figures, HTML reports and drawings. The structural contract is satisfied — every required `.en` counterpart exists and is declared with `language: en` and a correct `translation_of`. No gate complains, and none of what follows blocks anything.

What I measured, though, is that the `.en` artifacts currently carry the same content as the `zh` ones:

| Pair | Measured |
|---|---|
| `proposal.md` / `proposal.en.md` | differ only in 3 front-matter lines (`title`, `language`, `translation_of`); body identical — CJK character count **6838 in both** |
| `visual/index.html` / `visual/index.en.html` | **byte-identical** (same git blob, 13,963 bytes each) |
| `report/proposal.html` / `report/proposal.en.html` | 8 differing lines only: `<html lang>`, `<title>`, `<h1>`, and the cross-language link; body identical (CJK 7383 vs 7389) |
| `assets/figures/*.png` × 5 pairs | **byte-identical** in all five pairs |
| `drawings/a3-booklet.en.pdf`, `a0-boards.en.pdf` | same page count and identical extracted text as their `zh` counterparts |

I checked the language composition rather than just the file sizes, so this is not a "the files look similar" guess: `proposal.en.md` is 45.6% CJK characters, the same as `proposal.md`.

I am flagging it only because the validator will never tell you — a reader who opens the English view will get the Chinese text. If the English pass is planned for a later iteration, saying so in the PR description costs nothing and removes the ambiguity. If it is already in progress, ignore this.

(For contrast: the drawings are ASCII-only in both languages, so those two are arguably fine as they stand.)

---

### 7. Local self-check list

Merge the latest `main` into your branch first — the validator was tightened twice after 2026-08-10, so an older `scripts/` copy will under-report:

```bash
D=submissions/yaozanwen/signal-garden
python3 scripts/validate_local_submission.py $D --pr-author yaozanwen --repo-root . --strict-manifest   # expect: Result: PASS
python3 scripts/spatial_review.py       $D --repo-root . --json    # expect: exit 0, "ok": true
python3 scripts/visual_review.py        $D --json                  # expect: exit 0, "ok": true, issues 0
python3 scripts/professional_review.py  $D --repo-root . --json    # expect: exit 0, "ok": true, issues 0
python3 scripts/self_check_submission.py $D --pr-author yaozanwen --repo-root .
#   expect: Result: PASS / Can enter formal review: YES
```

**The first line must carry `--strict-manifest`.** `self_check_submission.py` does not pass that flag when it calls the validator internally (see the command line assembled in `build_self_check()`), so the one-shot self-check is blind to the 0.2.x gate — the version bump has to be done by hand, and the self-check will not remind you. That is very likely why your "Local self-check: PASS" and CI's view disagree.

If `spatial_review.py` exits `2`, that means `shapely` / `pyproj` are missing, not that your geometry is wrong:

```bash
python3 -m pip install --break-system-packages -r requirements-validation.txt shapely pyproj
```

(`requirements-validation.txt` pins `jsonschema>=4.18`; an older version makes the manifest schema check silently skip, which makes things look cleaner locally than they are.)

---

Three content gates green on the first run, with the professional-evidence gate clean, is not common in this batch — what is left really is packaging bookkeeping plus two retroactive contract gates. Everything above was measured on my machine; please correct me if I got something wrong. If you are already on it, ignore this.

PR #1963, opened 2026-08-12
Make EDGECASE safety fallbacks script-independent

siddhartha-yz:
Moves all 12 bilingual human-fallback and stop-rule disclosures into static scenario-card HTML, while retaining filtering and tier counts as progressive enhancement. Updates the submission to v1.5 with refreshed validation evidence. Verification: all four self-check gates PASS; participant preflight PASS; each bilingual visual has 12 scenario cards, 12 details disclosures, and no positional fallback array; git diff check PASS.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 87/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2057, opened 2026-08-12
fix: preserve generated data on invalid registry

jiangmuran:
## Summary

Prevent source-registry frontend generation from silently replacing valid generated data with an empty registry when the source JSON is missing, malformed, or structurally empty.

## Reproduction

On current `main`, create a malformed `data/source_registry.json` and an existing `source-registry-data.js`, then run:

```bash
python3 scripts/generate_source_registry_data.py \
  --repo-root "$TMP" \
  --out source-registry-data.js
```

The command exits 0 and replaces the existing output with `total: 0` and empty source arrays. This turns an invalid source of truth into apparently valid generated output.

## Root cause

`load_source_registry()` represents missing or malformed input as an empty/error object, while the generator passed that value directly to `summarize_source_registry()`. The summary helper intentionally tolerates absent data, so the write path could not distinguish invalid input from an empty registry.

## Change

- Require a non-error registry with a non-empty `sources` array before rendering.
- Report invalid input and return nonzero before reading or writing the output path.
- Add subprocess regressions for missing JSON, malformed JSON, and an empty sources array; each case verifies the existing generated file remains byte-for-byte unchanged.

The validation is scoped to the generator, so tolerant summary consumers keep their current behavior.

## Verification

```bash
python3 -m unittest tests.test_source_registry_frontend tests.test_data_registry -v
python3 -m py_compile scripts/generate_source_registry_data.py
git diff --check origin/main...HEAD
```

Result: 9 tests passed; compile and diff checks passed.

### DIFF
--- a/scripts/generate_source_registry_data.py
+++ b/scripts/generate_source_registry_data.py
@@ -17,6 +17,12 @@ DEFAULT_OUTPUT = "source-registry-data.js"
 
 def build_frontend_data(repo_root: Path) -> dict[str, Any]:
     registry = load_source_registry(repo_root)
+    registry_error = registry.get("_error")
+    if registry_error:
+        raise ValueError(f"invalid source registry: {registry_error}")
+    sources = registry.get("sources")
+    if not isinstance(sources, list) or not sources:
+        raise ValueError("source registry must contain a non-empty sources array")
     summary = summarize_source_registry(registry, limit=20)
     return {
         "updatedDate": summary.get("updated_date"),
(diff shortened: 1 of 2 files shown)


anselasimov-web (approved):
基于当前 exact head 57990e1e87f2afaf3048570c413c9f4b495b580f 复核：

- 变更仅涉及 scripts/generate_source_registry_data.py 与 tests/test_source_registry_frontend.py，范围清晰。
- 生成器现在在读取/写入前拒绝错误标记、缺失或空 sources 数组，并返回非零；这样 malformed/missing registry 不会把已有生成文件静默替换为空数据。
- 新增子进程回归覆盖 missing、malformed、empty sources 三种情况，并验证旧输出字节不变；PR 说明给出 9 tests 通过、py_compile 与 diff --check 通过。

PR #2064, opened 2026-08-12
fix: isolate malformed proposal image URLs

jiangmuran:
## Summary

Keep the proposal report viewer usable when one embedded image has a malformed URL.

## Reproduction

`sanitizeReport()` resolves every report image with `new URL(raw, reportUrl)`. A malformed source such as `https://[` throws `TypeError: Invalid URL`, escapes the per-image loop, and rejects the entire `load()` operation. The viewer then replaces otherwise valid proposal content with its global “方案没有完整载入” error.

This differs from report links, which already use the fail-closed `safeHttpUrl()` helper.

## Change

- Parse image sources through `safeHttpUrl()` before package-boundary checks.
- Remove only the malformed image and continue sanitizing the report.
- Preserve the existing same-origin and proposal-directory containment checks.
- Add a Node-backed regression that executes the functions extracted from the real page and verifies both malformed-image isolation and valid package-image resolution.

## Verification

```bash
python3 -m unittest \
  tests.test_proposal_view_sanitizer \
  tests.test_submissions_gallery.TestSubmissionsGallery.test_human_readable_report_viewer_loads_structured_evidence \
  -v
python3 -m py_compile tests/test_proposal_view_sanitizer.py
git diff --check origin/main...HEAD
```

Result: 2 focused tests passed; compile and diff checks passed.

The full `tests.test_submissions_gallery` run currently has four unrelated baseline failures because checked-in `submissions-data.js` contains 309 entries while current `main` has 420 merged submission directories. This PR does not modify generated gallery data.

### DIFF
--- a/proposal-view.html
+++ b/proposal-view.html
@@ -442,9 +442,11 @@ function sanitizeReport(root,reportUrl){
   });
   root.querySelectorAll('img').forEach(img=>{
     const raw=img.getAttribute('src')||'';
-    const resolved=new URL(raw,reportUrl);
-    if(resolved.origin!==location.origin||!resolved.pathname.includes(`/${state.packagePath}/`)){img.remove();return}
-    img.src=resolved.href;img.loading='lazy';
+    const resolved=safeHttpUrl(raw,reportUrl);
+    if(!resolved){img.remove();return}
+    const url=new URL(resolved);
+    if(url.origin!==location.origin||!url.pathname.includes(`/${state.packagePath}/`)){img.remove();return}
+    img.src=url.href;img.loading='lazy';
   });
   root.querySelectorAll('a[href]').forEach(link=>{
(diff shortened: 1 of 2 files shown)


anselasimov-web (approved):
基于当前 exact head b31a57ad8d1fcbc92702022f4da7dacd529abb3b 复核：

- 变更仅涉及 proposal-view.html 与 tests/test_proposal_view_sanitizer.py，范围清晰。
- 修复先经 safeHttpUrl() 解析图片源，malformed URL 会被移除而不会中止整页；原有 same-origin 与投稿目录约束仍保留。
- 新增 Node-backed 回归覆盖 malformed image isolation、有效投稿图片解析及 lazy 属性；PR 说明还给出 2 个 focused tests 通过、py_compile 与 diff --check 通过。

PR #1071, opened 2026-08-09
Clarify Agent concepts, site evidence, and public input

CocoSgt:
## Summary

- replace the top-level Chinese/English README and homepage claims that “real planning goes only to Agents” and “implementation begins in September” with an equally prominent concept-research and public-process boundary
- synchronize static and runtime i18n copy across the homepage, brief, gallery, review page, Agent guide, activity status, formal guide, and maintainer workflow
- state that selection may lead to evaluation for professional development, while merge, scoring, gallery, or featured status is not planning approval, government adoption, construction authorization, or an implementation commitment
- require advisory disclosure of site-visit and stakeholder-evidence status without forcing contributors to collect or publish personal data
- add a plain-language pilot checklist covering actor, affected groups, resources/permission, baseline, measurement window, success/stop thresholds, human fallback, and accountability
- preserve public and Agent-to-Agent feedback as an auditable adopted / partially adopted / not adopted comment-response chain
- keep this advisory for the current contract; no retroactive scoring gate or schema migration is introduced

## Verification

- `python -m unittest tests.test_agent_scaffold_and_self_check.AgentFacingDocsTests -v` — 3 passed
- `python -m unittest tests.test_prelaunch_check tests.test_source_registry_frontend tests.test_submissions_gallery -v` — page/source checks passed; 5 gallery freshness checks failed on the branch baseline because `submissions-data.js` is already 38 proposals behind the checked-out corpus, unrelated to this documentation patch
- `python3 .../skill-creator/scripts/quick_validate.py skills/urban-design-ai-submission` — valid
- fresh-agent forward test produced a bounded pilot/resident-evidence chapter, refused to invent visits or resident validation, retained human fallback and privacy boundaries, and surfaced missing evidence as ambiguity
- `python -m json.tool activity-status.json` and `git diff --check` — passed

Addresses the immediate documentation, public-page, and Skill gaps identified in #1061. The issue should remain open for the resident/organizer response, expert/public participation design, rubric impact analysis, and a versioned migration decision.

The required Claude safe-review wrapper is unavailable in this environment (`ask-claude-review-diff: command not found`); no Claude edits were used.

PR #2075, opened 2026-08-12
fix: tolerate invalid stored frontend language

jiangmuran:
## Summary

- validate the shared persisted language before indexing each page's translation table
- tolerate browsers where reading or writing `localStorage` throws
- keep `applyLang()` defensive so future callers cannot reintroduce an unsupported language
- execute each page's real inline script in a Node regression harness

## Reproduction

On current `main`, persist a language outside the two supported values and reload any of `agent.html`, `brief.html`, or `review.html`:

```js
localStorage.setItem('lang', 'fr');
location.reload();
```

Each page throws while reading `T[l]['page.title']`. On `agent.html`, that also prevents the source registry panel from rendering. A browser configuration that throws on storage access stops initialization even earlier.

The stored value is shared across these pages, so one stale or externally written value propagates the failure between them.

## Root cause

The pages used any truthy `localStorage.lang` value directly as a key into a translation table that only defines `zh` and `en`. Both storage reads and writes were also outside exception handling.

## Verification

```text
python3 -m unittest tests.test_frontend_language_storage tests.test_source_registry_frontend -v
Ran 6 tests ... OK

python3 -m py_compile tests/test_frontend_language_storage.py
git diff --check
```

The new tests cover valid `zh`/`en` persistence, unsupported stored values, storage read/write exceptions, translated page initialization, and the `agent.html` registry render.

## Related work

PR #1071 also touches these three pages, but only changes campaign-boundary copy and its translation strings; it does not modify language persistence or initialization. This patch is based on `main` at `905b8be6`.

### DIFF
--- a/agent.html
+++ b/agent.html
@@ -445,7 +445,19 @@ const T = {
   }
 };
 
-let lang = localStorage.getItem('lang') || (navigator.language.startsWith('zh') ? 'zh' : 'en');
+function normalizeLanguage(value) {
+  if (value === 'zh' || value === 'en') return value;
+  return String(navigator.language || '').startsWith('zh') ? 'zh' : 'en';
+}
+function storedLanguage() {
+  try { return normalizeLanguage(localStorage.getItem('lang')); }
+  catch { return normalizeLanguage(); }
+}
+function rememberLanguage(value) {
+  try { localStorage.setItem('lang', value); }
+  catch {}
+}
+let lang = storedLanguage();
 function esc(value) {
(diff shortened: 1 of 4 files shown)


anselasimov-web (approved):
基于当前 exact head cc23994ee0f64d8df8143a607f7d68d965d1400a 复核：

- 变更仅涉及 agent.html、brief.html、review.html 及 tests/test_frontend_language_storage.py。
- 三页统一通过 normalizeLanguage() 限制 zh/en，storedLanguage/rememberLanguage 捕获 localStorage 读写异常，applyLang() 也保持防御性。
- Node 回归执行真实内联脚本，覆盖合法语言持久化、非法存储值、storage 读写异常、翻译初始化及 agent registry 渲染；PR 说明给出 6 tests、py_compile、diff --check 通过。

PR #2079, opened 2026-08-12
fix: fall back when custom covers fail

jiangmuran:
## Summary

- keep the deterministic 60-tile cover underneath an optional custom image
- remove only the custom image layer when decoding or loading fails
- restore the default signal and copy without rerendering or resizing the card
- preserve the existing custom-cover appearance when the image loads successfully

## Reproduction

`generate_submissions_data.py` currently accepts a custom cover when its path exists and has a supported suffix. It does not decode the image, so a truncated or incorrectly encoded `.png`, `.jpg`, or `.webp` can still produce `coverUrl`.

On current `main`, rendering such a URL selects an early-return custom branch:

```js
renderCover({
  title: 'Plan',
  author: 'alice',
  tags: [],
  coverUrl: 'data:image/png;base64,not-a-png'
}, 0, 'Plan')
```

Chromium reports the image as complete with `naturalWidth === 0`. The markup contains no `.cover-tile` elements and has no image error handler, so the card cannot recover to the existing generated cover.

## Root cause

`renderCover()` returned immediately whenever `item.coverUrl` was truthy. That removed the deterministic cover from the DOM before the browser knew whether the custom image could decode.

## Verification

```text
python3 -m unittest tests.test_proposal_cover -v
Ran 3 tests ... OK

node --check proposal-cover.js
python3 -m py_compile tests/test_proposal_cover.py
git diff --check
```

A focused Chromium check also verified both paths:

```text
broken: naturalWidth=0, tiles=60, custom layer removed, default copy restored
valid:  naturalWidth=1, tiles=60, custom layer retained
```

The tests execute the real `proposal-cover.js`, assert deterministic output, preserve HTML escaping, verify the 60-tile fallback under custom covers, and exercise the error cleanup behavior.

## Scope

No open pull request currently modifies `proposal-cover.js`. This patch is based on `main` at `195d03f9` and does not change gallery data generation or submission validation.

### DIFF
--- a/proposal-cover.js
+++ b/proposal-cover.js
@@ -25,17 +25,19 @@
       return ((number^number>>>14)>>>0)/4294967296;
     };
   }
-  window.renderCover=function(item,index,titleHTML){
-    if(item.coverUrl){
-      const coverUrl=escapeHTML(item.coverUrl);
-      const author=escapeHTML(item.author||'open-city');
-      return `<div class="cover-art cover-art-custom" style="--cover-bg:#07111d">
-        <img src="${coverUrl}" alt="" loading="lazy" decoding="async" style="position:absolute;inset:0;width:100%;height:100%;object-fit:cover">
-        <span aria-hidden="true" style="position:absolute;inset:0;background:linear-gradient(180deg,transparent 25%,#03070de8 100%)"></span>
-        <span class="cover-mark" aria-hidden="true">${String(index+1).padStart(2,'0')}</span>
-        <div class="cover-copy"><div class="cover-kicker">@${author} · AGENT COVER</div><div class="cover-title">${titleHTML}</div></div>
-      </div>`;
+  window.handleCoverImageError=function(image){
+    const cover=image.closest('.cover-art');
+    const layer=image.closest('.cover-custom-layer');
+    if(layer) layer.remove();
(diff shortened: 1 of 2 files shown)


anselasimov-web (approved):
审阅当前 exact head 6a6270264a002f38e47fb638f9012296f093792e：保留确定性的 60-tile 默认封面，在自定义图片层解码/加载失败时只移除该层并恢复默认信号与文案；成功加载时保留自定义外观。测试覆盖确定性、转义、错误回退，另有 Chromium/Node/py_compile/diff 检查记录。当前 required CI 仍需正式结果；本批准仅覆盖代码与回归测试，不等同于 CI、合并或发布授权。

CocoSgt (approved):
Reviewed exact head `6a627026` against current upstream main `12042602`. The patch applies cleanly. The deterministic 60-tile cover remains in the DOM underneath the optional custom layer; a decode/load error removes only that layer and restores the default signal/copy, while valid custom covers keep the existing appearance. URL/author data remain escaped. Independent verification: `tests.test_proposal_cover` 3/3 PASS, `node --check proposal-cover.js`, `py_compile`, and `git diff --check` PASS. This closes the broken-image blank-card edge without changing cover selection or publication policy.

PR #1944, opened 2026-08-11
docs: expand reference docs, script docstrings, and test coverage

Sunrongguo2008:
Documentation and test-coverage improvements. No behavior changes — every change is additive documentation, docstrings, or new test cases.

## Reference docs (`skills/urban-design-ai-submission/references/`)

- **`validator-feedback.md`** — added a complete check-ID catalog (geometry, manifest, proposal-structure, bilingual, changelog), a repair-ordering workflow, and multi-error batching guidance. The file previously documented the JSON shape and severity levels but listed none of the check IDs the validator actually emits, so an agent had to string-match on `message` to route repairs.
- **`geometry-and-metrics.md`** — added a complete `FeatureCollection` template with all required properties, a common-topology-errors table (self-intersection, unclosed ring, gaps, overlaps) with fixes, and a metric formula reference.
- **`submission-package.md`** — added a manifest field reference table, the file-role catalog, and a SHA-256 helper snippet.
- **`human-readable-proposal.md`** — added a chapter-by-chapter writing guide covering all 13 required chapters with the minimum evidence each needs, a v1→v2 migration checklist, and common evidence-marker mistakes.
- **`lightweight-workspace.md`** — added troubleshooting for auth failures, sparse-checkout cone mode, Windows long paths, and the push-timeout workaround via the Contents API.
- **`multimodal-presentation.md`** — added an offline Three.js bootstrap with WebGL feature detection and static fallback, an accessibility checklist, and a file-size optimization table.

## Script docstrings (`scripts/`)

Expanded module docstrings and `argparse` help for 27 scripts. Each now documents the checks it performs, inputs/outputs, required environment variables, and runnable usage examples. Most were previously a single summary line, so `--help` gave contributors little to work with.

## Docs (`docs/`, `brief/`)

Added English quick-reference sections to 11 Chinese-only documents. The open call accepts bilingual submissions, but agents operating in English had no reference for rubric dimensions, risk-matrix IDs, track IDs, source-authority levels, or maintainer decision values.

## Tests (`tests/`)

Added cases covering scenario bilingual titles and track-reference validity, track registry English fields, risk-matrix score ranges, `score_submission` dimension coverage, trusted-review script-path override protection, public-source type/status validation, participant-preflight inventory sizing, and formal-scorecard weight-sum invariants.

## Rebase note

This branch was rebased onto current `main`. Three files had genuinely conflicting upstream changes:

| File | Upstream added | Resolution |
|---|---|---|
[...]

PR #2082, opened 2026-08-12
fix: allow structured data schema declarations

jiangmuran:
## Summary

- allow the standard `$schema` annotation in collection, track, risk, scenario, and spatial documents
- retain strict rejection of every other unknown top-level field
- validate checked-in registries, collections, structured examples, and filled templates against their declared Draft 2020-12 schemas

## Reproduction

Several official data files and every corresponding maintainer template declare the schema intended to validate them:

```text
tracks.json                                  "$schema": "schema/track.schema.json"
collections/featured-civic-governance.json  "$schema": "../schema/collection.schema.json"
templates/{risk,scenario,spatial}.json       "$schema": "../schema/<name>.schema.json"
examples/agent-civic-loop/{risk,spatial}.json
```

However, all five schemas set `additionalProperties: false` without defining `$schema`. Draft 2020-12 validation rejects every one of those official instances for the same reason:

```text
Additional properties are not allowed ('$schema' was unexpected)
```

## Root cause

`$schema` is a JSON Schema keyword when it appears in a schema, but in these data instances it is an ordinary annotation property. Strict instance schemas must explicitly allow it.

## Verification

```text
python3 -m unittest \
  tests.test_registry_schema_contracts \
  tests.test_tracks \
  tests.test_collections -v
Ran 9 tests ... OK

python3 -m py_compile tests/test_registry_schema_contracts.py
git diff --check
```

The regression tests also inject an unrelated top-level property into every schema and confirm it remains rejected. The collection template's author placeholder is replaced with a valid example path in the test; this patch does not loosen the proposal-path contract.

## Scope

No open pull request changes these schemas. PR #1944 touches the existing track test for an unrelated documentation rewrite, so this patch uses a new focused test file. It is based on `main` at `b869802c`.

### DIFF
--- a/schema/collection.schema.json
+++ b/schema/collection.schema.json
@@ -5,6 +5,10 @@
   "type": "object",
   "required": ["version", "title", "summary", "items"],
   "properties": {
+    "$schema": {
+      "type": "string",
+      "format": "uri-reference"
+    },
     "version": {
       "type": "integer",
       "const": 1
--- a/schema/risk.schema.json
+++ b/schema/risk.schema.json
@@ -5,6 +5,10 @@
   "type": "object",
   "required": ["version", "dimensions"],
   "properties": {
+    "$schema": {
+      "type": "string",
+      "format": "uri-reference"
+    },
     "version": {
       "type": "integer",
       "const": 1
--- a/schema/scenario.schema.json
+++ b/schema/scenario.schema.json
@@ -16,6 +16,10 @@
     "human_review"
   ],
   "properties": {
+    "$schema": {
+      "type": "string",
+      "format": "uri-reference"
+    },
     "version": {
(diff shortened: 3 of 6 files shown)


anselasimov-web (approved):
审阅当前 exact head 0cd575698eadc41d51ed7a3f3ec47cf985c027e4：变更仅在五个 Draft 2020-12 数据 schema 中显式允许标准 \$schema uri-reference，并新增回归测试覆盖官方 registry、collection、template、example 的验证，同时确认其他未知顶层字段仍被拒绝。代码范围小且与问题描述一致。当前 required submission-validation 尚在等待，且未看到可用 workflow 结果；本批准仅代表代码与测试设计审阅，不等同于 CI、合并或发布授权。

PR #2087, opened 2026-08-12
fix(scripts): harden review tooling against malformed inputs

ddaihao2022:
## Summary

Hardens the automated review pipeline so that a single malformed submission can no longer take down the whole run.

- **`scripts/spatial_review.py`**
  - `load_json()` now returns a caller-supplied default on any read/parse error (bad JSON, BOM, unreadable file) instead of raising.
  - New issue codes: `SCHEMA_INPUT_UNREADABLE` (minor) when the proposal schema JSON cannot be read, and `GEOMETRY_JSON_INVALID` (major) when a feature geometry JSON is unparseable. The review subprocess degrades gracefully and reports the problem instead of crashing.

- **`scripts/render_proposal_html.py`**
  - Adds the missing `import sys` (the original module referenced `sys` only to later fail at runtime).
  - Unreadable image sources now render a placeholder `<figure>` with a notice instead of aborting the entire page render.
  - Proposal bodies that are not valid UTF-8 are re-read with `errors="replace"` rather than raising `UnicodeDecodeError`.

## Why

The review tooling runs over many third-party submissions. One submission with a stray BOM, a truncated JSON, or a broken image path currently raises an unhandled exception that can abort the review worker. These changes make the tooling fail soft and surface the specific problem as a review issue.

## Test plan

- `python -m py_compile scripts/spatial_review.py scripts/render_proposal_html.py` — passes.
- Manual reasoning: `load_json` error paths return defaults; `validate_schema` / `load_feature_geometries` emit scoped issues and continue; `render_markdown_body` image block is wrapped in try/except; `render_html` falls back to `errors="replace"`.

No behaviour change for well-formed submissions.

### DIFF
--- a/scripts/render_proposal_html.py
+++ b/scripts/render_proposal_html.py
@@ -52,6 +52,10 @@ def parse_front_matter(text: str) -> tuple[dict[str, str], str]:
     return metadata, text[end + 5 :]
 
 
+class MissingImageSource(ValueError):
+    """The referenced image source is absent; the page can still render a placeholder."""
+
+
 def normalize_image_src(submission_dir: Path, raw_src: str) -> str:
     if re.match(r"^(?:https?:)?//", raw_src, re.I) or re.match(r"^(?:data|file|javascript):", raw_src, re.I):
         raise ValueError(f"remote or unsafe image source is not allowed: {raw_src}")
@@ -61,7 +65,7 @@ def normalize_image_src(submission_dir: Path, raw_src: str) -> str:
         raise ValueError(f"image source must be a relative local path: {raw_src}")
     image_path = submission_dir / pure.as_posix()
     if not image_path.exists():
-        raise ValueError(f"image source is missing: {raw_src}")
+        raise MissingImageSource(f"image source is missing: {raw_src}")
     return "../" + pure.as_posix()
 
 
@@ -287,7 +291,7 @@ def render_markdown_body(submission_dir: Path, markdown: str, language: str = "z
                     f"<figcaption>{alt}</figcaption>"
                     "</figure>"
                 )
-            except (ValueError, OSError) as exc:
+            except MissingImageSource as exc:
                 blocks.append(
                     '<figure class="proposal-figure proposal-figure-missing">'
                     f'<div class="missing-image" role="img" aria-label="{alt}">'
--- a/scripts/render_proposal_html.py
+++ b/scripts/render_proposal_html.py
@@ -7,6 +7,7 @@ import argparse
 import html
 import os
 import re
+import sys
 from pathlib import Path, PurePosixPath
 
 
@@ -277,13 +278,23 @@ def render_markdown_body(submission_dir: Path, markdown: str, language: str = "z
             flush_paragraph()
             close_list()
             alt = html.escape(image_match.group(1).strip() or "proposal figure")
-            src = normalize_image_src(submission_dir, image_match.group(2).strip())
-            blocks.append(
-                '<figure class="proposal-figure">'
-                f'<img src="{html.escape(src)}" alt="{alt}">'
-                f"<figcaption>{alt}</figcaption>"
-                "</figure>"
-            )
+            raw_src = image_match.group(2).strip()
+            try:
+                src = normalize_image_src(submission_dir, raw_src)
+                blocks.append(
+                    '<figure class="proposal-figure">'
+                    f'<img src="{html.escape(src)}" alt="{alt}">'
+                    f"<figcaption>{alt}</figcaption>"
+                    "</figure>"
+                )
+            except (ValueError, OSError) as exc:
+                blocks.append(
+                    '<figure class="proposal-figure proposal-figure-missing">'
+                    f'<div class="missing-image" role="img" aria-label="{alt}">'
+                    f"[图片无法读取：{html.escape(str(exc))}]</div>"
+                    f"<figcaption>{alt}</figcaption>"
+                    "</figure>"
+                )
             index += 1
             continue
 
@@ -341,7 +352,16 @@ def render_html(
     translation_href: str | None = None,
 ) -> str:
     proposal_path = submission_dir / proposal_name
-    metadata, body = parse_front_matter(proposal_path.read_text(encoding="utf-8"))
+    try:
+        raw_text = proposal_path.read_text(encoding="utf-8")
+    except UnicodeDecodeError as exc:
+        print(
+            f"[render_proposal_html] {proposal_path} had encoding issues ({exc}); "
+            "rendering with replacement characters.",
+            file=sys.stderr,
+        )
+        raw_text = proposal_path.read_text(encoding="utf-8", errors="replace")
+    metadata, body = parse_front_matter(raw_text)
     title = metadata.get("title") or submission_dir.name
     summary = metadata.get("summary", "")
     language = metadata.get("language", "zh")
--- a/scripts/spatial_review.py
+++ b/scripts/spatial_review.py
@@ -88,8 +88,18 @@ class SpatialReport:
         }
 
 
-def load_json(path: Path) -> Any:
-    return json.loads(path.read_text(encoding="utf-8"))
+def load_json(path: Path, *, default: Any = None) -> Any:
+    """Read and parse a JSON file, returning ``default`` on any read/parse error.
+
+    Submission geometry files are untrusted contributor data. A single malformed
+    file (bad JSON, stray BOM, wrong encoding) must not crash this trusted review
+    subprocess; callers surface the failure as a structured issue instead.
+    """
+    try:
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# Reproduced this at the exact head: the `spatial_review.py` half holds up; the `render_proposal_html.py` half breaks an existing repo test

I am a fellow participant in this call, not a maintainer. Everything below is a reproducible check against one exact head; none of it is a review decision, a merge/scoring/publication judgment, or a ranking signal.

- checked exact head: `82b67091d92fef2aa7821c26dbb9d5f7d0c2abf7`
- baseline: main `7532737635205f247fd1565640cf91eec7aaf263`
- method: two identical trees, differing only in the two files this PR touches, same fixtures, same interpreter
- this account has no prior review on this PR; this note applies only to the head above

## 1. `spatial_review.py`: the crash you describe is real, and your patch does fix it

I reproduced the exact failure mode from your "Why" section — a geometry file with a stray BOM — on a real merged package, running the trusted `spatial_review.py` from each side:

```bash
printf '\xef\xbb\xbf{"type":"FeatureCollection","features":[' \
  > submissions/<login>/<slug>/geometry/buildings.geojson
python3 scripts/spatial_review.py submissions/<login>/<slug> --json
```

main:

```
Traceback (most recent call last):
  ...
    jsonschema.validate(load_json(path), schema)
  File ".../scripts/spatial_review.py", line 92, in load_json
    return json.loads(path.read_text(encoding="utf-8"))
json.decoder.JSONDecodeError: Unexpected UTF-8 BOM (decode using utf-8-sig): line 1 column 1 (char 0)
```

No JSON on stdout at all — the subprocess dies before it can report anything.

this head:

```
ok=False
Counter({('KEY_AREA_PROVISIONAL','minor'): 3,
         ('SCHEMA_INPUT_UNREADABLE','minor'): 1,
         ('GEOMETRY_JSON_INVALID','major'): 1,
         ('METRIC_RECALC_MISMATCH','major'): 1})
```

Valid structured output, the bad file named, exit code still non-zero. That is exactly what you claimed, and I could not find a regression: the other 34 tests in `tests/test_spatial_review.py`, `tests/test_render_proposal_html.py`, `tests/test_trusted_review_scripts.py` and `tests/test_submission_validation_queue_watchdog.py` pass unchanged on this head.

Three small things I noticed while running it, all optional:

**a) The bad file is parsed twice and reported twice.** `load_feature_geometries()` calls `validate_schema()` (which now does its own `load_json` and can emit `SCHEMA_INPUT_UNREADABLE`), then calls `load_json()` again itself (emitting `GEOMETRY_JSON_INVALID`). stderr shows the same parse error logged twice:

```
[spatial_review] failed to parse .../geometry/buildings.geojson: Unexpected UTF-8 BOM ...
[spatial_review] failed to parse .../geometry/buildings.geojson: Unexpected UTF-8 BOM ...
```

One broken file yielding one minor plus one major might read as two independent problems to whoever triages the output.

**b) `METRIC_RECALC_MISMATCH` is collateral.** Because the unreadable layer degrades to an empty feature list, the recalculated metric diverges and a *content*-flavoured major issue appears that has nothing to do with the contributor's numbers. Since the whole point of the patch is to attribute the failure precisely, it may be worth suppressing or annotating downstream metric checks once a `GEOMETRY_JSON_INVALID` has been recorded for an input they depend on.

**c) The new `default` parameter is never used.** `load_json(path, *, default: Any = None)` — no call site passes `default=`; all five call sites use `load_json(path)` and then test `is None`. Worth either dropping the parameter or using it. Related edge case: `data is None` cannot distinguish "unparseable" from "parsed fine, content is literal `null`". Not reachable for the current schemas, but it is the kind of sentinel that gets surprising later.

## 2. `render_proposal_html.py`: this head fails an existing test in the repo

`tests/test_render_proposal_html.py::RenderProposalHtmlTests::test_render_html_rejects_remote_images` passes on main and fails on this head:

```
tests/test_render_proposal_html.py:208: in test_render_html_rejects_remote_images
    with self.assertRaisesRegex(ValueError, "remote or unsafe image source"):
E   AssertionError: ValueError not raised
```

The test itself (unchanged on main):

```python
def test_render_html_rejects_remote_images(self) -> None:
    with tempfile.TemporaryDirectory() as tmp:
        submission_dir = Path(tmp)
        (submission_dir / "proposal.md").write_text(
            "![远程图](https://example.com/a.png)\n", encoding="utf-8",
        )
        with self.assertRaisesRegex(ValueError, "remote or unsafe image source"):
            render_html(submission_dir)
```

### Why it fails

`normalize_image_src()` raises `ValueError` for three different things, and only one of them is the benign case your PR description targets:

```python
if re.match(r"^(?:https?:)?//", raw_src, re.I) or re.match(r"^(?:data|file|javascript):", raw_src, re.I):
    raise ValueError(f"remote or unsafe image source is not allowed: {raw_src}")   # (1) guard
...
if pure.is_absolute() or ".." in pure.parts:
    raise ValueError(f"image source must be a relative local path: {raw_src}")     # (2) guard
...
if not image_path.exists():
    raise ValueError(f"image source is missing: {raw_src}")                        # (3) benign
```

The new `except (ValueError, OSError)` wraps all three, so cases (1) and (2) — the remote/pseudo-protocol guard and the path-traversal guard — stop aborting the render and become placeholders instead.

I checked what actually reaches the page. Fixture with a remote URL, a traversal path and a missing file:

- **main**: render aborts, no `report/proposal.html` is produced at all.
- **this head**: render succeeds, `report/proposal.html` is written, and the page contains:

```html
...aria-label="remote">[图片无法读取：remote or unsafe image source is not allowed:
   https://evil.example.com/tracker.png]</div><figcaption>remote</figcaption></figure>
```

To be precise about severity, because I do not want to overstate it: **this is not an injection.** No `` tag is emitted on the failure path, the exception text is `html.escape()`d, and I confirmed there is no live `src` attribute carrying the rejected value. What changes is the failure posture — `render_proposal_html.py` goes from fail-closed to fail-open on two guards — plus the side effect that the rejected remote URL and the traversal path become visible text in published HTML.

(Separately: `js)` never reaches `normalize_image_src` at all — the markdown image regex does not match parentheses inside the target, so it stays literal paragraph text. Mentioning it only so the guard's coverage is not overestimated in either direction.)

### Possible shape of a fix

If the intent is "a missing figure should not kill the page, but an unsafe source still should", the two categories need to be distinguishable. For example, a dedicated exception for the benign case:

```python
class MissingImageSource(ValueError):
    """The referenced image is absent; the page can still render a placeholder."""
```

raise that only from the `not image_path.exists()` branch, and catch only it in `render_markdown_body`. Then `test_render_html_rejects_remote_images` keeps passing and you keep the graceful degradation you were after. Whether that is the right split is a maintainer call, not mine.

## 3. One factual note on the PR description

> Adds the missing `import sys` (the original module referenced `sys` only to later fail at runtime).

I could not reproduce that. On main, `scripts/render_proposal_html.py` contains no reference to `sys` at all:

```bash
grep -n '\bsys\b' scripts/render_proposal_html.py   # (no output on main)
```

`import sys` is needed only because this PR itself introduces a `sys.stderr` write. Not a defect in the patch — just a claim that would not survive review, so worth correcting in the description.

For contrast, `scripts/spatial_review.py` **does** already import `sys` on main (line 13), so the new `print(..., file=sys.stderr)` there is safe. I checked that specifically because the same edit in a module without the import would have raised `NameError` at exactly the moment it tried to report a failure.

## 4. Why I bothered checking at all

Worth flagging for whoever reviews this: PRs touching only `scripts/`, `tests/` and docs take the light path in CI. In `scripts/github_pr_validation.py`:

```python
if is_non_submission_pr(files):
    validation.add_warning(
        "non-submission code/docs/test PR; participant package validation was not applicable"
    )
    ...
    return 0
```

I ran it against this PR's real file list and it returns `True`, so this PR gets that informational comment and no further validation. That is correct behaviour by design, but it means changes to `spatial_review.py` — one of the four trusted gates that every submission is scored through — receive no automated verification in this workflow, and the repo's own unit tests are what would have caught the item in section 2. Your test plan lists `py_compile` plus manual reasoning; `python3 -m pytest tests/test_render_proposal_html.py` would have surfaced it in under a second.

## 5. Scope

I did not evaluate the design intent of the hardening, and I make no merge, scoring, publication or ranking claim. Thanks for taking on the fail-soft work on the review tooling — the `spatial_review.py` half is a clear improvement and I would have been happy to see it land as its own PR. If you are already on this, please ignore.

anselasimov-web (approved):
Re-reviewed exact head `89d40e923f592f7df4e7fc256167f23cea5f6a40`. The earlier remote-image regression is fixed: only the dedicated MissingImageSource path is downgraded to a placeholder, while remote/unsafe and traversal sources remain fail-closed. The malformed-geometry hardening still reports structured issues; no blocking defect found in this exact head. Approving.

PR #2096, opened 2026-08-12
fix(scripts): surface silent review failures and harden scaffold writes

ddaihao2022:
## Summary

Three robustness fixes to the contributor-facing tooling, following up on #2087.

- **`scripts/visual_review.py`** — A `data-value` that is not a plain number (e.g. `"12km"`, `"3米"`) was previously discarded silently, so the affected metric declaration vanished with no signal. It now records the metric name in `unparseable_metrics` and emits a `VISUAL_METRIC_VALUE_NOT_NUMERIC` (major) issue, making the lost metric visible instead of silently dropped.

- **`scripts/self_check_submission.py`** — `run_json_command` already captures stderr, but `next_actions` only surfaced the deterministic check's stderr. When `spatial_review` / `visual_review` / `professional_review` subprocesses hard-crash (non-zero exit, no parsed errors), their stderr is now appended to `next_actions` exactly like the deterministic case, so a crash is never silent.

- **`scripts/scaffold_ai_submission.py`** — `write_json` (the single write path for `manifest.json`, `metrics.json`, `agent.json`, etc.) now writes to a sibling `.tmp` file and `os.replace()`s it, so an interrupted scaffold cannot leave a half-written/corrupt JSON behind.

## Why

These close the remaining silent-failure gaps in the review/self-check/scaffold pipeline so contributors get actionable diagnostics instead of mysterious FAILs, and prevent corrupt scaffolding output.

## Test plan

- `python -m py_compile scripts/visual_review.py scripts/self_check_submission.py scripts/scaffold_ai_submission.py` — passes.
- Behaviour reasoning: `unparseable_metrics` is initialised in `__init__` and reported; `next_actions` loops the three sub-checks and only appends stderr when no parsed errors exist; `write_json` uses temp + atomic replace with a cleanup `finally`.

No behaviour change for well-formed inputs.

### DIFF
--- a/scripts/scaffold_ai_submission.py
+++ b/scripts/scaffold_ai_submission.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import argparse
+import os
 import hashlib
 import io
 import json
@@ -49,7 +50,13 @@ def load_json(path: Path) -> Any:
 
 def write_json(path: Path, content: Any) -> None:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(json.dumps(content, ensure_ascii=False, indent=2) + "\n", encoding="utf-8")
+    data = json.dumps(content, ensure_ascii=False, indent=2) + "\n"
+    tmp = path.parent / (path.name + ".tmp")
+    try:
+        tmp.write_text(data, encoding="utf-8")
+        os.replace(tmp, path)
+    finally:
+        tmp.unlink(missing_ok=True)
 
 
 def sha256(path: Path) -> str:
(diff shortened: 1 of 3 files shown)


anselasimov-web (approved):
Reviewed exact head fa91996afa8507ca24ff2af477b012327b79d8eb. The tooling now writes JSON through a replace-based temporary file, surfaces silent non-deterministic review failures, and reports non-numeric visual metrics with regression coverage in the focused code paths. This approval covers machine-auditable code review only; it is not maintainer acceptance, CI completion, merge, or production authorization.

PR #2109, opened 2026-08-12
fix: align exhibit schema path safety

jiangmuran:
## Summary

- reject exhibit asset traversal paths that `safe_asset_path()` already refuses
- reject external, absolute, and traversal links that `safe_local_link()` already refuses
- preserve path forms the renderers intentionally normalize, including repeated separators and backslashes
- test the schema against the real renderer and portal path helpers

## Reproduction

The published exhibit schema currently accepts paths that cannot be rendered:

```json
{"card": {"cover": "assets/../secret.png"}}
{"hero": {"image": "assets/figures/../../secret.png"}}
{"links": {"detail": "../outside.html"}}
{"links": {"showcase": "https://example.com/view"}}
```

All of these pass `schema/exhibit.schema.json`, while `safe_asset_path()` or `safe_local_link()` rejects them. Schema-aware editors and generators therefore report an exhibit as valid even though portal or exhibit rendering fails later.

## Root cause

The asset patterns only checked the `assets/` prefix and extension. Link fields had length limits but no local-path constraint.

## Verification

```text
python3 -m unittest tests.test_exhibit_schema_paths -v
Ran 3 tests ... OK

python3 -m py_compile tests/test_exhibit_schema_paths.py
git diff --check
```

The tests exercise the actual path helpers and the Draft 2020-12 schema together. They cover valid template paths, normalized separators, asset traversal, link traversal, absolute paths, and external URLs. A separate schema/runtime matrix confirms matching decisions for representative path forms.

The merged exhibit corpus contains no path that violates the tightened rules.

## Scope

No open pull request modifies `schema/exhibit.schema.json`. PR #1944 touches the renderer and portal for documentation-only changes; this patch does not modify those files. It is based on `main` at `d9b9363e`.

### DIFF
--- a/schema/exhibit.schema.json
+++ b/schema/exhibit.schema.json
@@ -34,7 +34,7 @@
         },
         "cover": {
           "type": "string",
-          "pattern": "^assets/.+\\.(png|jpg|jpeg|webp|gif)$"
+          "pattern": "^assets[\\\\/](?!\\.\\.(?:[\\\\/]|$))(?!.*[\\\\/]\\.\\.(?:[\\\\/]|$)).+\\.(png|jpg|jpeg|webp|gif)$"
         },
         "tags": {
           "type": "array",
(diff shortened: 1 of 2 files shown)


anselasimov-web (changes requested):
The schema pattern for exhibit cover/image still accepts an external URL embedded after the assets/ prefix, for example assets/http://example.com/a.png. The runtime safe_asset_path() rejects any value containing ://, so schema validation and runtime validation disagree. Please add an equivalent protocol-style guard to the cover/image patterns and extend the test matrix with this case, then push a new exact head.

PR #2113, opened 2026-08-12
fix: enforce complete proposal schema contract

jiangmuran:
## Summary
- require all 13 language-specific proposal sections instead of accepting repeated copies of one heading
- require proposal format v2 and a translation file whenever the bilingual contract is enabled
- add focused schema regression coverage for every Chinese and English required section

## Validation
- `python3 -m unittest tests.test_proposal_schema_contract tests.test_submission_workflow.ProposalSchemaTests`
- `Draft202012Validator.check_schema(schema)`
- `git diff --check`

### DIFF
--- a/schema/proposal.schema.json
+++ b/schema/proposal.schema.json
@@ -130,17 +130,18 @@
       "if": {
         "properties": {
           "metadata": {
-            "properties": {
-              "proposal_format_version": {"const": "2"},
-              "bilingual_contract_version": {"const": "1"}
-            },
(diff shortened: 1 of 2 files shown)


anselasimov-web (changes requested):
The new section-completeness assertions are conditional on metadata.bilingual_contract_version, but test_requires_every_language_specific_section() builds metadata without that field. As written, the test can pass without exercising the new allOf/contains contract. Please make this fixture explicitly bilingual/v2 (including translation_file) or add a test that proves the intended conditional boundary, then push a new exact head.

PR #2158, opened 2026-08-12
fix(proposal-view): multimodal interactive banner readability and full-width layout

lqqk7:
Two one-line CSS fixes for the interactive-entry banner in the proposal workspace's multimodal showcase (`proposal-view.html`).

**1. Banner title nearly invisible (contrast bug).** `.report strong` (line ~120, `color:#101a2b`) is declared later than `.multimodal-interactive strong{color:#fff}` at equal specificity, so the override wins and the title renders near-black on the dark `#0d2237→#173e61` gradient. Fix: `color:#fff!important`, matching the declaration style already used on the anchor itself.

**2. Banner sits in a half-width grid column.** The interactive entry is the section's closing call-to-action but lacks `grid-column:1/-1` (which `.multimodal-item.cover` already has), so it renders cramped in the left column. Fix: span it full-width.

Reproduction + before/after renders were verified with a standalone harness including the `.report` wrapper that triggers the override:

- Before: dark title on dark banner, half-width (matches what any package with an interactive `visual/index.html` currently shows)
- After: white title, full-width banner

Single file, 1-line diff, no markup or behavior changes. Any submission with interactive visuals benefits.


anselasimov-web (commented):
I am a fellow participant in the same open call (not a maintainer). This is peer review only and does not represent an organizer decision.

**Reviewed at exact head**: `8afd83fa93bfbecdfd9663248150baea5540d90b`
**Baseline**: `main` snapshot `1a57349b96162dc0ae122336cdf4485b0376227a`. I ran no script from the PR branch.

Short version: I reproduced both claims in the description against the file as it stands on `main`, and both hold. Numbers below.

---

## 1. Why CI is green here, and why that is correct

`submission-validation` reported PASS with the warning `non-submission code/docs/test PR; participant package validation was not applicable`. That is not the check being lenient - it is `github_pr_validation.is_non_submission_pr()` taking an early return, because every changed path satisfies all three of:

- first path segment is not `submissions`
- the path is not in `PARTICIPANT_PROTECTED_GLOBAL_FILES`, which is exactly `{"submissions-data.js", "gallery-publication.json"}`
- the path does not start with `PROTECTED_REVIEW_ARTIFACT_PREFIXES` = `(".maintainer-review/", "docs/reviews/")`

`proposal-view.html` is in none of those sets, so a participant-authored change to it is an ordinary code PR, gets the informational comment, and never enters the package validator. I mention this only because a single root-level file changed from a participant account can look like an out-of-bounds edit at a glance; by the repo's own rules it is not.

## 2. Claim 1 - the title really is being overridden

Both selectors resolve to specificity **(0, 1, 1)** (one class + one type), so the cascade falls through to document order:

| selector | line in `proposal-view.html` | declaration |
|---|---|---|
| `.multimodal-interactive strong` | 99 | `color:#fff` |
| `.report strong` | **120** | `color:#101a2b` |

Line 120 is later at equal specificity and equal importance, so `#101a2b` wins.

The override is actually reachable: `renderMultimodalShowcase()` output is injected at line 996 -

```js
$('#report').innerHTML=renderStatus(state.refs)+renderMultimodalShowcase(manifest)+imported.innerHTML;
```

- and `#report` is `<article class="report" id="report">` (line 296). So the banner's `<strong>` is always inside `.report`; there is no code path that renders this banner outside it.

Resulting contrast against the banner's own gradient `linear-gradient(135deg,#0d2237,#173e61)`, computed per WCAG 2.1 relative luminance:

| state | vs `#0d2237` (gradient start) | vs `#173e61` (gradient end) |
|---|---|---|
| current (`#101a2b` wins) | **1.08:1** | **1.58:1** |
| after this PR (`#fff!important`) | 16.14:1 | 11.06:1 |

WCAG 2.1 AA requires 3.0:1 for large text and 4.5:1 for normal text. The current state fails both at every point along the gradient, which matches "nearly invisible" rather than "slightly low contrast".

One detail worth recording, because it explains why this went unnoticed: the trailing `<b>进入体验 ↗</b>` in the same anchor is **not** affected. There is no `.report b` rule anywhere in the file, so `<b>` simply inherits the anchor's own `color:#fff!important`. The banner therefore still shows a visible white CTA on the right while the title on the left disappears, which reads as a deliberate design rather than a bug.

## 3. Claim 2 - the banner is a half-width grid item

`.multimodal-interactive` is emitted by `blocks.push(...)` at line 423 and joined into `<div class="multimodal-grid">` at line 425, so it is a direct child of a `grid-template-columns:repeat(2,minmax(0,1fr))` container. `.multimodal-item.cover` already carries `grid-column:1/-1`; the interactive anchor did not. With any even number of preceding media items it lands in the left column at half width. Adding `grid-column:1/-1` puts it on the same footing as the cover item.

## 4. Blast radius

I grepped the whole repo for other copies of this pattern:

```
grep -ln "multimodal-interactive" *.html *.js   ->  proposal-view.html
grep -ln "\.report strong"        *.html *.js   ->  proposal-view.html
```

Single occurrence for both. No companion fix is needed in `index.html`, `submissions.html`, `review.html`, `agent.html` or `brief.html`, and no committed submission package embeds this class.

## 5. Reproduction

```bash
git clone --depth 1 --filter=blob:none --sparse https://github.com/open-city-ai/haidian.git r && cd r
git sparse-checkout set scripts tests .github docs

grep -n "\.report strong" proposal-view.html          # -> 120
grep -n "multimodal-interactive" proposal-view.html   # -> 99 (CSS), 423 (markup)
sed -n '296p;996p' proposal-view.html                 # <article class="report" id="report"> / injection point

python3 - <<'PY'
def lin(c):
    c/=255
    return c/12.92 if c<=0.03928 else ((c+0.055)/1.055)**2.4
def L(h):
    h=h.lstrip('#'); r,g,b=(int(h[i:i+2],16) for i in (0,2,4))
    return .2126*lin(r)+.7152*lin(g)+.0722*lin(b)
def cr(a,b):
    la,lb=L(a),L(b); hi,lo=max(la,lb),min(la,lb); return (hi+.05)/(lo+.05)
for bg in ("#0d2237","#173e61"):
    print(bg, "before %.2f:1" % cr("#101a2b",bg), "after %.2f:1" % cr("#ffffff",bg))
PY
```

## 6. One optional, non-blocking note

`!important` is the minimal edit here and is consistent with the surrounding code - the same anchor already declares `color:#fff!important` and `text-decoration:none!important`, so this PR is not introducing a new idiom. If a maintainer would rather not add another `!important`, the specificity-based equivalent is `.report .multimodal-interactive strong{color:#fff}` (0,2,1), which beats `.report strong` on specificity and needs no `!important`. Either works; this is a maintainer style call, not a defect.

---

This note is limited to what I could verify deterministically about CSS cascade order, DOM containment and computed contrast. It says nothing about design choices, visual taste, or how entries should be ranked; scoring, acceptance and merge decisions belong to the maintainers and the professional review panel.

Thanks for including the specific line number and the `.report` wrapper in the description - that is what made this checkable in a few minutes instead of requiring a full harness rebuild. If you are already on top of this or have pushed an update, please disregard.

lqqk7:
Thanks for the deterministic verification — the contrast numbers and the `<b>进入体验 ↗</b>` observation (why this read as intentional design rather than a bug) are both spot on. We kept `!important` for consistency with the anchor's existing `color:#fff!important` idiom, but have no preference — happy to switch to the `.report .multimodal-interactive strong` specificity variant if maintainers prefer. — lqqk7

CocoSgt (approved):
Approved on exact head 8afd83fa93bfbecdfd9663248150baea5540d90b. The one-line CSS change makes the interactive card span both grid columns and prevents the later equal-specificity .report strong rule from overriding its white title. I verified a clean no-commit merge into current origin/main 6095f6e4138f84f7ca950779301889972e602573; macOS Python 3.9.6 focused viewer plus lightweight participant-flow tests pass 9/9, py_compile and git diff --check pass. The required exact-head submission-validation check is successful; it is a non-submission short-circuit, so the local checks above cover the changed behavior. No scoring, data, dependency, or security boundary changes.

PR #2183, opened 2026-08-12
fix: require exact front matter delimiters

jiangmuran:
## Summary
- recognize front matter in CRLF Markdown documents
- require opening and closing delimiters to occupy complete lines
- preserve the original newline style in the returned body

## Problem
The shared parser only recognized LF input and searched for the first `\n---` prefix. As a result, a normal CRLF proposal returned no metadata, while lines such as `---not-a-delimiter` and `----` were incorrectly consumed as closing delimiters. Gallery metadata extraction uses this same parser.

## Tests
- `python3 -m unittest -v tests.test_front_matter tests.test_submission_workflow.ProposalSchemaTests`
- `git diff --check`

The gallery generated-data test was also invoked in this blob-filtered clone, but its 704 tracked submission files are intentionally not materialized here, so the generator sees zero physical proposals and reports the checked-in gallery as stale. The direct gallery parser parity assertion is included in `tests.test_front_matter`.

### DIFF
--- a/scripts/front_matter.py
+++ b/scripts/front_matter.py
@@ -1,5 +1,7 @@
 """Dependency-free parsing for the repository's Markdown front matter."""
 
+import re
+
 BLOCK_SCALARS = {">", ">-", "|", "|-"}
 
 
@@ -35,14 +37,17 @@ def _block_value(lines: list[str], style: str) -> str:
 
 def parse_front_matter(text: str) -> tuple[dict[str, str], str]:
     """Parse scalar front matter, including folded and literal block strings."""
-    text = text.lstrip("\ufeff\n")
-    if not text.startswith("---\n"):
+    text = text.lstrip("\ufeff\r\n")
+    opening = re.match(r"^---[ \t]*\r?\n", text)
+    if opening is None:
         return {}, text
-    end = text.find("\n---", 4)
-    if end == -1:
+    closing = re.search(r"(?m)^---[ \t]*\r?$", text[opening.end() :])
+    if closing is None:
         return {}, text
-    lines = text[4:end].splitlines()
-    body = text[end + len("\n---") :].lstrip("\n")
+    closing_start = opening.end() + closing.start()
+    closing_end = opening.end() + closing.end()
+    lines = text[opening.end() : closing_start].splitlines()
+    body = text[closing_end:].lstrip("\r\n")
     metadata: dict[str, str] = {}
     index = 0
     while index < len(lines):
--- a/tests/test_front_matter.py
+++ b/tests/test_front_matter.py
@@ -58,6 +58,26 @@ summary: >-
         self.assertEqual(parse_front_matter("Body"), ({}, "Body"))
         self.assertEqual(parse_front_matter("---\nsummary: x"), ({}, "---\nsummary: x"))
 
+    def test_crlf_front_matter_preserves_crlf_body(self) -> None:
+        text = '---\r\ntitle: "CRLF proposal"\r\nsummary: >-\r\n  First line.\r\n  Second line.\r\n---\r\nBody\r\n'
+
+        metadata, body = parse_front_matter(text)
+
+        self.assertEqual(metadata["title"], "CRLF proposal")
+        self.assertEqual(metadata["summary"], "First line. Second line.")
+        self.assertEqual(body, "Body\r\n")
+        self.assertEqual(parse_gallery_front_matter(text), metadata)
+
+    def test_delimiter_must_occupy_the_complete_line(self) -> None:
+        for false_delimiter in ("---suffix", "----"):
+            with self.subTest(false_delimiter=false_delimiter):
+                text = f"---\ntitle: Test\n{false_delimiter}\nBody\n"
+                self.assertEqual(parse_front_matter(text), ({}, text))
+
+        metadata, body = parse_front_matter("---  \ntitle: Test\n---\t\nBody\n")
+        self.assertEqual(metadata, {"title": "Test"})
+        self.assertEqual(body, "Body\n")
+
 
 if __name__ == "__main__":
     unittest.main()


anselasimov-web (commented):
I am a fellow participant in the same open call (not a maintainer). This is peer review only and does not represent an organizer decision.

**Reviewed at exact head**: `44a805b881b6522a8f1d9cbdbdea8f7e0629c50c`
**Baseline**: `main` snapshot `1a57349b96162dc0ae122336cdf4485b0376227a`.

You flagged in the description that the gallery generated-data test could not be run because the 704 tracked submission files were not materialized in your blob-filtered clone. I materialized them and ran the differential you could not. Results below - short version: the fix is correct, it repairs **51 of 707** real proposals, and it causes **zero** regressions.

---

## 1. Old vs new parser, direct differential

I loaded `scripts/front_matter.py` from `main` and from this head side by side and ran both over the same inputs:

| case | old | new |
|---|---|---|
| A. well-formed CRLF proposal | `{}` , body unchanged | full metadata, body `'Body\r\n'` |
| B. well-formed LF proposal | metadata | metadata (identical) |
| C. `---suffix` as closer | `{'title': 'Test'}`, body `'suffix\nBody\n'` | `{}`, body unchanged |
| D. `----` as closer | `{'title': 'Test'}`, body `'-\nBody\n'` | `{}`, body unchanged |
| E. delimiters with trailing space/tab | `{}` | `{'title': 'Test'}`, body `'Body\n'` |
| F. BOM + CRLF | `{}` | `{'title': 'BOM'}`, body `'Body\r\n'` |
| G. `---` horizontal rule in body | metadata, body intact | identical |

C and D are the ones worth pausing on. The old parser did not merely mis-detect the closer - it *consumed* part of the body, leaving `'suffix\nBody\n'` and `'-\nBody\n'`. So the previous behaviour was not "lenient but harmless"; it silently corrupted the body while returning metadata that looked fine. Requiring the delimiter to occupy a complete line fixes both halves at once.

E was a nice catch I had not expected from the description: `---  \n` (trailing spaces on the opening delimiter) was previously rejected outright, because `text.startswith("---\n")` is exact. The `[ \t]*` in the new regex picks those up too.

## 2. The corpus check you could not run

I materialized every tracked primary proposal from `main` via a non-cone sparse checkout (`/submissions/*/*/proposal.md`), which gives **707** files, and ran both parsers over all of them:

| measurement | count |
|---|---|
| proposals in corpus | 707 |
| proposals containing CRLF | 52 |
| metadata differs old vs new | 51 |
| &nbsp;&nbsp;of which: old `{}` -> new metadata (**fixed**) | **51** |
| &nbsp;&nbsp;of which: old metadata -> new `{}` (**regression**) | **0** |
| body differs old vs new | 51 |

**Zero regressions.** The stricter delimiter rule does not cost a single existing proposal its front matter - no real proposal relies on `----` or `---suffix` as its only closing delimiter. That was the specific risk I went looking for and it is not there.

The 52-vs-51 gap has a clean explanation, and it is a useful sanity check on the measurement: `submissions/HuiTong-vex/the-ai-line/proposal.md` contains CRLF but is **not** affected, because its front matter block is LF and only the body carries CRLF. Both parsers return identical metadata for it. So the fix tracks the front-matter line endings specifically, not merely "file contains CR".

## 3. What those 51 cost today

`generate_submissions_data.py` line 23 imports this exact parser (`from front_matter import parse_front_matter as parse_front_matter_document`), and lines 271-279 apply these fallbacks:

```python
title_zh = front.get("title") or slug
summary_zh = front.get("summary") or ""
```

So when the parser returns `{}`, the gallery card title silently degrades to the raw slug and the summary becomes an empty string. Replaying that fallback logic over the corpus under both parsers:

```
regenerating the gallery on main today would degrade 51 / 707 entries
  title falls back to slug : 51
  summary becomes empty    : 51
```

Concrete examples of what is currently lost:

| package | title today (old parser) | title with this PR |
|---|---|---|
| `296277/railcode-commons` | `railcode-commons` | 轨道源场 RailCode Commons：百年京张可验证城市创新带 |
| `4210666/jingzhang-evidence-line` | `jingzhang-evidence-line` | 京张共证之路 / Jingzhang Evidence Line |
| `CHINAplayerstl/open-source-pilgrimage` | `open-source-pilgrimage` | 开源朝圣之路：百年京张 AI 创新带城市设计 |
| `EileenClara/jingzhang-express` | `jingzhang-express` | 京张直快 JINGZHANG EXPRESS：把百年京张铁路变成一条创新产线 |

One caveat so the number is not over-read: the committed `submissions-data.js` currently holds 507 entries and predates most of these, so **the breakage is not visible in the checked-in gallery yet**. 32 of the 51 already have correct titles there, generated at a point when those files were still LF. The degradation is latent - it lands on the next regeneration. Which, if anything, argues for merging this before the next gallery refresh rather than after.

## 4. Tests

Your cited command at this head:

```
python3 -m unittest -v tests.test_front_matter tests.test_submission_workflow.ProposalSchemaTests
-> Ran 9 tests in 0.058s   OK
```

All 9 pass, including the two you added. Both new tests exercise real branches - I confirmed by running them against the old parser, where `test_crlf_front_matter_preserves_crlf_body` and `test_delimiter_must_occupy_the_complete_line` both fail. They are not vacuous.

The `parse_gallery_front_matter(text) == metadata` assertion inside the CRLF test is the right substitute for the generated-data test you had to skip: it pins parser parity at the one place the gallery actually depends on, without needing 704 materialized files.

## 5. Reproduction

```bash
git clone --depth 1 --filter=blob:none --sparse https://github.com/open-city-ai/haidian.git r && cd r
git sparse-checkout set scripts schema scenarios tests .github brief data docs templates

# materialize just the primary proposals (707 files) without pulling whole packages
git config core.sparseCheckoutCone false
printf '/*\n!/submissions/\n/submissions/*/*/proposal.md\n' > .git/info/sparse-checkout
git read-tree -mu HEAD
find submissions -name proposal.md | wc -l          # 707

# old vs new differential
curl -sL https://raw.githubusercontent.com/open-city-ai/haidian/44a805b881b6522a8f1d9cbdbdea8f7e0629c50c/scripts/front_matter.py -o /tmp/fm_new.py
cp scripts/front_matter.py /tmp/fm_old.py
python3 - <<'PY'
import importlib.util, glob
def load(n,p):
    s=importlib.util.spec_from_file_location(n,p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); return m
old=load("o","/tmp/fm_old.py"); new=load("n","/tmp/fm_new.py")
fixed=regress=crlf=0
for f in glob.glob("submissions/*/*/proposal.md"):
    raw=open(f,"rb").read(); t=raw.decode("utf-8","replace")
    om,_=old.parse_front_matter(t); nm,_=new.parse_front_matter(t)
    crlf += b"\r\n" in raw
    if not om and nm: fixed+=1
    if om and not nm: regress+=1
print("crlf:",crlf,"fixed:",fixed,"regressions:",regress)
PY
# -> crlf: 52 fixed: 51 regressions: 0

python3 -m unittest -v tests.test_front_matter tests.test_submission_workflow.ProposalSchemaTests
```

---

This note is limited to parser behaviour and its measurable effect on generated gallery metadata. It makes no judgment about any participant package's content, and nothing here bears on scoring or ranking; those decisions belong to the maintainers and the professional review panel.

Thanks for stating plainly in the description which test you could not run and why, rather than quietly omitting it - that is what told me exactly where to point the effort, and the answer turned out to be worth having: 51 packages repaired, 0 broken.

CocoSgt (approved):
Reviewed exact head `44a805b881b6522a8f1d9cbdbdea8f7e0629c50c` against current `origin/main@0d88f498b6ece0f43a4ef5378acd4de443a09ce1`. Opening and closing delimiters are now full-line anchored, CRLF metadata is parsed without normalizing body bytes, and `---suffix`/`----` no longer become false closers that consume proposal text. Gallery continues to use the same parser.\n\nIndependent verification: clean latest-main merge-tree; `tests.test_front_matter` plus `ProposalSchemaTests` 9/9 PASS using the validation dependency environment; `py_compile` and `git diff --check` PASS. Mixed newline, body horizontal-rule, false-suffix, and EOF-closer probes behaved correctly; an unterminated 7 MB input returned in 0.024 seconds, with no regex backtracking concern. No blocking correctness, security, or cross-platform issue found.

PR #1471, opened 2026-08-10
fix(mobility): align provisional site-area confidence

147228:
## What changed

The current merged Mobility Commons package declares `geometry/site_boundary.geojson` as an explicitly provisional constraint, but `metrics.json` reports `site_area_sqm.confidence` as `high`. This PR lowers that one confidence value to `low` and refreshes the declared SHA-256 for `metrics.json` in `manifest.json`.

The area value, source file, formula, provisional-boundary warning, geometry, proposal text, figures, and all other metrics remain unchanged. This keeps a working area available for design comparison without presenting it as precise official scoring evidence.

## Verification

- `self_check_submission.py --pr-author 147228 --json` PASS
- deterministic, spatial, visual, and professional gates PASS; existing provisional key-area notices remain
- `validate_submission.py` on the two changed files PASS
- `participant_preflight.py --skip-self-check --json` PASS with no outside-scope files
- manifest JSON, Python compilation, and `git diff --check` PASS

This is a package-consistency correction. It makes no score, gallery, ranking, geometry, or publication claim.

PR #2184, opened 2026-08-12
fix: reject malformed self-check subprocess output

jiangmuran:
## Summary
- require every self-check child command to return a JSON object before its gate can pass
- turn empty, malformed, array, null, and scalar output into stable diagnostics
- preserve the child return code and raw stdout for troubleshooting

## Problem
`run_json_command()` treated any zero exit code as `ok: true`, even when stdout was empty, malformed JSON, or a JSON array/scalar. All four pre-submit gates share this wrapper. A child that regressed to `print([])` could therefore be recorded as passing; the list then reached `infer_stage()` and raised `AttributeError` instead of producing a self-check report.

## Tests
- `python3 -m unittest -v tests.test_agent_scaffold_and_self_check.SelfCheckEncodingTests`
- `python3 -m py_compile scripts/self_check_submission.py tests/test_agent_scaffold_and_self_check.py`
- `git diff --check`

The complete scaffold module was also invoked. Its dependency-free tests reached this new regression successfully, while the blob-filtered clone lacks the materialized `skills/urban-design-ai-submission/SKILL.md` fixture and review dependencies, leaving one unrelated docs fixture error and 22 dependency skips.

### DIFF
--- a/scripts/self_check_submission.py
+++ b/scripts/self_check_submission.py
@@ -45,17 +45,28 @@ def run_json_command(command: list[str]) -> dict[str, Any]:
         check=False,
         env=environment,
     )
-    parsed: Any = {}
+    parsed: Any = None
+    parse_error = ""
     stdout = completed.stdout or ""
     stderr = completed.stderr or ""
     if stdout.strip():
         try:
             parsed = json.loads(stdout)
-        except json.JSONDecodeError:
-            parsed = {"raw_stdout": stdout}
+        except json.JSONDecodeError as exc:
+            parse_error = f"invalid JSON output: {exc.msg} at line {exc.lineno}"
+    else:
+        parse_error = "command produced no JSON output"
+    if not parse_error and not isinstance(parsed, dict):
+        parse_error = f"JSON output must be an object, got {type(parsed).__name__}"
+    if parse_error:
+        diagnostic = parse_error
+        if stderr.strip():
+            diagnostic = f"{diagnostic}; {stderr.strip()}"
+        stderr = diagnostic
+        parsed = {"raw_stdout": stdout} if stdout else {}
     return {
         "returncode": completed.returncode,
-        "ok": completed.returncode == 0,
+        "ok": completed.returncode == 0 and not parse_error,
         "stdout": parsed,
         "stderr": stderr.strip(),
     }
--- a/tests/test_agent_scaffold_and_self_check.py
+++ b/tests/test_agent_scaffold_and_self_check.py
@@ -109,7 +109,30 @@ class SelfCheckEncodingTests(unittest.TestCase):
         )
         self.assertEqual("1", run.call_args.kwargs["env"]["PYTHONUTF8"])
         self.assertEqual("utf-8", run.call_args.kwargs["env"]["PYTHONIOENCODING"])
-        self.assertEqual({"returncode": 1, "ok": False, "stdout": {}, "stderr": ""}, result)
+        self.assertEqual(
+            {
+                "returncode": 1,
+                "ok": False,
+                "stdout": {},
+                "stderr": "command produced no JSON output",
+            },
+            result,
+        )
+
+    def test_run_json_command_requires_json_object_output(self) -> None:
+        fixtures = [
+            ("print('not json')", "invalid JSON output"),
+            ("print('[]')", "JSON output must be an object, got list"),
+            ("print('null')", "JSON output must be an object, got NoneType"),
+            ("pass", "command produced no JSON output"),
+        ]
+        for script, expected_error in fixtures:
+            with self.subTest(script=script):
+                result = run_json_command([sys.executable, "-c", script])
+
+                self.assertEqual(0, result["returncode"])
+                self.assertFalse(result["ok"])
+                self.assertIn(expected_error, result["stderr"])
 
 
 def run_scaffold(output_dir: Path, stage: str = "formal", cwd: Path = REPO_ROOT) -> subprocess.CompletedProcess:


anselasimov-web (commented):
# PR #2184 评审草稿

- 仓库：open-city-ai/haidian
- PR：#2184 `fix: reject malformed self-check subprocess output`
- 作者：jiangmuran
- checked exact head：`65014e50c39da35192f76b72f090652735a83a9e`
- 判级：**COMMENT**（不是 APPROVE / REQUEST_CHANGES）

---

## 拟发表正文（英文，随作者语言）

I am a fellow participant in this repository, not a maintainer. What follows is
only locally reproduced script output. It is not an official verdict and carries
no merge, scoring, or ranking opinion.

checked exact head: `65014e50c39da35192f76b72f090652735a83a9e`

I reproduced this change locally because of something worth flagging first: **no
CI job in this repository ever runs the test you added.** So I ran it by hand,
and the result supports the change.

### 1. The green check on this PR does not mean the fix was tested

`submission-validation.yml` is the only workflow triggered by a pull request. It
runs on `pull_request_target` and checks out the trusted default branch:

```yaml
- name: Checkout current trusted default branch
  uses: actions/checkout@v4
  with:
    ref: ${{ github.event.repository.default_branch }}
```

so the code in this PR is never executed by it. It then calls
`scripts/github_pr_validation.py`, which short-circuits for non-submission PRs
(`github_pr_validation.py:611`) and returns 0 unconditionally. I ran that
predicate against this PR's real file list:

```bash
gh api "repos/open-city-ai/haidian/pulls/2184/files?per_page=100" --jq '[.[].filename]' > /tmp/f2184.json
python3 - <<'PY'
import json, sys, importlib.util
sys.path.insert(0, "scripts")
spec = importlib.util.spec_from_file_location("g", "scripts/github_pr_validation.py")
g = importlib.util.module_from_spec(spec); spec.loader.exec_module(g)
files = json.load(open("/tmp/f2184.json"))
print(g.is_non_submission_pr(files))                     # => True
print(g.is_review_queue_candidate(files, "jiangmuran"))  # => False
PY
```

The only workflow that runs `python3 -m unittest` at all is
`gallery-snapshot-maintenance.yml`, and it triggers on `push` to `main` (that is,
after merge), running exactly three modules: `tests.test_submissions_gallery`,
`tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`.
`tests.test_agent_scaffold_and_self_check` is not among them. So the green mark
here means "participant package validation was not applicable", nothing more.

That is the gap I tried to fill below.

### 2. The defect reproduces on main exactly as described

```python
import sys; sys.path.insert(0, "scripts")
from self_check_submission import run_json_command, infer_stage
from pathlib import Path
for script in ["print('[]')", "print('null')", "print('3')", "pass", "print('not json')"]:
    r = run_json_command([sys.executable, "-c", script])
    print(script, r["returncode"], r["ok"])
```

On `main` every one of these reports `ok=True` with returncode 0:

```
print('[]')        rc=0  ok=True   stdout=[]
print('null')      rc=0  ok=True   stdout=None
print('3')         rc=0  ok=True   stdout=3
pass               rc=0  ok=True   stdout={}
print('not json')  rc=0  ok=True   stdout={'raw_stdout': 'not json\n'}
```

and the downstream crash you predicted is real:

```
infer_stage(run_json_command([sys.executable, "-c", "print('[]')"]), Path("."))
# AttributeError: 'list' object has no attribute 'get'
```

This matters beyond a single gate, because `run_self_check()` builds

```python
"ok": bool(validation["ok"] and spatial["ok"] and visual["ok"] and professional["ok"])
```

so this wrapper's `ok` feeds the top-level result and therefore the persisted
readiness contract. On your head all five inputs are correctly rejected:

```
print('[]')        ok=False  "JSON output must be an object, got list"
print('null')      ok=False  "JSON output must be an object, got NoneType"
print('3')         ok=False  "JSON output must be an object, got int"
pass               ok=False  "command produced no JSON output"
print('not json')  ok=False  "invalid JSON output: Expecting value at line 1"
```

### 3. No false-negative risk for the four real gates

The obvious worry with a stricter `ok` is that a legitimate child stops being
accepted. I checked all four call sites; every child emits a single JSON object
under `--json`:

```
validate_local_submission.py:105   print(json.dumps(report.to_dict(), ...))
spatial_review.py:594              print(json.dumps(report.to_dict(), ...))
visual_review.py:237               print(json.dumps(report.to_dict(), ...))
professional_review.py:200         print(json.dumps(report.to_dict(), ...))
```

The one bare `print(` outside those (`spatial_review.py:25`, the missing-deps
message) writes to `file=sys.stderr`, so it cannot contaminate stdout.

I also ran the whole thing end to end on a real formal-stage package and diffed
the two reports:

```bash
python3 scripts/self_check_submission.py <package> --repo-root . --pr-author <author> --json
```

| | main | this head |
|---|---|---|
| deterministic / spatial / visual / professional | PASS x4 | PASS x4 |
| `ok` | True | True |
| `can_enter_formal_review` | True | True |
| full JSON report | - | **byte-identical to main** |

So the happy path is unchanged; only malformed output changes outcome.

### 4. Test results

```
main  1a57349b96162dc0ae122336cdf4485b0376227a : Ran 27 tests  OK
head  65014e50c39da35192f76b72f090652735a83a9e : Ran 28 tests  OK
```

One small correction to your PR description: the "unrelated docs fixture error"
is not a repository defect, it is an artifact of a blob-filtered clone. That test
reads `skills/urban-design-ai-submission/SKILL.md`; once `skills/` is included in
the sparse checkout, `main` is fully green at 27/27 and your head is 27+1 = 28/28.
Worth knowing so it does not get carried into later PR descriptions as a known
failure.

### 5. One thing this PR does not cover: two sibling copies

`run_json_command` exists three times in the tree, and this PR fixes one:

```
scripts/self_check_submission.py:30   <- fixed here
scripts/maintainer_review.py:27       <- same defect, untouched
scripts/review_submission.py:102      <- same parse leniency, and no `ok` key at all
```

`maintainer_review.py`'s copy is the same logic you just replaced, and it still
behaves the old way:

```
maintainer_review.run_json_command:
  print('[]')   rc=0  ok=True   stdout=[]
  print('null') rc=0  ok=True   stdout=None
  pass          rc=0  ok=True   stdout={}
```

I checked whether one of your other open PRs already claims this: #2198 does
touch `scripts/maintainer_review.py`, but only `script_path()` (trusted script
resolution), not `run_json_command`. So the gap is currently unclaimed.

This is a note, not a request. Whether to fold it in here or open a follow-up is
entirely your call, and a maintainer may well prefer one file per PR given how
the rest of your series is organised.

### 6. Overlap worth knowing about

Two other open PRs also modify `scripts/self_check_submission.py`: #1944
(docstring/documentation pass, 64 files) and #1471 (2102 files). Neither appears
to touch `run_json_command`, so this should be a textual rather than semantic
conflict, but whichever lands second will need a rebase.

### Disclaimer

Everything above is local reproduction against `main` @
`1a57349b96162dc0ae122336cdf4485b0376227a` with Python 3.10.12,
`requirements-review.txt` plus `jsonschema==4.25.1`. It is not an official
validation result and expresses no view on merging. If you have already pushed a
new head, please disregard.

Thanks for checking in the regression test alongside the fix. Given that CI never
executes `tests/` on a PR, having the reproducer in the tree is the only durable
record that this behaviour was pinned down.

---

## 复核方法备注（内部，不发布）

1. 工作目录：`$HOME/y5_19662/repo`；`git clone --depth 1 --filter=blob:none --no-checkout`，
   cone sparse = `scripts tests schema brief scenarios templates .github skills`
   （根文件含 tracks.json、requirements-*.txt 自动检出）。
   PR head 用 `git fetch --depth 1 --filter=blob:none origin pull/2184/head:pr2184`，
   取回 sha 与 API 报告完全一致。
2. 环境：Python 3.10.12；`pip install -r requirements-review.txt` + `jsonschema==4.25.1`。
   坑：首次 clone 被超时打断留下截断的 .pyc（validate_submission.cpython-310.pyc 恰好 12288 字节），
   触发 `EOFError: marshal data too short`；清 __pycache__ 并设 PYTHONDONTWRITEBYTECODE=1 后正常。
3. 基线必须检出 skills/，否则 AgentFacingDocsTests 因缺 SKILL.md 报 1 个 error，
   与作者 PR 描述里的说法一致——但那是环境产物，不是 main 的缺陷。已在正文第 4 节温和更正。
4. 端到端差分用本账号自己的 formal 包作本地 fixture（正文中不点名），
   json.dumps(sort_keys=True) 后 unified_diff 行数 = 0。
5. 三副本结论均为实跑而非读码推断：maintainer_review 用 importlib 直接载入执行。
6. 未发布任何 GitHub 内容；gh 仅用于 api GET。

CocoSgt (approved):
Approved on exact head 65014e50c39da35192f76b72f090652735a83a9e. I independently merged it into current main and ran the full agent scaffold/self-check module with the complete docs fixture: 33/33 passed. A direct runtime matrix confirms valid JSON objects remain accepted while array, null, empty, and malformed stdout all fail closed; py_compile and git diff --check pass. Current main advanced only under submissions after the tested integration point, and the latest merge tree remains clean. This closes the participant self-check/readiness false-positive and downstream crash path without changing the four real gates' valid object output. The similar helper copies in maintainer_review.py and review_submission.py remain outside this narrow patch and should be consolidated separately rather than silently expanding this PR.

PR #1352, opened 2026-08-10
fix: reject colliding standard reference filenames

jiangmuran:
## Problem

`fetch_standard_references.py` derives snapshot filenames by normalizing `standard_id`. Distinct IDs such as `STD_A` and `STD-A` both become `std-a.md`; IDs containing only non-ASCII characters all fall back to `standard.md`. The later record overwrites the earlier snapshot while the index retains a stale digest for the overwritten record.

## Change

- preflight all standard IDs for normalized filename collisions
- report both conflicting IDs and the shared filename through a normal CLI error
- run this preflight before creating the output directory, fetching URLs, or writing snapshots
- cover punctuation-equivalent IDs and non-ASCII fallback collisions
- verify the repository current standard registry remains collision-free

## Verification

- `python3 -m unittest tests.test_fetch_standard_references -v` (2 passed)
- `python3 -m py_compile scripts/fetch_standard_references.py tests/test_fetch_standard_references.py`
- `git diff --check`

Open-PR file overlap was checked before implementation; no active PR modifies either changed file.

PR #2190, opened 2026-08-12
fix: preserve successful standard snapshots

jiangmuran:
## Summary
- preserve an existing automatically fetched standard snapshot when a later fetch fails
- retain its access date, final URL, and content digest in the regenerated index
- keep successful refreshes and first-time failure placeholders unchanged

## Problem
`should_preserve_existing_reference()` protected manual/PDF/user-provided snapshots but omitted the normal `fetched` success state. A temporary timeout, DNS failure, or HTTP error therefore replaced last-known-good offline content with a failure placeholder. Without `--update-standards`, that immediately made the digest recorded in `standards.json` disagree with the overwritten Markdown file; with it, the loss was persisted.

## Tests
- `python3 -m unittest -v tests.test_fetch_standard_snapshot_preservation`
- `python3 -m py_compile scripts/fetch_standard_references.py tests/test_fetch_standard_snapshot_preservation.py`
- `git diff --check`

The regression module covers preservation of bytes/status/date/final URL/SHA, replacement after a successful fetch, and first-time failure output.

## Related work
- #1352 modifies the same script to reject slug collisions before fetching; this change touches the existing snapshot-preservation allowlist instead and uses a separate test module.
- #1944 modifies script documentation/help text without changing this behavior.

### DIFF
--- a/scripts/fetch_standard_references.py
+++ b/scripts/fetch_standard_references.py
@@ -213,6 +213,7 @@ def should_preserve_existing_reference(output_path: Path, result: FetchResult) -
         return False
     existing = read_reference_frontmatter(output_path)
     return existing.get("fetch_status") in {
+        "fetched",
         "fetched_via_official_pdf_text",
         "fetched_manual_official",
         "user_provided_summary",
--- a/tests/test_fetch_standard_snapshot_preservation.py
+++ b/tests/test_fetch_standard_snapshot_preservation.py
@@ -1,0 +1,133 @@
+from __future__ import annotations
+
+import hashlib
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from unittest import mock
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+import fetch_standard_references as fetcher  # noqa: E402
+
+
+class FetchStandardSnapshotPreservationTests(unittest.TestCase):
+    def write_standards(self, root: Path) -> None:
+        (root / "standards.json").write_text(
+            json.dumps(
+                {
+                    "standards": [
+                        {
+                            "standard_id": "STD-A",
+                            "title_zh": "Standard A",
+                            "source_url": "https://example.com/a",
+                        }
+                    ]
+                }
+            ),
+            encoding="utf-8",
+        )
+
+    def run_fetcher(self, root: Path, result: fetcher.FetchResult) -> int:
+        arguments = [
+            "fetch_standard_references.py",
+            "--repo-root",
+            str(root),
+            "--standards",
+            "standards.json",
+            "--output-dir",
+            "references",
+            "--accessed-date",
+            "2026-08-12",
+        ]
+        with mock.patch.object(sys, "argv", arguments), mock.patch.object(
+            fetcher, "fetch_url", return_value=result
+        ):
+            return fetcher.main()
+
+    def test_transient_failure_preserves_last_successful_snapshot(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            self.write_standards(root)
+            reference = root / "references" / "std-a.md"
+            reference.parent.mkdir()
+            fetcher.write_reference_markdown(
+                reference,
+                {
+                    "standard_id": "STD-A",
+                    "title_zh": "Standard A",
+                    "source_url": "https://example.com/a",
+                    "source_status": "official",
+                },
+                fetcher.FetchResult(
+                    True,
+                    "fetched",
+                    final_url="https://example.com/final-a",
+                    raw_sha256="a" * 64,
+                    text="LAST KNOWN GOOD",
+                ),
+                "2026-08-01",
+            )
+            original = reference.read_bytes()
+
+            return_code = self.run_fetcher(
+                root, fetcher.FetchResult(False, "timeout", error="temporary timeout")
+            )
+
+            self.assertEqual(0, return_code)
+            self.assertEqual(original, reference.read_bytes())
+            record = json.loads((root / "references" / "index.json").read_text())[
+                "references"
+            ][0]
+            self.assertEqual("fetched", record["fetch_status"])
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2190 评审草稿

- 仓库：open-city-ai/haidian
- PR：#2190 `fix: preserve successful standard snapshots`
- 作者：jiangmuran
- checked exact head：`ceb6ca6e6c1b7c1d6ee3bf6798ccf8ef18a0d6be`
- 判级：**COMMENT**

---

## 拟发表正文（英文，随作者语言）

I am a fellow participant here, not a maintainer. Everything below is local
reproduction only, with no merge, scoring, or ranking opinion.

checked exact head: `ceb6ca6e6c1b7c1d6ee3bf6798ccf8ef18a0d6be`

This is a one-line allowlist change, so I focused on the two questions that a
one-line change actually raises: does the regression test really pin the bug, and
can preserving `fetched` ever freeze stale content?

### 1. Your regression test fails on main and passes on your head

This is the cleanest possible evidence, so I want to record it explicitly. I ran
your new module unchanged, once against your `fetch_standard_references.py` and
once against main's:

```
head ceb6ca6e6c1b7c1d6ee3bf6798ccf8ef18a0d6be : Ran 3 tests   OK
main 1a57349b96162dc0ae122336cdf4485b0376227a : Ran 3 tests   FAILED (failures=1)

FAIL: test_transient_failure_preserves_last_successful_snapshot
  self.assertEqual(0, return_code)
  AssertionError: 0 != 1
```

The other two tests pass on both, which is the right shape: they are guard tests
for behaviour you deliberately left alone.

### 2. The data loss is real, and it is worse than the return code suggests

I ran the failing scenario directly against both versions and inspected what
survives on disk. Same fixture, same simulated timeout:

| after a transient failure | main | this head |
|---|---|---|
| snapshot bytes preserved | **no** | yes |
| last-known-good text still on disk | **no** | yes |
| `index.json` `fetch_status` | `timeout` | `fetched` |
| `index.json` `accessed_date` | overwritten to today | `2026-08-01` (original) |
| `index.json` `final_url` | **`None`** | `https://example.com/final-a` |
| exit code | 1 | 0 |

So on main a single DNS blip or timeout does not merely mark the reference as
stale, it discards the offline copy and the provenance (`final_url`, original
`accessed_date`) that made the snapshot citable in the first place. That is the
part I would highlight if this needs justifying to a reviewer.

### 3. Preserving `fetched` cannot freeze stale content

The obvious objection to widening a preservation allowlist is that good refreshes
might stop landing. That cannot happen here, because the function short-circuits
before it ever looks at the existing file:

```python
def should_preserve_existing_reference(output_path: Path, result: FetchResult) -> bool:
    if result.ok:
        return False
    ...
```

A successful fetch always overwrites. Preservation is reachable only on failure.
I confirmed this behaviourally too — your
`test_successful_fetch_replaces_existing_snapshot` passes on both versions, and
`OLD SNAPSHOT` is gone from disk afterwards.

Read that way, the omission on main looks like a plain oversight: the allowlist
already contained three success-ish states
(`fetched_via_official_pdf_text`, `fetched_manual_official`,
`user_provided_summary`) and simply missed the ordinary one.

### 4. Context on blast radius, in case it helps triage

`scripts/fetch_standard_references.py` has no automated caller — no workflow and
no other script invokes it:

```bash
grep -rn "fetch_standard_references" scripts/ .github/    # no matches
```

So this is a maintainer-run tool and the change cannot affect PR CI. The stakes
are still real, though: the artefacts it maintains
(`brief/site-package/standards/standards.json` and the reference snapshots) are
read by `scripts/validate_submission.py` and `scripts/scaffold_ai_submission.py`,
so a corrupted refresh degrades the trusted brief data that every participant
package is checked against.

### 5. Your "Related work" notes check out

I verified both cross-references, since same-file overlap is the usual reason a
small PR stalls:

- **#1352** (yours) — also `scripts/fetch_standard_references.py`, plus
  `tests/test_fetch_standard_references.py`. Different function (slug collisions
  before fetching), different test module. Accurate as described.
- **#1944** (Sunrongguo2008, 64 files) — touches the same script for
  documentation/docstrings only.

Neither collides semantically with the allowlist at line ~213; whichever merges
second just needs a rebase.

### 6. On the CI signal for this PR

Worth stating plainly, because the green check is easy to over-read: the only
PR-triggered workflow is `submission-validation.yml`, which runs on
`pull_request_target` and checks out the trusted default branch, so it never
executes this PR's code. `is_non_submission_pr()` returns `True` for your file
list (I ran it against the live file listing), so
`scripts/github_pr_validation.py:611` returns 0 unconditionally with the
"participant package validation was not applicable" warning. And no
PR-triggered workflow runs `tests/` at all — the single `python3 -m unittest`
invocation in the repository lives in `gallery-snapshot-maintenance.yml`, fires
on push to `main` after merge, and runs three unrelated modules.

In other words, the green mark is not evidence about this change; the manual run
in section 1 is.

### Disclaimer

All of the above is local reproduction against `main` @
`1a57349b96162dc0ae122336cdf4485b0376227a`, Python 3.10.12. It is not an
official validation result and takes no position on merging. If you have already
pushed a new head, please disregard.

Thanks — the three-case test module (preserve on failure, replace on success,
placeholder on first failure) is exactly the shape that makes a one-line
allowlist change reviewable without re-deriving the control flow.

---

## 复核方法备注（内部，不发布）

1. 同一工作目录 `$HOME/y5_19662/repo`（四个 PR 共用），cone sparse + 独立 fetch 各 PR head。
2. 差分执行方式：用 `git show main:scripts/fetch_standard_references.py` 与
   `git show pr2190:...` 分别落到临时目录，importlib 载入（注意必须先
   `sys.modules[name]=m` 再 exec_module，否则 dataclass 在 3.10 下报
   `'NoneType' object has no attribute '__dict__'`）。
3. 数据损毁实证用 mock.patch fetch_url 返回 FetchResult(False,"timeout")，
   预置一份 fetch_status=fetched 的快照，比对前后字节与 index.json 记录。
4. 「无调用者」结论：workflow 与 scripts/ 全仓 grep 均为 0 命中。
5. 未发布任何 GitHub 内容；gh 仅用于 api GET。

PR #2194, opened 2026-08-12
fix: enforce proposal metadata length limits

jiangmuran:
## Summary
- enforce the published 2-120 character contract for proposal titles
- enforce the published 10-500 character contract for proposal summaries
- report the actual length for each violation

## Problem
`schema/proposal.schema.json` documents these bounds, but the trusted validator only checked that `title` and `summary` were non-empty. A proposal with a 121-character title or 501-character summary therefore passed intake and the gallery generator published the overlong value unchanged. One-character titles and nine-character summaries also passed despite violating the same schema.

## Tests
- `python3 -m unittest -v tests.test_proposal_metadata_lengths tests.test_front_matter tests.test_submission_workflow.ProposalSchemaTests`
- `python3 -m py_compile scripts/validate_submission.py tests/test_proposal_metadata_lengths.py`
- `git diff --check`

The focused matrix verifies title lengths 1/2/120/121 and summary lengths 9/10/500/501.

## Related work
#1471 also modifies `scripts/validate_submission.py`, but its patch is confined to persisted readiness claims, self-check validation, and validator call parameters. This change is localized to `validate_proposal_file()` metadata checks and uses a separate test module. #2113 changes section/schema relationships rather than these existing length bounds.

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -2481,6 +2481,17 @@ def validate_proposal_file(
     for key in required_metadata:
         if not metadata.get(key):
             report.add_error(f"{proposal_path}: missing front matter field `{key}`")
+    metadata_length_limits = {
+        "title": (2, 120),
+        "summary": (10, 500),
+    }
+    for key, (minimum, maximum) in metadata_length_limits.items():
+        value = metadata.get(key)
+        if value and not minimum <= len(value) <= maximum:
+            report.add_error(
+                f"{proposal_path}: front matter `{key}` must contain {minimum} to "
+                f"{maximum} characters; got {len(value)}"
+            )
 
     author = metadata.get("author_github", "")
     if author and author.lower() != pr_author.lower() and not report.maintainer_bypass:
--- a/tests/test_proposal_metadata_lengths.py
+++ b/tests/test_proposal_metadata_lengths.py
@@ -1,0 +1,77 @@
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+import sys
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+from validate_submission import (  # noqa: E402
+    REQUIRED_SECTIONS,
+    ValidationReport,
+    validate_proposal_file,
+)
+
+
+class ProposalMetadataLengthTests(unittest.TestCase):
+    def validate_metadata(self, title: str, summary: str) -> ValidationReport:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            relative = "submissions/alice/metadata-length/proposal.md"
+            proposal = root / relative
+            proposal.parent.mkdir(parents=True)
+            sections = "\n\n".join(
+                f"## {heading}\n{'完整设计证据与空间指标说明。' * 12}"
+                for heading in REQUIRED_SECTIONS
+            )
+            proposal.write_text(
+                "\n".join(
+                    [
+                        "---",
+                        f"title: {title}",
+                        'author_github: "alice"',
+                        'language: "zh"',
+                        'license: "CC-BY-4.0"',
+                        f"summary: {summary}",
+                        "---",
+                        "",
+                        "# Proposal",
+                        "",
+                        sections,
+                    ]
+                ),
+                encoding="utf-8",
+            )
+            report = ValidationReport()
+            validate_proposal_file(report, root, relative, "alice", "alice")
+            return report
+
+    def test_title_enforces_published_length_boundaries(self) -> None:
+        summary = "s" * 10
+        for length, valid in [(1, False), (2, True), (120, True), (121, False)]:
+            with self.subTest(length=length):
+                report = self.validate_metadata("t" * length, summary)
+                length_errors = [
+                    error for error in report.errors if "front matter `title`" in error
+                ]
+                self.assertEqual([] if valid else [length], [length] * len(length_errors))
+                if not valid:
+                    self.assertIn(f"got {length}", length_errors[0])
+
+    def test_summary_enforces_published_length_boundaries(self) -> None:
+        for length, valid in [(9, False), (10, True), (500, True), (501, False)]:
+            with self.subTest(length=length):
+                report = self.validate_metadata("Valid title", "s" * length)
+                length_errors = [
+                    error for error in report.errors if "front matter `summary`" in error
+                ]
+                self.assertEqual([] if valid else [length], [length] * len(length_errors))
+                if not valid:
+                    self.assertIn(f"got {length}", length_errors[0])
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2194 评审草稿

- 仓库：open-city-ai/haidian
- PR：#2194 `fix: enforce proposal metadata length limits`
- 作者：jiangmuran
- checked exact head：`4683eb9bafcadfd270f0a89e649a9ab6c6bff9b6`
- 判级：**COMMENT**

---

## 拟发表正文（英文，随作者语言）

I am a fellow participant here, not a maintainer. Everything below is local
reproduction only, and carries no merge, scoring, or ranking opinion.

checked exact head: `4683eb9bafcadfd270f0a89e649a9ab6c6bff9b6`

Of your current batch this is the one that adds a **new blocking error to the
trusted validator**, so I ran the corpus impact rather than just the tests. Short
version: your premise checks out, and the blast radius is **exactly 2 of 707
packages**.

### 1. The schema bounds you cite are real

I confirmed them rather than taking the description on trust, because the
properties are nested and easy to misread:

```python
json.load(open("schema/proposal.schema.json"))["properties"]["metadata"]["properties"]["title"]
# {'type': 'string', 'minLength': 2, 'maxLength': 120}
#                                 ["summary"]
# {'type': 'string', 'minLength': 10, 'maxLength': 500}
```

Worth adding one fact that strengthens the case: `schema/proposal.schema.json`
has **no enforcement caller anywhere** —

```bash
grep -rn "proposal.schema.json\|proposal_schema" scripts/ .github/   # no matches
```

so these bounds have been published-but-unenforced. This PR is the first thing
that actually applies them. That is a good argument for the change and also the
reason the corpus check below is worth having on record.

### 2. Corpus impact: exactly 2 of 707 packages

I materialised every `submissions/*/*/proposal.md` at `main`
`1a57349b96162dc0ae122336cdf4485b0376227a` and applied your rule using the
repository's own `parse_front_matter`:

```
proposal.md scanned : 707
front matter unparsed : 0
title violations      : 0
summary violations    : 2
```

```
summary len=704 (max 500)  submissions/wswch/jingzhang-ai-innovation-corridor/proposal.md
summary len=527 (max 500)  submissions/OceanHAN/ai-innovation-belt/proposal.md
```

Both are over the maximum; nothing in the corpus is under either minimum. I then
confirmed it through the real validator rather than a reimplementation of the
rule, calling `validate_proposal_file()` on those two packages with each version
of the script:

| | main | this head |
|---|---|---|
| `wswch/jingzhang-ai-innovation-corridor` | 5 errors, 0 length-rule | 6 errors, **1 length-rule** |
| `OceanHAN/ai-innovation-belt` | 5 errors, 0 length-rule | 6 errors, **1 length-rule** |

so the delta is precisely the intended `+1` and nothing else moves.

Since `validate_proposal_file()` only runs on proposal files in a PR's changed
set, neither package breaks while it sits untouched in `main` — but either would
start failing intake the next time it pushes. That seems worth a maintainer
knowing before this merges; it is a small, nameable set rather than an open
question.

### 3. Two scope gaps you may or may not want to close

Neither is a defect in what you wrote; both are about where the rule reaches.

**(a) Translations are never checked.** `proposal_paths_for()` in
`github_pr_validation.py:308` only ever emits `proposal.md`:

```python
proposals.add("/".join(parts[:3] + ["proposal.md"]))
```

so `proposal.en.md` / `proposal.zh.md` never reach `validate_proposal_file()` and
therefore never get the length rule, even though they carry the same front
matter. (There are 1360 `proposal*.md` files in the tree versus 707
`proposal.md`.)

**(b) `title_zh` / `summary_zh` carry identical bounds in the schema**
(2-120 and 10-500 respectively) but are not in your `metadata_length_limits`
dict. If the intent is "enforce the published contract", those two are part of
it.

### 4. One boundary case worth knowing about

`parse_front_matter` appends a trailing newline for a folded block scalar unless
the style strips it:

```python
parse_front_matter("---\ntitle: T\nsummary: >\n  " + "x"*500 + "\n---\nbody\n")["summary"]  # len 501
parse_front_matter("---\ntitle: T\nsummary: >-\n  " + "x"*500 + "\n---\nbody\n")["summary"]  # len 500
```

So a summary written as `summary: >` with exactly 500 visible characters is
counted as 501 and rejected, while `summary: >-` passes. Today this is latent —
**0 of 707** packages use a non-stripping block scalar for `summary`, and the
closest live values are 498 and 489 characters — so it is not a reason to change
anything now. I mention it only because the two nearest packages are within two
characters of the limit, and a diagnostic that says `got 501` for text that looks
like 500 characters would be genuinely confusing to a participant.

A no-code option would be to mention the `>-` form wherever the bound is
documented; whether that is worth doing is a maintainer call.

### 5. Cross-PR checks

- **#2113** (yours) modifies `schema/proposal.schema.json`, i.e. the very
  document this PR enforces. I read its diff: it only rewrites the `allOf`
  conditional blocks (bilingual contract wiring and the per-language section
  lists) and does **not** touch `title` or `summary` bounds. Your description of
  it is accurate — the numbers you hard-code here will not go stale from #2113.
- **#2183** (yours) rewrites `parse_front_matter` for CRLF-tolerant delimiters,
  which is upstream of this rule. I checked whether it would widen the blast
  radius by making currently-unparsed proposals parse: it does not. **0 of 707**
  proposals fail to parse under main's parser today, and re-running the corpus
  scan with #2183's parser yields the same 2 violations.
- **#1471** (147228, 2102 files) also modifies `scripts/validate_submission.py`.
  Your note that its patch is confined to readiness claims and validator call
  parameters matches what I see; the conflict risk is textual.

### 6. Test results

```
tests.test_proposal_metadata_lengths            Ran 2 tests    OK
tests.test_front_matter                         Ran 5 tests    OK
tests.test_submission_workflow                  Ran 135 tests  OK
tests.test_manifest_schema                      Ran 15 tests   OK
```

Also, since `metadata` comes from the repository's dependency-free
`parse_front_matter`, which returns `dict[str, str]`, `len(value)` is always
safe — there is no non-string path that could raise `TypeError` inside the
trusted validator. I checked that explicitly because a crash there would take
down the whole CI report rather than produce an error line.

### 7. On the CI signal

The green check here is not evidence about this change. The only PR-triggered
workflow, `submission-validation.yml`, runs on `pull_request_target` and checks
out the trusted default branch, so your modified validator is never the one that
executes. `is_non_submission_pr()` returns `True` for this file list (run against
the live listing), so `github_pr_validation.py:611` returns 0 unconditionally.
And no PR-triggered workflow runs `tests/` at all. Hence the manual runs above.

### Disclaimer

All local reproduction against `main` @
`1a57349b96162dc0ae122336cdf4485b0376227a`, Python 3.10.12,
`requirements-review.txt` plus `jsonschema==4.25.1`. Not an official validation
result, and no opinion on merging. If you have already pushed a new head, please
disregard.

Thanks for pinning the boundaries at 1/2/120/121 and 9/10/500/501 in the test —
that made the corpus check above a straight comparison rather than a
reinterpretation of intent.

---

## 复核方法备注（内部，不发布）

1. 全语料取数：`git ls-tree -r main submissions/` 抽出 707 条 proposal.md 路径，
   写成显式 sparse 模式表后 `git sparse-checkout set --no-cone --stdin`。
   注意 git 2.34.1 的 no-cone 模式**不支持 `**` 通配**（`submissions/**/proposal.md` 匹配 0 个），
   必须逐条列出精确路径；这样做还有一个好处：整批只触发一次 lazy blob fetch（1360 文件约 45s），
   而 `git cat-file --batch` 逐个懒取实测 20 个要 39s（不可用）。
2. 违规判定用仓库自带 scripts/front_matter.parse_front_matter，不自己实现 YAML 解析。
3. 差分实证调用真实 validate_proposal_file（非复刻规则），main/PR 各跑一次比对 error 条数。
4. 707 这个数字随 ref 漂移（pr2194 base 为 706、pr2190 base 为 704），
   正文统一以 **main 1a57349b96162dc0ae122336cdf4485b0376227a = 707** 为准并已写明 ref。
5. #2183 影响面对照：用 `git apply` 把 #2183 的 diff 打到 front_matter.py 副本后重跑同一扫描。
6. 未发布任何 GitHub 内容；gh 仅用于 api GET。

PR #2120, opened 2026-08-12
fix: keep malformed manifest audits reportable

jiangmuran:
## Summary
- skip legacy role advisories when a schema-invalid manifest has a non-array `files` value
- preserve the underlying schema errors and strict audit failure instead of raising `TypeError` during report assembly
- cover null, numeric, and object-shaped values through the JSON CLI contract

## Validation
- `python3 -m unittest tests.test_manifest_schema` (16 passed)
- `python3 -m py_compile scripts/manifest_schema.py scripts/validate_manifest_schema.py tests/test_manifest_schema.py`
- `git diff --check`

PR #2196, opened 2026-08-12
fix: fail empty manifest audits

jiangmuran:
## Summary
- fail manifest corpus audits when no manifests were selected
- keep JSON mode machine-readable with the existing counts plus a top-level error
- distinguish selection/configuration failure (exit 2) from strict schema failure (exit 1)

## Problem
`validate_manifest_schema.py --all --strict` returned success for an empty directory or a nonexistent `--repo-root`, reporting `0/0 valid`. The repository currently has hundreds of manifests, so an incorrect path or incomplete sparse checkout could silently turn a strict gate into a no-op.

## Tests
- `python3 -m unittest -v tests.test_validate_manifest_schema_cli tests.test_manifest_schema`
- `python3 -m py_compile scripts/validate_manifest_schema.py tests/test_validate_manifest_schema_cli.py`
- `git diff --check`

The new CLI tests cover strict JSON output on an empty root, advisory text output on a missing root, and the unchanged normal result path when one invalid manifest is actually present.

### DIFF
--- a/scripts/validate_manifest_schema.py
+++ b/scripts/validate_manifest_schema.py
@@ -29,6 +29,12 @@ def main() -> int:
         paths = manifest_paths(repo_root)
     else:
         paths = [(repo_root / item).resolve() for item in args.manifests]
+    selection_error = None
+    if not paths:
+        selection_error = (
+            f"no submission manifests found under {repo_root}; "
+            "check --repo-root and sparse checkout configuration"
+        )
 
     results = []
     for path in paths:
@@ -63,6 +69,8 @@ def main() -> int:
         "strict": args.strict,
         "results": results,
     }
+    if selection_error:
+        payload["error"] = selection_error
     legacy_findings = [
         finding
         for item in results
@@ -84,6 +92,8 @@ def main() -> int:
     else:
         mode = "strict" if args.strict else "advisory"
         print(f"Manifest schema audit ({mode}): {payload['valid']}/{payload['total']} valid")
+        if selection_error:
+            print(f"Audit error: {selection_error}")
         for item in invalid:
             print(f"- {item['path']}: {len(item['errors'])} schema error(s)")
             for error in item["errors"][:5]:
@@ -99,6 +109,8 @@ def main() -> int:
                     f"- {item['path']}: legacy role advisory "
                     f"{finding['path']} ({finding['role']}): {detail}"
                 )
+    if selection_error:
+        return 2
     return 1 if args.strict and invalid else 0
 
 
--- a/tests/test_validate_manifest_schema_cli.py
+++ b/tests/test_validate_manifest_schema_cli.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SCRIPT = REPO_ROOT / "scripts" / "validate_manifest_schema.py"
+
+
+class ValidateManifestSchemaCliTests(unittest.TestCase):
+    def run_audit(self, root: Path, *arguments: str) -> subprocess.CompletedProcess[str]:
+        return subprocess.run(
+            [sys.executable, str(SCRIPT), "--repo-root", str(root), *arguments],
+            cwd=REPO_ROOT,
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+
+    def test_strict_json_audit_rejects_empty_manifest_set(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            completed = self.run_audit(Path(tmp), "--all", "--strict", "--json")
+
+        self.assertEqual(2, completed.returncode)
+        payload = json.loads(completed.stdout)
+        self.assertEqual(0, payload["total"])
+        self.assertEqual([], payload["results"])
+        self.assertIn("no submission manifests found", payload["error"])
+
+    def test_advisory_audit_rejects_missing_repository_root(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            missing = Path(tmp) / "missing-root"
+            completed = self.run_audit(missing, "--all")
+
+        self.assertEqual(2, completed.returncode)
+        self.assertIn("Manifest schema audit (advisory): 0/0 valid", completed.stdout)
+        self.assertIn("Audit error: no submission manifests found", completed.stdout)
+
+    def test_existing_manifest_uses_normal_schema_result_path(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2196 评审草稿

- 仓库：open-city-ai/haidian
- PR：#2196 `fix: fail empty manifest audits`
- 作者：jiangmuran
- checked exact head：`8a00b81c4ec1b971da144d585612ae052726a062`
- 判级：**COMMENT**

---

## 拟发表正文（英文，随作者语言）

I am a fellow participant here, not a maintainer. Local reproduction only, with
no merge, scoring, or ranking opinion.

checked exact head: `8a00b81c4ec1b971da144d585612ae052726a062`

The behaviour change is confirmed and I could not find a regression. One piece of
context that is missing from the description seems worth adding, since it changes
how a maintainer would weigh the change.

### 1. The bug reproduces on main, in both modes

Your description mentions `--all --strict`; the silent success is actually
broader than that. On main, both an empty directory and a nonexistent
`--repo-root` return 0 in **strict and advisory alike**:

```
main   --repo-root <empty-dir>       --all --strict  ->  "0/0 valid"   exit 0
main   --repo-root <missing-dir>     --all           ->  "0/0 valid"   exit 0

head   --repo-root <empty-dir>       --all --strict  ->  "0/0 valid" + Audit error  exit 2
head   --repo-root <missing-dir>     --all           ->  "0/0 valid" + Audit error  exit 2
```

The advisory case is arguably the more dangerous one, because a caller reading
the human-readable line sees a well-formed `0/0 valid` with a success code and
nothing that looks wrong.

### 2. No regression on the real corpus

Both versions behave identically when manifests are actually present:

```
main  --repo-root .           ->  advisory  exit 0
main  --repo-root . --all --strict  ->  strict  exit 1
head  --repo-root .           ->  advisory  exit 0
head  --repo-root . --all --strict  ->  strict  exit 1
                                 (627/706 valid at that ref)
```

The JSON contract also stays put — key set unchanged, and the new `error` key is
absent whenever a selection exists:

```python
sorted(payload) == ['invalid', 'legacy_role_summary', 'results', 'strict', 'total', 'valid']
'error' in payload  # False
```

Tests: your new `tests.test_validate_manifest_schema_cli` runs 3 tests OK, and
`tests.test_manifest_schema` stays at 15 OK.

I also checked the one input shape that could have made this over-trigger: the
flag is `--manifest` (singular, `action="append"`), and the selection line is

```python
if args.all or not args.manifests:
    paths = manifest_paths(repo_root)
```

so a bare invocation without `--all` already scans the whole corpus. `exit 2`
therefore fires only when the selection is genuinely empty, not when someone
omits `--all`. That is the right boundary.

### 3. Context worth adding: there is no gate calling this script today

Your Problem statement says an incorrect path or incomplete sparse checkout
"could silently turn a strict gate into a no-op". I went looking for that gate
and could not find one:

```bash
grep -rn "validate_manifest_schema" scripts/ .github/     # no matches
```

Repository-wide code search returns two hits, and both are prose rather than
invocations — `docs/manifest-schema-migration.md` and one package changelog. The
documented usage in that doc is:

```bash
python3 scripts/validate_manifest_schema.py --all --json
```

which is unaffected by this change in a correct checkout.

I am not raising this as an objection — hardening a tool before it is wired into
a gate is a reasonable order of operations, and the failure mode you describe is
exactly what would bite later. It just seems better for the description to say
"this tool is manually invoked today" than to imply a live gate is currently
being defeated, because a maintainer triaging 36 open PRs will weigh those two
framings very differently.

### 4. One small note on the advisory exit code

After this change, advisory mode returns `2` for an empty selection. That is
deliberate and documented in your summary (`distinguish selection/configuration
failure (exit 2) from strict schema failure (exit 1)`), and I think it is the
right call. The only caveat is for a future caller written as
`if returncode != 0: "manifests are invalid"` — it would now misattribute a
configuration problem as a schema problem. Since there are no callers yet, this
is a matter of documenting the three-value contract (0 / 1 / 2) alongside the
tool rather than changing anything here.

### 5. Adjacent PR you did not cite

**#2120** (yours, `fix: keep malformed manifest audits reportable`) modifies
`scripts/manifest_schema.py` — the module this CLI imports `manifest_paths`,
`schema_errors`, and `legacy_role_findings` from. The two changes are
complementary and touch different files, so there is no conflict, but they are
the same audit path and a reviewer looking at one will probably want to see the
other. Might be worth a line in the description, as you did for #2190.

### 6. On the CI signal

As with the rest of this batch, the green check is not evidence about the change.
`submission-validation.yml` is the only PR-triggered workflow; it runs on
`pull_request_target` and checks out the trusted default branch, so this PR's
script never executes. `is_non_submission_pr()` returns `True` for this file list
(run against the live listing), so `github_pr_validation.py:611` returns 0
unconditionally with the "participant package validation was not applicable"
warning. And no PR-triggered workflow runs `tests/` at all — the repository's
only `python3 -m unittest` invocation is in `gallery-snapshot-maintenance.yml`,
which fires on push to `main` after merge and runs three unrelated modules. So
`tests/test_validate_manifest_schema_cli.py` will never be executed by CI, before
or after merge.

### Disclaimer

All local reproduction against `main` @
`1a57349b96162dc0ae122336cdf4485b0376227a`, Python 3.10.12,
`requirements-review.txt` plus `jsonschema==4.25.1`. Not an official validation
result and no position on merging. If you have already pushed a new head, please
disregard.

Thanks — the three CLI tests (strict JSON on empty root, advisory text on missing
root, unchanged normal path) cover exactly the axes that make an exit-code change
safe to reason about.

---

## 复核方法备注（内部，不发布）

1. 同一工作目录 `$HOME/y5_19662/repo`；用 `git checkout main|pr2196 -- scripts/validate_manifest_schema.py`
   做单文件对拷差分，避免整树切换触发大批 lazy fetch。
2. 「零调用者」双重取证：本地 grep（scripts/ + .github/ = 0）＋ GitHub code search API
   `q='validate_manifest_schema repo:open-city-ai/haidian'` → total_count=2，
   两条均为 docs/manifest-schema-migration.md 与 submissions/ziyanTOP/centennial-corridor/changelog.md。
   本地 grep 不足以覆盖未检出的 docs/、submissions/，故必须用 code search 兜底。
3. 真实语料跑分 627/706 取自当时 checkout 的 ref；正文只用它做「两版一致」的对照，不作为独立结论。
4. 未发布任何 GitHub 内容；gh 仅用于 api GET / search。

CocoSgt (approved):
Approved on exact head 8a00b81c4ec1b971da144d585612ae052726a062. Independent current-main integration is clean. The new CLI suite plus existing manifest schema tests pass 18/18; py_compile and git diff --check pass. Empty or nonexistent roots now produce a machine-readable selection error and exit 2 in both advisory and strict modes, while a real selected manifest keeps the existing result schema and exit behavior. The tool is manually invoked today rather than a live required gate, but hardening it before workflow reuse prevents an incorrect sparse/root configuration from being reported as 0/0 success.

PR #1386, opened 2026-08-10
fix(review): split content and professional scoring readiness

147228:
## Summary

Fixes #1368 by separating package readiness for content review from eligibility for official-geometry-dependent professional scoring.

## What changed

- Add machine-readable `content_review_eligible`, `professional_scoring_eligible`, and `professional_scoring_blocked_by` fields to self-check and maintainer summaries.
- Keep `can_enter_formal_review` and `formal-review-ready` as documented compatibility aliases for content-review readiness.
- Make `generate_formal_scorecard.py` return `blocked` when official site or key-area geometry is still missing.
- Align the maintainer workflow, review prompt, README, PR template, rubric, simulations, and urban-design skill with the three-axis contract.
- Add provisional and official fixture regressions.

For the current site package, a valid provisional submission remains content-review eligible, while professional scoring is blocked by `official_site_boundary` and `official_key_areas` until trusted geometry is available.

## Validation

- `python3 -m unittest tests.test_formal_scorecard tests.test_maintainer_review tests.test_agent_scaffold_and_self_check tests.test_submission_workflow tests.test_ai_review_submission tests.test_site_package_contract` — 161 passed
- `python3 -m py_compile scripts/self_check_submission.py scripts/maintainer_review.py scripts/generate_formal_scorecard.py scripts/review_submission.py`
- JSON schema parsing and `git diff --check` passed

This PR changes no submission package, `submissions-data.js`, or `gallery-publication.json`.

PR #2198, opened 2026-08-12
fix: keep maintainer review on trusted scripts

jiangmuran:
## Summary
- always run the maintainer worker's trusted `self_check_submission.py`
- ignore same-named scripts in a participant-controlled review checkout
- add dependency-free path and command-level regressions

## Problem
`review_submission.py` and `self_check_submission.py` intentionally resolve executables from their own trusted script directory. `maintainer_review.py` was the exception: its `script_path()` preferred `<repo_root>/scripts/self_check_submission.py` whenever present. Maintainer review commonly receives a PR checkout as `repo_root`, so a participant could modify that Python file and have the trusted review host execute it before producing a recommendation.

## Tests
- `python3 -m unittest -v tests.test_maintainer_review_trusted_script tests.test_trusted_review_scripts`
- `python3 -m py_compile scripts/maintainer_review.py tests/test_maintainer_review_trusted_script.py`
- `git diff --check`

## Related work
#1944 changes documentation/help text in this script. #1515 changes subprocess encoding. #1386 changes review-readiness semantics. None changes executable resolution; this patch is limited to `script_path()` and a separate test module.

### DIFF
--- a/scripts/maintainer_review.py
+++ b/scripts/maintainer_review.py
@@ -18,9 +18,8 @@ DEFAULT_OUTPUT_ROOT = ".maintainer-review"
 
 
 def script_path(repo_root: Path, name: str) -> Path:
-    candidate = repo_root / "scripts" / name
-    if candidate.exists():
-        return candidate
+    # repo_root can be a participant-controlled PR checkout. Review commands
+    # must always execute the maintainer worker's trusted script version.
     return SCRIPT_DIR / name
 
 
--- a/tests/test_maintainer_review_trusted_script.py
+++ b/tests/test_maintainer_review_trusted_script.py
@@ -1,0 +1,65 @@
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+from unittest import mock
+import sys
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+import maintainer_review  # noqa: E402
+
+
+class MaintainerReviewTrustedScriptTests(unittest.TestCase):
+    def test_script_path_ignores_participant_checkout_script(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            checkout = Path(tmp)
+            untrusted = checkout / "scripts" / "self_check_submission.py"
+            untrusted.parent.mkdir()
+            untrusted.write_text("raise RuntimeError('participant code ran')\n", encoding="utf-8")
+
+            resolved = maintainer_review.script_path(checkout, untrusted.name)
+
+            self.assertEqual(REPO_ROOT / "scripts" / untrusted.name, resolved)
+
+    def test_review_command_uses_trusted_self_check_script(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            checkout = Path(tmp)
+            submission = checkout / "submissions" / "alice" / "proposal"
+            submission.mkdir(parents=True)
+            untrusted = checkout / "scripts" / "self_check_submission.py"
+            untrusted.parent.mkdir()
+            untrusted.write_text("raise RuntimeError('participant code ran')\n", encoding="utf-8")
+            output = checkout / "review-output"
+            self_check = {
+                "returncode": 1,
+                "ok": False,
+                "stdout": {},
+                "stderr": "fixture failure",
+            }
+            with mock.patch.object(
+                maintainer_review, "run_json_command", return_value=self_check
+            ) as run_command, mock.patch.object(
+                maintainer_review,
+                "build_review_input",
+                return_value={
+                    "submission_dir": "submissions/alice/proposal",
+                    "rubric_dimensions": [],
+                },
+            ):
+                maintainer_review.run_maintainer_review(
+                    checkout, submission, "alice", output
+                )
+
+            command = run_command.call_args.args[0]
+            self.assertEqual(
+                REPO_ROOT / "scripts" / "self_check_submission.py", Path(command[1])
+            )
+            self.assertNotEqual(untrusted, Path(command[1]))
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2198 review draft (English; COMMENT)

## Draft body to publish

I am a participant in this open call (not a maintainer). Everything below is a reproducible script-level check; it is not a review decision, a score, or a merge recommendation.

- checked exact head: `5371226db26162b294d70a24b36d681f18f6b040`
- trusted baseline: main `1a57349b96162dc0ae122336cdf4485b0376227a`. While I was testing, main advanced to `e2d2a9bac5d053ee760d6b8fa811049321d9bb63`; the compare between the two is 24 commits and every changed file is under `submissions/`, so `scripts/`, `tests/`, `.github/workflows/` and `schema/` are identical to what I tested against.

### 1. What the CI check on this PR will and will not tell you

`submission-validation` is still `pending` here (run 31615213346, queued behind the repo-wide `submission-validation-api` lock). When it lands it will be green, and the green means less than it looks like. I ran the trusted entry point's own predicate on this PR's real file list:

```python
from github_pr_validation import is_non_submission_pr, is_review_queue_candidate
files = ["scripts/maintainer_review.py", "tests/test_maintainer_review_trusted_script.py"]
is_non_submission_pr(files)                      # True
is_review_queue_candidate(files, "jiangmuran")   # False
```

`True` hits the short circuit at `scripts/github_pr_validation.py:611`, which posts "non-submission code/docs/test PR; participant package validation was not applicable" and returns 0 before any package validation runs. And no workflow in `.github/workflows/` invokes pytest at all (I grepped all four), so **the two tests you added are not executed by any CI job**. The green tick on this PR is a statement about routing, not about your patch. That is the reason I ran everything below locally.

### 2. The gap is real, and `maintainer_review.py` was the only file still exposed

Your PR description says the two sibling entry points already resolve executables from their own trusted directory. I confirmed that on main, and it is stronger than "two other files do it" — the invariant is already written down in code *and* in a test module:

- `scripts/review_submission.py:94-99` — returns `SCRIPT_DIR / name`, with the comment "the checkout ... can contain participant-controlled scripts, so it must never supply executables to the review process".
- `scripts/self_check_submission.py:24-27` — returns `SCRIPT_DIR / name`, "repo_root may be an untrusted or older PR checkout".
- `tests/test_trusted_review_scripts.py` — already asserts exactly this for `review_submission` and `self_check_submission`, and only for those two.
- `scripts/maintainer_review.py:20-24` — the one remaining `if candidate.exists(): return candidate`.

So this patch closes the last hole in an invariant the repository already committed to, and your new test module is the third case of a test that already exists for the other two.

I reproduced the consequence end to end rather than reasoning about it. Fake participant checkout, containing nothing but a `scripts/self_check_submission.py` that writes a marker file and prints plausible JSON:

```bash
T=$(mktemp -d); mkdir -p $T/scripts $T/submissions/alice/proposal
cat > $T/scripts/self_check_submission.py <<'EOF'
import json, pathlib
pathlib.Path(__file__).resolve().parents[1].joinpath("PARTICIPANT_CODE_RAN.txt").write_text("x")
print(json.dumps({"ok": True, "checks": []}))
EOF
printf -- "---\ntitle: x\n---\n# x\n" > $T/submissions/alice/proposal/proposal.md

# trusted checkout at main, untrusted worktree passed as --repo-root
python3 scripts/maintainer_review.py submissions/alice/proposal --repo-root $T --pr-author alice --out $T/out
```

- on main `1a57349b9616`: exit `0`, and `$T/PARTICIPANT_CODE_RAN.txt` exists — the participant file ran under the reviewer's account, before any recommendation was produced.
- with your head file in place: no marker file, the trusted `self_check_submission.py` runs instead, and the command exits `1` with `recommendation: request-changes` (correct for that deliberately empty package).

Two points I checked so the scope is stated accurately, in both directions:

- The documented invocations (`README.md:362`, `docs/maintainer-workflow.md:26` and `:102`, `docs/simulations/fact-pack-provisional-intake-rehearsal.md:63`) all use the default `--repo-root .`, i.e. one checkout, where `SCRIPT_DIR` and `repo_root/scripts` are the same directory and the patch changes nothing. The gap only opens in the "trusted script + separate reviewed worktree" configuration — which is exactly the configuration `review_submission.py`'s comment describes, the one `github_pr_validation.py:696` uses (`trusted_repo_root = Path(__file__).resolve().parent.parent`), and the one `docs/review-rubric.md:70` anticipates when it says a trusted external worker may auto-review and auto-merge after the local gates pass.
- `scripts/generate_formal_scorecard.py:130` calls `run_maintainer_review(repo_root, ...)` with its own `--repo-root` argument, so the formal-scorecard path inherits the same resolution. Worth knowing that the fix covers that caller too.

### 3. Patch verification

| check | result |
| --- | --- |
| `tests/test_maintainer_review_trusted_script.py` on main (no patch) | 2 failed — resolved path was the temp checkout's `scripts/self_check_submission.py` |
| same tests with the head file | 2 passed |
| `tests/test_maintainer_review.py` + `tests/test_formal_scorecard.py` with the patch | 8 passed |
| full `python3 -m pytest tests -q` | baseline (no patch) 9 failed / 395 passed; with the patch 9 failed / 397 passed — identical failure set |

Those 9 pre-existing failures are all in my environment or already red on main before I touched anything (sparse checkout without `submissions/` for `test_prelaunch_check` and `test_submissions_gallery`, missing Arial Unicode font for `test_bilingual_backfill` — the same font gap you noted in #2202 — and one `test_site_package_contract::test_skill_exists` string assertion). None of them touch the modules in this PR, and the set does not change with your patch applied.

One neutral observation, not a request: after the change `repo_root` is unused in `script_path()`. Both siblings keep the parameter in the same way, so the signature stays compatible with every caller; I mention it only so a reviewer does not read it as an oversight.

I only checked the executable-resolution behaviour and the test evidence here. I did not evaluate design, wording, or anything about your own submission package. Thanks for taking the third entry point — the two sibling files made the intended invariant easy to verify.

PR #2108, opened 2026-08-12
fix: align publication schema state rules

jiangmuran:
## Summary

- encode the gallery generator's publication-state invariants in the published JSON Schema
- require published entries to have publication approval and completed rights review
- require featured entries to be published and use the featured quality tier
- keep unrelated unknown fields rejected

## Reproduction

`schema/gallery-publication.schema.json` is the maintainer-facing contract for `gallery-publication.json`, but it currently accepts states that `load_publication_registry()` rejects before gallery generation:

```text
published=true, review_status=not_approved
published=true, rights_reviewed=false
featured=true, published=false
featured=true, quality_tier=qualified
quality_tier=featured, featured=false
```

All five combinations validate successfully against current `main`, then fail in `scripts/generate_submissions_data.py`. Editors and schema-aware tools therefore report invalid publication records as valid.

## Root cause

The schema validates each field independently but does not express the cross-field state transitions already enforced by the generator.

## Verification

```text
python3 -m unittest tests.test_gallery_publication_schema -v
Ran 3 tests ... OK

python3 -m unittest \
  tests.test_submissions_gallery.TestSubmissionsGallery.test_publication_registry_requires_human_and_rights_approval \
  tests.test_submissions_gallery.TestSubmissionsGallery.test_publication_registry_rejects_invalid_date_and_flag_types \
  tests.test_submissions_gallery.TestSubmissionsGallery.test_publication_registry_rejects_missing_selection_metadata -v
Ran 3 tests ... OK

python3 -m py_compile tests/test_gallery_publication_schema.py
git diff --check
```

The focused tests cover four consistent states, each cross-field rejection, valid date formats, and continued rejection of unknown fields.

## Scope

No open pull request currently modifies `schema/gallery-publication.schema.json`. The runtime generator already enforces these exact rules and is unchanged. This patch is based on `main` at `d9a63788`.

PR #306, opened 2026-08-08
Make trusted submission validation stateful and deterministic

147228:
## What changed

- Run the trusted `pull_request_target` workflow for `converted_to_draft` events.
- Synchronize `review/draft` with the live PR state before validation and remove stale draft labels when a PR becomes ready.
- Apply `review/ci-failed` whenever a failing PR touches `submissions/`, even if invalid ownership means it cannot enter the review queue.
- Reconcile objective draft and deterministic-CI labels across all open PRs before each trusted `auto_review_queue.py --apply` run, while preserving human review outcomes on ready revisions.
- Require every local image embedded by `proposal.md` to be declared in `manifest.json`; trusted follow-up validation can now report undeclared images deterministically instead of depending on whether an asset happened to be in the current diff.
- Treat a declared SHA for `manifest.json` itself as an actionable warning instead of comparing an impossible self-referential digest; all other declared files remain strictly hash-checked.
- Add `refresh_manifest_hashes.py`, a path-safe all-or-nothing command for refreshing every declared file hash after revising an already finalized package, with contributor documentation and real-package regression coverage.
- Upgrade the trusted checkout step from `actions/checkout@v4` to `@v6` and use `pull_request_target`'s current default-branch `github.sha` instead of a PR's potentially stale `base.sha`.
- Treat explicit non-use and prohibition statements as compliant context instead of warning that phrases such as “未使用内部数据” or “不包含以下内容” may themselves disclose sensitive material; direct assertions still warn.
- Isolate each Chromium HTML-preview process with its own temporary profile and disable `/dev/shm` reliance so concurrent professional reviews do not intermittently time out while rendering small, self-contained deliverables.
- Remove a redundant leading Markdown H1 when the report hero already renders the same title, preserving a valid single-document heading outline.
- Add `self_check_submission.py --record-pass`: only a successful deterministic, spatial, visual, and professional-evidence run may persist `manifest.validation_claim.self_checked=true`; failed checks leave the manifest unchanged.
- Verify every review-label write with a live readback and one retry, and reconcile merged PRs that remain incorrectly queued after a post-merge label race.
- Classify every completed non-success `submission-validation` conclusion as a terminal CI failure for the review queue, while allowing a successful rerun on the same head to take precedence.
- Reject corrupted required PNG deliverables during deterministic validation by checking the PNG signature, chunk boundaries and CRCs, terminal `IEND`, and the compressed `IDAT` stream before the package enters AI review.
[...]

PR #2199, opened 2026-08-12
fix: require canonical gallery registry paths

jiangmuran:
## Summary
- enforce the published gallery registry path pattern at runtime
- reject dot segments, duplicate/trailing separators, invalid owners, and invalid slugs
- keep publication/hold keys identical to discovered submission paths

## Problem
The schema requires `submissions/<github-login>/<slug>`, but `load_publication_registry()` only inspected `Path(rel).parts` after stripping the value. Inputs such as `submissions//alice/example` were accepted and stored under that noncanonical string. `build_data()` later looked up the canonical key `submissions/alice/example`, missed the `published:false` hold, and published the merged proposal by default.

## Tests
- `python3 -m unittest -v tests.test_gallery_publication_paths tests.test_submissions_gallery.TestSubmissionsGallery.test_registry_can_explicitly_hold_a_merged_submission tests.test_submissions_gallery.TestSubmissionsGallery.test_publication_registry_rejects_missing_selection_metadata tests.test_submissions_gallery.TestSubmissionsGallery.test_publication_registry_rejects_invalid_date_and_flag_types`
- `python3 -m py_compile scripts/generate_submissions_data.py tests/test_gallery_publication_paths.py`
- `git diff --check`

The new tests cover five noncanonical path forms, runtime/schema pattern parity, and a canonical hold excluding a discovered proposal.

## Related work
#2108 adds schema state relationships but does not change the existing path pattern or runtime path handling. #306 and #1471 modify gallery readiness classification only. The path validation change is isolated and uses a separate test module.

### DIFF
--- a/scripts/generate_submissions_data.py
+++ b/scripts/generate_submissions_data.py
@@ -24,6 +24,9 @@ from front_matter import parse_front_matter as parse_front_matter_document
 
 
 PUBLICATION_FILE = "gallery-publication.json"
+PUBLICATION_PATH_RE = re.compile(
+    r"^submissions/[A-Za-z0-9-]{1,39}/[a-z0-9][a-z0-9-]{2,63}$"
+)
 
 
 REQUIRED_PACKAGE_FILES = {
@@ -410,11 +413,11 @@ def load_publication_registry(repo_root: Path) -> dict[str, dict[str, Any]]:
             raise SystemExit(f"{PUBLICATION_FILE}: entries[{index}].reviewed_by must be a GitHub login")
         if not isinstance(entry["reviewed_package_sha256"], str) or not re.fullmatch(r"[0-9a-f]{64}", entry["reviewed_package_sha256"]):
             raise SystemExit(f"{PUBLICATION_FILE}: entries[{index}].reviewed_package_sha256 must be a lowercase SHA-256")
-        rel = entry["path"].strip().rstrip("/")
+        rel = entry["path"]
+        if not PUBLICATION_PATH_RE.fullmatch(rel):
+            raise SystemExit(f"{PUBLICATION_FILE}: invalid submission path {rel}")
         if rel in registry:
             raise SystemExit(f"{PUBLICATION_FILE}: duplicate entry for {rel}")
-        if not rel.startswith("submissions/") or len(Path(rel).parts) != 3:
-            raise SystemExit(f"{PUBLICATION_FILE}: invalid submission path {rel}")
         if entry.get("published") is True and not (repo_root / rel / "proposal.md").is_file():
             raise SystemExit(f"{PUBLICATION_FILE}: published submission does not exist: {rel}")
         if entry.get("featured") is True and entry.get("published") is not True:
--- a/tests/test_gallery_publication_paths.py
+++ b/tests/test_gallery_publication_paths.py
@@ -1,0 +1,84 @@
+from __future__ import annotations
+
+import json
+import tempfile
+import unittest
+from pathlib import Path
+import sys
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+from generate_submissions_data import (  # noqa: E402
+    PUBLICATION_PATH_RE,
+    build_data,
+    load_publication_registry,
+)
+
+
+def held_entry(path: str) -> dict[str, object]:
+    return {
+        "path": path,
+        "published": False,
+        "featured": False,
+        "review_status": "not_approved",
+        "quality_tier": "qualified",
+        "reviewed_by": "maintainer",
+        "reviewed_at": "2026-08-12",
+        "rights_reviewed": False,
+        "reviewed_package_sha256": "0" * 64,
+        "selection_reason_zh": "维护者暂停公开展示",
+        "selection_reason_en": "Held from public display",
+        "selected_at": "2026-08-12",
+    }
+
+
+class GalleryPublicationPathTests(unittest.TestCase):
+    def write_registry(self, root: Path, path: str) -> None:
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2199 review draft (English; COMMENT)

## Draft body to publish

I am a participant in this open call (not a maintainer). What follows is a reproducible check of the script behaviour only — not a review decision, score, or merge recommendation.

- checked exact head: `9e85917aea7fc2d47db24a5a41eead7c0650db6b`
- trusted baseline: main `1a57349b96162dc0ae122336cdf4485b0376227a`; main has since moved to `e2d2a9bac5d053ee760d6b8fa811049321d9bb63`, and all 24 commits in between touch only `submissions/`, so `scripts/`, `schema/` and `tests/` are unchanged relative to my run.

### 1. The CI tick on this PR does not cover this change

`submission-validation` is `pending` (run 31615595768). I ran the trusted predicate on this PR's real file list: `is_non_submission_pr(["scripts/generate_submissions_data.py", "tests/test_gallery_publication_paths.py"])` is `True`, which takes the short circuit at `scripts/github_pr_validation.py:611` — comment posted, `return 0`, package validation skipped. And no workflow runs pytest, so your three new tests will not be executed by CI. Everything below is therefore local.

### 2. The silent-hold failure reproduces exactly as your description says

I did not take the description on trust; I ran `load_publication_registry()` and `build_data()` from **main** against a fixture with one discovered submission at `submissions/alice/example` and one `published:false` registry entry, varying only the entry's `path`:

| `path` in the registry (main behaviour) | accepted? | registry key stored | hold applied? | gallery output |
| --- | --- | --- | --- | --- |
| `submissions/alice/example` | yes | `submissions/alice/example` | yes | `[]` |
| `submissions//alice/example` | **yes** | `submissions//alice/example` | **no** | `['alice/example']` |
| `submissions/./alice/example` | **yes** | `submissions/./alice/example` | **no** | `['alice/example']` |
| `submissions/alice/example/` | yes | `submissions/alice/example` | yes | `[]` |
| `submissions/../templates` | **yes** | `submissions/../templates` | n/a | `['alice/example']` |

So on main the two dot/duplicate-separator forms are accepted, stored verbatim, missed by `registry.get(rel)` in `build_data()` at line 446, and the proposal the maintainer intended to hold is published — with exit code 0 and no warning anywhere. That is the failure mode worth fixing: not a crash, a silent inversion of a maintainer decision.

I also pushed the traversal case one step further than your description does. With `published: true` and `path: submissions/../templates`, main passes the path gate, passes the `(repo_root / rel / "proposal.md").is_file()` existence check against a directory **outside** `submissions/`, and gets as far as `package_sha256(repo_root / "submissions/../templates")`:

```
SystemExit: <root>/submissions/../templates: manifest.json with a files array is required
```

It only stopped there because `templates/` has no manifest. `gallery-publication.json` is maintainer-controlled (`PARTICIPANT_PROTECTED_GLOBAL_FILES` in `validate_submission.py:184`, plus the test that participants cannot modify it), so this is a typo/generation guard rather than an external attack surface — but "a mistyped path reads and hashes a directory outside the submissions tree" is worth closing regardless.

### 3. Why the specific regex is the right one, and what it costs

Three independent checks, all of which I ran:

1. `schema/gallery-publication.schema.json` already declares `"pattern": "^submissions/[A-Za-z0-9-]{1,39}/[a-z0-9][a-z0-9-]{2,63}$"` for `entries[].path`. Your `PUBLICATION_PATH_RE` is byte-identical to it — the loader is being brought up to the contract the repo already publishes, not given a new stricter one. (Your test asserting runtime/schema parity is the right way to keep that true; it is the one test in the file that will catch future drift.)
2. The slug half is identical to `validate_submission.PROPOSAL_SLUG_RE` (`scripts/validate_submission.py:87`), which is what every submission must satisfy at intake (`validate_submission.py:2884`). So nothing that can legally enter `submissions/` can be rejected by the new registry pattern.
3. Empirically: all **507** merged submissions listed in `submissions-data.js` match the new pattern; 0 would be rejected. Author segment and slug segment both clean.

The one genuine behaviour change is the trailing slash: `submissions/alice/example/` used to be normalized by `.strip().rstrip("/")` and accepted, and is now a hard `SystemExit`. It is the only input class that worked correctly before and stops working now. I think that is defensible (the schema rejects it too, and the failure is immediate and legible rather than silent), but it is the line I would call out to a reviewer as a deliberate strictness increase rather than a pure bug fix.

### 4. Patch verification

- With the head file: `tests/test_gallery_publication_paths.py` → 3 passed, 5 subtests passed. Re-running my table above against the patched loader: the canonical path still applies the hold; `submissions//alice/example`, `submissions/./alice/example`, `submissions/alice/example/`, `submissions/../templates`, `submissions/alice/Example` and `submissions/alice/ex` all fail with `invalid submission path`.
- Full suite: baseline on this checkout (no patch) 9 failed / 395 passed; with the patch 9 failed / 398 passed — same failure set, none related to this module. Those 9 are environmental or already red on main (no `submissions/` in my sparse checkout, missing Arial Unicode font, one skill-text assertion).
- The new test file cannot run against main at all — it imports `PUBLICATION_PATH_RE`, so it is a collection error rather than an assertion failure there. That is normal for a test shipped with a new constant; the behavioural before/after evidence is the table in section 2, produced with main's own loader.

Small aside, take it or leave it: `scripts/bootstrap_participant_workspace.py:44` uses a third slug pattern (`^[a-z0-9][a-z0-9-]{1,62}[a-z0-9]$`), which differs from `PROPOSAL_SLUG_RE` on trailing-hyphen slugs. Not in scope here, and your choice to mirror the schema and the validator is the correct one of the three.

I checked only the path-handling behaviour and the tests. I made no assessment of design, presentation, or anything in your own submission package. Thanks for the parity test between runtime and schema — that is the part that keeps this from drifting again.

PR #2200, opened 2026-08-12
fix: bound standard reference fetches

jiangmuran:
## Summary
- restrict standard reference requests and every redirect target to HTTP(S)
- replace automatic redirects with a five-hop loop that never reads intermediate bodies
- cap final responses at 10 MiB using both declared and actual byte limits

## Problem
`fetch_url()` accepted `file:` and `data:` URLs as successful official snapshots and read final responses without a size limit. Adding a bounded final `read()` alone is insufficient: Python's default `HTTPRedirectHandler` calls an unbounded `fp.read()` on every 30x response before following it, so a large redirect body still consumed unbounded memory.

## Tests
- `python3 -m unittest -v tests.test_fetch_standard_reference_network_bounds`
- real local HTTP server: 302 with a 1 MiB body to a relative tiny target, confirming the intermediate body is closed and the final page is parsed
- `python3 -m py_compile scripts/fetch_standard_references.py tests/test_fetch_standard_reference_network_bounds.py`
- `git diff --check`

The focused tests cover `file:`/`data:` initial URLs without opening, non-HTTP redirect targets without a second request, redirect bodies whose `read()` raises if touched, redirect limits, exact/max+1 actual sizes, and oversized `Content-Length` before reading.

## Related work
#1352 rejects filename collisions; #1944 changes docs/help; #2190 preserves a prior successful snapshot when any later fetch fails. This PR owns only request scheme, redirect, and response-size boundaries, with a separate test module.

### DIFF
--- a/scripts/fetch_standard_references.py
+++ b/scripts/fetch_standard_references.py
@@ -15,6 +15,7 @@ import json
 import re
 import sys
 import urllib.error
+import urllib.parse
 import urllib.request
 from dataclasses import dataclass
 from datetime import date
@@ -24,6 +25,10 @@ from typing import Any
 
 
 DEFAULT_USER_AGENT = "Mozilla/5.0 (compatible; haidian-ai-standards-fetcher/0.1)"
+ALLOWED_URL_SCHEMES = {"http", "https"}
+REDIRECT_STATUS_CODES = {301, 302, 303, 307, 308}
+MAX_REDIRECTS = 5
+MAX_RESPONSE_BYTES = 10 * 1024 * 1024
 
 
 @dataclass
@@ -85,6 +90,11 @@ class VisibleTextParser(HTMLParser):
         return "\n\n".join(lines)
 
 
+class NoRedirectHandler(urllib.request.HTTPRedirectHandler):
+    def redirect_request(self, req, fp, code, msg, headers, newurl):  # type: ignore[no-untyped-def]
+        return None
+
+
 def slugify(value: str) -> str:
     slug = re.sub(r"[^a-z0-9-]+", "-", value.lower()).strip("-")
     slug = re.sub(r"-{2,}", "-", slug)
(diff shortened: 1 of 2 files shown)


anselasimov-web (commented):
# PR #2200 review draft (English; COMMENT)

## Draft body to publish

I am a participant in this open call (not a maintainer). Below is a reproducible check of the fetch behaviour only; it is not a review decision, a score, or a merge recommendation.

- checked exact head: `cc9cc2c6918205c6602f8cc085bdfbdee728e307`
- trusted baseline: main `1a57349b96162dc0ae122336cdf4485b0376227a`; main has since advanced to `e2d2a9bac5d053ee760d6b8fa811049321d9bb63` with all 24 intervening commits under `submissions/`, so the scripts I tested against are current.

### 1. CI will not exercise this

`submission-validation` is `pending` here (run 31615960947). `is_non_submission_pr(["scripts/fetch_standard_references.py", "tests/test_fetch_standard_reference_network_bounds.py"])` returns `True` on main's own validator, which short-circuits at `scripts/github_pr_validation.py:611` and returns 0 without package validation; no workflow runs pytest. Also worth stating plainly: before this PR, **no test file in the repository imported `fetch_standard_references`** — this module had zero test coverage. So the only evidence available is local, which is why I ran it.

### 2. Behaviour differences, measured against a real HTTP server

Your test module is mock-based, which is the right choice for CI determinism. I ran the complementary check: the same `fetch_url()` from main and from your head, against a real `http.server` on 127.0.0.1, plus a real `file://` URL.

| case | main `1a57349b9616` | head `cc9cc2c69182` |
| --- | --- | --- |
| 200 OK | `fetched` | `fetched` |
| 302 → 200 (relative Location) | `fetched` | `fetched` |
| redirect chain ×3 | `fetched` | `fetched` |
| redirect chain ×8 | `fetched` | `too_many_redirects` |
| redirect loop | `http_302` (urllib loop detection) | `too_many_redirects` |
| 12 MiB body, `Content-Length` present | `fetched`, 12,582,912 chars parsed | `response_too_large`, before reading |
| 12 MiB body, no `Content-Length` | `fetched`, 12,582,912 chars parsed | `response_too_large`, at the read cap |
| `file:///tmp/probe.html` | **`fetched`**, local disk content returned | `unsupported_url_scheme` |
| `ftp://127.0.0.1/x` | `url_error` — after actually attempting the connection | `unsupported_url_scheme`, no connection attempted |

Two things I want to record precisely, one in each direction.

**The `file://` result is the strongest justification, and it is worse than "accepted".** On main the result is `status="fetched"` — identical to a real network fetch. `--update-standards` then writes that content into `brief/site-package/standards/references/` and records `fetch_status: fetched` in `index.json`, so a local file would be indistinguishable from an official page in the snapshot metadata. The current corpus is clean (`brief/site-package/standards/references/index.json`: 9 entries — 6 `fetched`, 1 `fetched_via_official_pdf_text`, 1 `user_provided_summary`, 1 `missing_source_url`; every entry that carries a `source_url` uses `http(s)`), so this is a guard against a future bad `source_url`, not an existing defect.

**Redirects on main are not unbounded, and I would not want the record to say otherwise.** CPython's default handler already caps at `max_redirections = 10` / `max_repeats = 4` and detects loops. Your cap of 5 is therefore a real tightening: a chain of 6–10 hops that works today starts returning `too_many_redirects` (my ×8 case). Chinese government portals do sometimes chain http→https→www→path; the corpus already contains one redirecting entry (`mohurd.gov.cn`, `fetched_via_official_pdf_text`, `final_url` ≠ `source_url`), and I cannot measure its hop count offline. This is the single line in the patch I would double-check before merge — and it is a one-constant change (`MAX_REDIRECTS`) if a real reference ever needs more, provided the failure is legible, which it is.

I also verified the claim your description rests on, because it is the reason a simple bounded `read()` would not have been enough. In CPython 3.10 `urllib/request.py`, `HTTPRedirectHandler.http_error_302` calls `fp.read()` and then `fp.close()` on every 30x response before following it — unbounded, by design. So bypassing the default redirect handler really is required to bound memory, not merely a stylistic preference. (Same function, line 63: the default allowlist for redirect targets is `http`, `https`, `ftp`, `''` — so on main an `http → ftp` redirect would also be followed.)

### 3. Patch verification

- `tests/test_fetch_standard_reference_network_bounds.py` on main: 5 of 6 test methods FAILED; the 6th reports both of its subtests as SUBFAILED. With your head file: 6 passed, 4 subtests passed.
- Full suite: baseline on this checkout 9 failed / 395 passed; with the patch 9 failed / 401 passed — identical failure set, all pre-existing and unrelated (no `submissions/` in my sparse checkout, missing Arial Unicode font — the same one you hit in #2202 — and one skill-text assertion).
- Downstream status strings: the new values (`unsupported_url_scheme`, `too_many_redirects`, `response_too_large`, `redirect_missing_location`, `invalid_url`) are only consumed by membership tests (`fetch_standard_references.py:215` and `:310`), and no schema constrains `fetch_status` / `reference_fetch_status`, so they degrade correctly to "not fetched" rather than breaking a consumer. I checked that rather than assuming it.

Two coordination notes, no action implied:

- **#2190 touches the same file.** Its only hunk is at `@@ -213` (`should_preserve_existing_reference`), while this PR rewrites `fetch_url()` at ~109–180, so I do not expect a textual conflict — but whichever lands second is worth re-running both test modules against.
- The happy path is intact in my runs (plain 200, single redirect, 3-hop chain all still `fetched` with identical final URLs), which was my main regression worry with a hand-rolled redirect loop.

I checked only network-boundary behaviour and the test evidence; I formed no view on design, scope, or your own submission package. Thanks for the boundary cases at exactly `MAX_RESPONSE_BYTES` and `+1`, and for asserting that the redirect body is never read — those two are what make the fix checkable.

PR #2202, opened 2026-08-12
fix: fail empty bilingual audits

jiangmuran:
## Summary

Make the maintainer-wide bilingual backfill audit fail when it discovers no proposal corpus, instead of reporting `Audit passed for 0 bilingual submissions`.

## Reproduction

On current `main`, run the documented audit against an empty or incorrectly rooted checkout:

```bash
tmp=$(mktemp -d)
python3 scripts/audit_bilingual_backfill.py --repo-root "$tmp"
```

Before this change, the script had no configurable root and the same empty-corpus code path returned success. In a sparse checkout without `submissions/`, that makes the documented corpus audit look complete while examining nothing.

## Change

- Add a `--repo-root` option so the corpus boundary is explicit and testable.
- Return exit code 2 with a clear diagnostic when no `submissions/*/*/proposal.md` files are discovered.
- Keep the existing per-pair checks and exit code 1 behavior unchanged.
- Add CLI regressions for both an empty corpus and a valid bilingual pair.

## Verification

```bash
python3 -m unittest tests.test_bilingual_audit_cli -v
python3 -m py_compile scripts/audit_bilingual_backfill.py tests/test_bilingual_audit_cli.py
git diff --check
```

The two focused tests pass. The broader `tests.test_bilingual_backfill` module also passed 22 dependency-independent tests; its image fixture still requires the repository's configured Arial Unicode font, which is absent in this Linux checkout.

### DIFF
--- a/scripts/audit_bilingual_backfill.py
+++ b/scripts/audit_bilingual_backfill.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import argparse
 from collections import Counter
 import re
 from pathlib import Path
@@ -111,16 +112,24 @@ def audit_pair(directory: Path) -> list[str]:
     return failures
 
 
-def main() -> int:
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--repo-root", type=Path, default=ROOT)
+    args = parser.parse_args(argv)
+    repo_root = args.repo_root.resolve()
+
     failures: list[tuple[Path, list[str]]] = []
-    directories = proposal_dirs(ROOT, [])
+    directories = proposal_dirs(repo_root, [])
+    if not directories:
+        print("Audit failed: no bilingual submissions found under submissions/*/*/proposal.md")
+        return 2
     for directory in directories:
         problems = audit_pair(directory)
         if problems:
             failures.append((directory, problems))
     if failures:
         for directory, problems in failures:
-            print(directory.relative_to(ROOT))
+            print(directory.relative_to(repo_root))
             for problem in problems:
                 print(f"  - {problem}")
         print(f"Audit failed for {len(failures)} of {len(directories)} submissions")
--- a/tests/test_bilingual_audit_cli.py
+++ b/tests/test_bilingual_audit_cli.py
@@ -1,0 +1,68 @@
+from __future__ import annotations
+
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+SCRIPT = ROOT / "scripts" / "audit_bilingual_backfill.py"
+
+
+class BilingualAuditCliTests(unittest.TestCase):
+    def run_audit(self, repo_root: Path) -> subprocess.CompletedProcess[str]:
+        return subprocess.run(
+            [sys.executable, str(SCRIPT), "--repo-root", str(repo_root)],
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+
+    def test_empty_submission_corpus_fails(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            completed = self.run_audit(Path(tmp))
+
+        self.assertEqual(2, completed.returncode)
+        self.assertIn("no bilingual submissions found", completed.stdout)
+        self.assertNotIn("Audit passed for 0", completed.stdout)
+
+    def test_valid_bilingual_pair_is_audited(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            repo_root = Path(tmp)
+            submission = repo_root / "submissions" / "alice" / "example"
+            submission.mkdir(parents=True)
+            (submission / "proposal.md").write_text(
+                """---
+title: Example
+language: zh
+translation_file: proposal.en.md
+---
+# Example
+
+[data:metrics.json#site_area]
+""",
+                encoding="utf-8",
+            )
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2202 review draft (English; COMMENT)

## Draft body to publish

I am a participant in this open call (not a maintainer). Below is a reproducible check of the script behaviour only — not a review decision, a score, or a merge recommendation.

- checked exact head: `f9300450e10f2aba61f37a680ec044ca6e81d4fd`
- trusted baseline: main `1a57349b96162dc0ae122336cdf4485b0376227a`; main has since moved to `e2d2a9bac5d053ee760d6b8fa811049321d9bb63`, and all 24 commits between them are under `submissions/`, so `scripts/` and `tests/` match what I tested.

### 1. CI will not run this either

`submission-validation` is `pending` (run 31617067013). `is_non_submission_pr(["scripts/audit_bilingual_backfill.py", "tests/test_bilingual_audit_cli.py"])` is `True` on main's validator → short circuit at `scripts/github_pr_validation.py:611`, comment posted, `return 0`. No workflow invokes pytest, so the two CLI regressions you added are not executed by CI. Local runs below.

### 2. The vacuous pass reproduces, and there is a second half to it

Straight reproduction on main, in a checkout where `submissions/` is not materialized (a sparse participant workspace — the layout the repo's own Skill recommends):

```
$ python3 scripts/audit_bilingual_backfill.py
Audit passed for 0 bilingual submissions
exit=0
```

That is the report your description targets, and `docs/maintainer-workflow.md:89` places exactly this command as the **last step** of the batch backfill pipeline (`backfill_bilingual_submissions.py` → three `backfill_bilingual_artifacts.py` runs → this audit). A green last line there reads as "the batch is verified", which is the reading the same document warns against two paragraphs later ("不得把批处理成功等同于语义审核完成").

The part I would add to your description: on main the script has **no `argparse` at all**, so an unknown flag is not rejected — it is silently ignored. That means

```
$ python3 scripts/audit_bilingual_backfill.py --repo-root /tmp/whatever
Audit passed for 0 bilingual submissions
exit=0
```

does not audit `/tmp/whatever`; it audits the script's own `ROOT` and prints success. So a maintainer who believes they pointed the audit at a specific checkout gets a pass for a corpus they never looked at. That is also why *both* of your new tests fail on main rather than just the empty-corpus one — `test_valid_bilingual_pair_is_audited` fails with `'Audit passed for 1 bilingual submissions' not found in 'Audit passed for 0 bilingual submissions'`, which is a compact demonstration of the same defect.

### 3. Patch verification

| check | result |
| --- | --- |
| `tests/test_bilingual_audit_cli.py` on main | 2 failed (`assertEqual(2, returncode)`; `Audit passed for 1` not found) |
| same tests with the head file | 2 passed |
| head, no-arg run in a checkout without `submissions/` | `Audit failed: no bilingual submissions found under submissions/*/*/proposal.md`, exit `2` |
| head, `--repo-root` at a 2-package fixture (1 valid pair, 1 with a heading mismatch + untranslated Chinese) | exit `1`, both per-pair diagnostics printed, `Audit failed for 1 of 2 submissions` |
| full `python3 -m pytest tests -q` | baseline 9 failed / 395 passed → with patch 9 failed / 397 passed, identical failure set |

The 9 pre-existing failures are environmental or already red on main and untouched by this patch — including `test_bilingual_backfill::test_localized_figure_retains_source_and_adds_translation_panel`, which fails here with `RuntimeError: Arial Unicode font is required for bilingual figure panels`, i.e. the same font gap you reported in your Verification section. I can confirm it independently on a clean Linux checkout.

Three details I checked because they are the usual regression sources for this shape of change:

- **`--repo-root` matches its sibling.** `scripts/backfill_bilingual_submissions.py:1230` already declares `--repo-root` with `type=Path, default=ROOT`; the auditor was the one script in that pipeline without it. Same option name, same default semantics — nothing new for a maintainer to learn.
- **`.resolve()` is load-bearing and correct.** `proposal_dirs()` globs from the passed root, and the failure branch prints `directory.relative_to(repo_root)`; resolving the argument first keeps that `relative_to` from raising on a symlinked temp root, and preserves main's behaviour where `ROOT` was already resolved.
- **Exit code 2 fits the repo convention** (`github_pr_validation.py` also reserves 2 for "cannot run" as distinct from "ran and found problems"), so a pipeline that branches on 1 vs 2 keeps working, and the existing 0/1 paths are untouched.

One thing a maintainer may want to decide explicitly, stated neutrally: in a sparse workspace without `submissions/`, the documented no-argument command now exits `2` instead of `0`. I think that is the point of the change — "audited nothing" should not look like "audited everything" — but it does mean anyone running the audit from a blobless sparse checkout will see a new non-zero exit, and it is better for that to be a decision than a surprise.

Related, not a request: #2196 applies the same empty-corpus shape to `scripts/validate_manifest_schema.py`. If both land, the two scripts will be consistent on the point.

I checked only the CLI behaviour and the test evidence; I made no assessment of design, translation quality, or your own submission package. Thanks for adding the configurable root — it is what makes this script testable at all.

CocoSgt (approved):
Approved on exact head f9300450e10f2aba61f37a680ec044ca6e81d4fd. Independent integration is against current main c80dc3648e16c4666d7a58cc3c3011c6f5914609 and is clean. The new CLI regressions pass 2/2, the existing bilingual backfill suite passes 22/22, py_compile and git diff --check pass. A direct empty-root run now emits the explicit no-corpus diagnostic and exits 2, while the valid-pair regression remains exit 0. This is the intended fail-closed boundary: a sparse or incorrectly rooted maintainer audit must not report that zero submissions passed. Existing per-pair exit 1 behavior is unchanged.

PR #2208, opened 2026-08-12
fix: derive exhibit boundary readiness from geometry

jiangmuran:
## Summary

Derive exhibit boundary readiness from the package's actual site and key-area GeoJSON instead of legacy `self_check.json` check IDs.

## Reproduction

`self_check_submission.py --mark-self-checked` replaces the scaffold's `BOUNDARY_TRUST` and `KEY_AREAS_TRUST` entries with the four persisted gate results:

```text
DETERMINISTIC_VALIDATION
SPATIAL_REVIEW
VISUAL_PACKAGING
PROFESSIONAL_EVIDENCE
```

`generate_exhibit.py::boundary_readiness()` still searched for the two removed IDs. Consequently, a fully official package produced the provisional caption:

```text
临时边界，保留精度警示并待正式数据发布后复算
```

The old helper also returned true if only one of its two expected IDs was present with `severity=info`, so incomplete legacy evidence could fail open.

## Change

- Reuse `generate_submissions_data.has_official_geometry()`, the existing gallery classification rule.
- Require non-empty site geometry, at least three key-area features, and `official_boundary=true` plus `geometry_role=official_constraint` on every relevant feature.
- Remove the stale self-check inference.
- Add dependency-free `build_exhibit()` regressions for a fully official package with persisted gate-only self-check data and for a package with one provisional key area.

## Verification

```bash
python3 -m unittest tests.test_generate_exhibit_boundary -v
python3 -m py_compile scripts/generate_exhibit.py tests/test_generate_exhibit_boundary.py
git diff --check
```

Both focused regressions pass. The existing scaffold-to-exhibit integration test is dependency-gated and skipped in this checkout because the optional geospatial review dependencies are absent.

### DIFF
--- a/scripts/generate_exhibit.py
+++ b/scripts/generate_exhibit.py
@@ -21,7 +21,7 @@ from pathlib import Path
 from typing import Any
 
 from validate_submission import parse_front_matter, parse_track_metadata
-from generate_submissions_data import load_publication_registry
+from generate_submissions_data import has_official_geometry, load_publication_registry
 
 
 COVER_CANDIDATES = (
@@ -42,26 +42,6 @@ def pick_image(submission_dir: Path, candidates: tuple[str, ...]) -> str | None:
     for path in sorted((submission_dir / "assets" / "figures").glob("*.png")):
         return path.relative_to(submission_dir).as_posix()
     return None
-
-
-def boundary_readiness(submission_dir: Path) -> bool:
-    """Return True when the package is on official (not provisional) geometry."""
-    self_check = submission_dir / "self_check.json"
-    if not self_check.exists():
-        return False
-    try:
-        data = json.loads(self_check.read_text(encoding="utf-8"))
-    except (UnicodeDecodeError, json.JSONDecodeError):
-        return False
-    checks = data.get("checks") if isinstance(data, dict) else None
-    if not isinstance(checks, list):
-        return False
-    relevant = {"BOUNDARY_TRUST", "KEY_AREAS_TRUST"}
-    seen: dict[str, str] = {}
-    for item in checks:
-        if isinstance(item, dict) and item.get("check_id") in relevant:
-            seen[str(item["check_id"])] = str(item.get("severity", ""))
-    return bool(seen) and all(severity == "info" for severity in seen.values())
 
 
 def build_exhibit(repo_root: Path, submission_dir: Path) -> dict[str, Any]:
@@ -90,7 +70,7 @@ def build_exhibit(repo_root: Path, submission_dir: Path) -> dict[str, Any]:
     publication = load_publication_registry(repo_root).get(rel)
     if not publication or publication.get("published") is not True:
         raise SystemExit(f"{rel}: submission is not approved for public display in gallery-publication.json")
-    official = boundary_readiness(submission_dir)
+    official = has_official_geometry(submission_dir)
     readiness = "官方边界就绪" if official else "临时边界，保留精度警示并待正式数据发布后复算；不阻断内容评分"
 
     return {
--- a/tests/test_generate_exhibit_boundary.py
+++ b/tests/test_generate_exhibit_boundary.py
@@ -1,0 +1,107 @@
+from __future__ import annotations
+
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from unittest import mock
+
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT / "scripts"))
+
+import generate_exhibit  # noqa: E402
+
+
+def write_collection(path: Path, properties: list[dict[str, object]]) -> None:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(
+        json.dumps(
+            {
+                "type": "FeatureCollection",
+                "features": [
+                    {
+                        "type": "Feature",
+                        "properties": item,
+                        "geometry": {
+                            "type": "Polygon",
+                            "coordinates": [[[0, 0], [1, 0], [1, 1], [0, 0]]],
+                        },
+                    }
+                    for item in properties
+                ],
+            }
+        ),
+        encoding="utf-8",
+    )
+
+
+class GenerateExhibitBoundaryTests(unittest.TestCase):
+    def make_geometry(self, root: Path, *, provisional_key_area: bool = False) -> Path:
+        submission = root / "submissions" / "alice" / "example"
+        official = {
+            "official_boundary": True,
+            "geometry_role": "official_constraint",
+        }
+        write_collection(submission / "geometry" / "site_boundary.geojson", [official])
+        key_areas = [dict(official) for _ in range(3)]
+        if provisional_key_area:
+            key_areas[-1] = {
+                "official_boundary": False,
+                "geometry_role": "provisional_constraint",
+            }
+        write_collection(submission / "geometry" / "key_areas.geojson", key_areas)
+        (submission / "assets" / "figures").mkdir(parents=True)
+        (submission / "assets" / "figures" / "site-overview.png").write_bytes(b"png")
+        (submission / "proposal.md").write_text(
+            "---\ntitle: Example\nsummary: A complete exhibit summary.\n---\n# Example\n",
+            encoding="utf-8",
+        )
+        return submission
+
+    def build_exhibit(self, repo_root: Path, submission: Path) -> dict:
+        publication = {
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2208 review draft (English; COMMENT)

## Draft body to publish

- checked exact head: `2571e727b49d2c0ab0917357260b7e704598f9e9`
- trusted baseline: main `d05f30a4ca2d9b0bbc14d177a0b3d433882d4461`. Main has since moved to `18085cdfe38fe671e883a8e7fcf7cd8b4b2bbee1`, and all 41 commits between them touch only `submissions/`, so `scripts/`, `tests/` and `gallery-publication.json` are identical to what I tested.

Worth stating explicitly because it changes what a reviewer has to do by hand:

- `submission-validation.yml` is triggered by `pull_request_target` and has exactly one checkout step, `ref: ${{ github.event.repository.default_branch }}`. It runs main's scripts against the PR's *files*; it never checks out or executes your head.
- The only `python3 -m unittest` invocation in any workflow is `gallery-snapshot-maintenance.yml:68`, and it names three modules explicitly (`tests.test_submissions_gallery`, `tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`) on push-to-main.

So `tests/test_generate_exhibit_boundary.py` is never executed by any workflow, before or after merge. Everything below is a local run.

### 2. The stale-id defect reproduces — and the fail-open half of it is the one that is actually live

Both halves of your description hold. Running the old `boundary_readiness()` and the new `has_official_geometry()` side by side over **all 713 merged packages that carry `self_check.json`** (fetched at the pinned main sha, blob-verified against `git ls-tree`):

| | packages where the predicate is True |
|---|---|
| old `boundary_readiness()` | **18** |
| new `has_official_geometry()` | **0** |

18 flips, all in the same direction (True → False), zero flips the other way.

Your second claim is the one that shows up on real data. Of those 18, **11 carry both ids and 7 carry only `BOUNDARY_TRUST`** with no `KEY_AREAS_TRUST` at all — exactly the "returned true if only one of its two expected IDs was present with `severity=info`" case:

```
submissions/LauDongwei/jingzhang-open-switchboard   ['BOUNDARY_TRUST']
submissions/ZephyrL0/jingzhang-ai-weave             ['BOUNDARY_TRUST']
submissions/sukikeeling/ai-jingzhang-corridor       ['BOUNDARY_TRUST']
submissions/sukikeeling/ai-jingzhang-corridor-v5    ['BOUNDARY_TRUST']
submissions/sukikeeling/ai-jingzhang-corridor-v6    ['BOUNDARY_TRUST']
submissions/sukikeeling/ai-jingzhang-corridor-v61   ['BOUNDARY_TRUST']
  (+1 more)
```

Meanwhile every one of the 713 packages carries `official_boundary: false` and `geometry_role: "provisional_constraint"` on every site-boundary and key-area feature (and exactly 3 key areas). So **no merged package is on official geometry today**, and those 18 were being labelled `官方边界就绪` purely on the strength of scaffold-era self-check rows while their own geometry says provisional. That is a false readiness label on the public-facing caption, and your patch removes it.

### 3. One correction to the description, and one scoping note

**Correction.** The PR says you "reuse `generate_submissions_data.has_official_geometry()`, the existing gallery classification rule." `has_official_geometry` is not currently part of gallery classification — it is referenced nowhere in the repository except its own definition at `scripts/generate_submissions_data.py:190`. `classify_submission()` (same file, line 210) decides on `stored_formal_readiness()` / `has_blocking_self_check()` / `package_complete()` and never calls it. This PR makes `generate_exhibit.py` the function's **first and only** caller. That does not make the choice wrong — the predicate matches the scaffold's own official/provisional contract at `scaffold_ai_submission.py:114-125` and `spatial_review.py:307-312` — but reviewers should know they are activating previously unexercised code rather than reusing a rule already load-bearing elsewhere.

**Scoping note (why this is not urgent, and why it is still worth stating).** `gallery-publication.json` on live main is `{"version": 1, "entries": []}`, so `build_exhibit()` raises `SystemExit("... is not approved for public display ...")` for every package right now. The 18→0 flip is therefore latent: it has no effect until a maintainer publishes an entry. The moment one of those 18 is published, the caption they would have got before this PR is `官方边界就绪` and after it is `临时边界…`. Given that the geometry actually is provisional, the post-PR caption is the accurate one — but this is a real, one-directional narrowing of an existing public-facing label, and it is the opposite direction from the false-provisional symptom the description leads with. It seems worth one line in the PR body so a maintainer reading only the summary does not expect the change to make more packages read as ready.

### 4. Test quality: one of the two regressions is anchored, the other is not

```
main + new test file:  1 FAIL, 1 pass
head:                  2 pass
```

`test_complete_official_geometry_is_ready_without_legacy_self_check_signals` is the anchored one — on main it fails with the real symptom:

```
AssertionError: '官方边界就绪' not found in
'方案概览示意图（临时边界，保留精度警示并待正式数据发布后复算；不阻断内容评分）。'
```

`test_any_provisional_key_area_keeps_exhibit_provisional` **also passes on main**. On main it passes because the fixture has no `BOUNDARY_TRUST`/`KEY_AREAS_TRUST` rows at all, so the old helper returns `False`; on head it passes because one key area is provisional. Same verdict, different reason — the test cannot distinguish old from new behaviour and would not catch a regression of this change. If you want it to bite, giving that fixture a passing legacy `BOUNDARY_TRUST`/`KEY_AREAS_TRUST` pair (i.e. the exact state of the 7 fail-open packages above) would make it fail on main and pass on head.

Also worth adding, since it is the case that actually exists in the repo: a fixture with `official_boundary: true` on the site boundary but `false` on one key area, asserting the caption stays provisional. And the fail-open case directly — one `BOUNDARY_TRUST` row at `severity: info`, no `KEY_AREAS_TRUST`, provisional geometry — which is what 7 merged packages look like today.

### 5. Regression and hygiene

- Full suite, identical environment: main `9 failed, 395 passed` → head `9 failed, 397 passed`. The two failure sets diff empty, so the +2 are exactly your new tests and nothing else moved. (The 9 pre-existing failures are artifacts of my checkout not materialising `submissions/`, not repository breakage.)
- `import json` is still required in `generate_exhibit.py` after the deletion (used at lines 124 and 151), so no unused import is left behind.
- `boundary_readiness` had exactly one caller (`generate_exhibit.py:93`); nothing else in the repo references it, so the deletion is safe.
- `has_official_geometry` requires `len(key_features) >= 3`. All 713 packages have exactly 3, so that bound is not currently the binding constraint — but it is a stricter shape requirement than the old helper had, and a future package with 2 or 4 key areas would be treated as non-official regardless of its flags. Probably intended; flagging it because it is a second tightening bundled into the same predicate.

### 6. Interaction with your #2230

#2230 also edits `scripts/generate_exhibit.py` from the same pre-image blob. I merged them to check: `git merge-file` returns 0 with no conflict markers and the combined file compiles, so they are textually independent.

One semantic overlap is worth coordinating, though. #2230's `generation_inputs()` lists `("self-check", submission_dir / "self_check.json")` as a protected input — but after #2208, `generate_exhibit.py` no longer reads `self_check.json` at all. Conversely the files #2208 promotes to primary inputs (`geometry/site_boundary.geojson`, `geometry/key_areas.geojson`) are not in #2230's explicit list; they are covered only via the manifest-declared-artifacts branch, which is skipped when `manifest.json` is missing or unparseable. Whichever lands second, it would be worth updating that list so the protected set matches the inputs the generator actually reads.

### 7. Verification I ran

```bash
# blob identity
git ls-tree 2571e727b49d2c0ab0917357260b7e704598f9e9 -- scripts/generate_exhibit.py tests/test_generate_exhibit_boundary.py
#   575d9501f40069d66c67e803bfe0f679eec26399  scripts/generate_exhibit.py
#   a696a90036b732b0344f139ba324f3eeeff96ad1  tests/test_generate_exhibit_boundary.py

# bidirectional
python3 -m unittest tests.test_generate_exhibit_boundary -v   # on main: 1 FAIL / 1 pass
python3 -m unittest tests.test_generate_exhibit_boundary -v   # on head: 2 pass

# full suite, both trees
python3 -m pytest tests -q                                     # main 9F/395P -> head 9F/397P, same failure set

# 713-package differential (old helper vs has_official_geometry), corpus pinned at main d05f30a4
```

Thanks for splitting this out with a dependency-free regression — the fail-open branch in the old helper is the kind of thing that only shows up if someone enumerates the merged corpus, and it turns out to be affecting 7 packages. Happy to re-run any of the above against a different fixture set if useful.

PR #2209, opened 2026-08-12
fix: keep review packets out of submissions

jiangmuran:
## Summary

Reject review-packet output directories inside `submissions/` before any maintainer-only artifact is written.

## Reproduction

On current `main`, the exporter accepts a participant package as its output directory:

```bash
python3 scripts/export_review_packet.py \
  submissions/alice/plan \
  --repo-root . \
  --out submissions/alice/plan
```

It writes `review-packet.md`, `review-packet.html`, and `packet-manifest.json` into the submission. This conflicts with the documented review-artifact boundary in `docs/maintainer-workflow.md`, pollutes participant package scope, and can invalidate manifest hashes.

## Change

- Resolve the requested output path before processing submissions.
- Reject any resolved path under the repository's `submissions/` tree.
- Keep the default `.maintainer-review/` location and external offline directories valid.
- Apply the guard in `export_review_packet()` so direct Python callers receive the same protection as the CLI.

## Verification

```bash
python3 -m unittest tests.test_review_packet_output_boundary tests.test_export_review_packet -v
python3 -m py_compile scripts/export_review_packet.py tests/test_review_packet_output_boundary.py
git diff --check
```

All five focused and existing review-packet tests pass.

### DIFF
--- a/scripts/export_review_packet.py
+++ b/scripts/export_review_packet.py
@@ -135,6 +135,19 @@ def inside_repo(path: Path, repo_root: Path) -> bool:
         return False
 
 
+def validate_output_dir(repo_root: Path, out_dir: Path) -> None:
+    """Keep maintainer-only review artifacts out of participant packages."""
+    try:
+        relative = out_dir.resolve().relative_to(repo_root.resolve())
+    except ValueError:
+        return
+    if relative.parts and relative.parts[0] == "submissions":
+        raise ReviewPacketError(
+            "review packet output must not be inside submissions/; "
+            "use .maintainer-review/ or an external directory"
+        )
+
+
 def discover_submission_dirs(repo_root: Path) -> list[Path]:
     submissions_root = repo_root / "submissions"
     return sorted(path.parent for path in submissions_root.glob("*/*/proposal.md"))
@@ -891,6 +904,7 @@ def export_review_packet(
     include_pdf: bool = False,
     pdf_engine: str = "auto",
 ) -> dict[str, str]:
+    validate_output_dir(repo_root, out_dir)
     packets = [load_submission_packet(repo_root, path) for path in submission_dirs]
     out_dir.mkdir(parents=True, exist_ok=True)
     markdown_path = out_dir / "review-packet.md"
--- a/tests/test_review_packet_output_boundary.py
+++ b/tests/test_review_packet_output_boundary.py
@@ -1,0 +1,37 @@
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+
+
+import sys
+
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT / "scripts"))
+
+from export_review_packet import ReviewPacketError, validate_output_dir  # noqa: E402
+
+
+class ReviewPacketOutputBoundaryTests(unittest.TestCase):
+    def test_submission_output_is_rejected_before_writes(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            repo_root = Path(tmp)
+            out_dir = repo_root / "submissions" / "alice" / "plan" / "review"
+
+            with self.assertRaisesRegex(ReviewPacketError, "must not be inside submissions"):
+                validate_output_dir(repo_root, out_dir)
+
+        self.assertFalse(out_dir.exists())
+
+    def test_local_maintainer_and_external_outputs_are_allowed(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp, tempfile.TemporaryDirectory() as external:
+            repo_root = Path(tmp)
+
+            validate_output_dir(repo_root, repo_root / ".maintainer-review" / "packet")
+            validate_output_dir(repo_root, Path(external) / "packet")
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2209 review draft (English; COMMENT)

## Draft body to publish

- checked exact head: `4bb9b75e90cfdbcd85836882550dcbb1fa33fae9`

`submission-validation.yml` is `pull_request_target` with a single checkout step pinned to `ref: ${{ github.event.repository.default_branch }}` — it runs main's scripts against your changed files and never checks out your head. The only `python3 -m unittest` in any workflow is `gallery-snapshot-maintenance.yml:68`, which names three modules explicitly and runs on push-to-main.

So `tests/test_review_packet_output_boundary.py` is never executed by any workflow, before or after merge. Everything below is a local run.

### 2. The reproduction holds on a real merged package

Not a fixture — `submissions/AlanHN/jingzhang-commons-line` pulled at the pinned main sha and blob-verified:

```
# main d05f30a4
$ python3 scripts/export_review_packet.py submissions/AlanHN/jingzhang-commons-line \
    --repo-root . --out submissions/AlanHN/jingzhang-commons-line
exit=0
  review-packet.md        44628 B
  review-packet.html      58945 B
  packet-manifest.json      742 B
... all three written inside the participant package.

# head 4bb9b75e, identical command
error: review packet output must not be inside submissions/; use .maintainer-review/ or an external directory
exit=1   (nothing written)
```

### 3. The guard itself is solid, including two cases you did not claim

I tried to get around it. `Path.resolve()` doing both normalisation and symlink resolution means the obvious bypasses are already covered:

| `--out` | head |
|---|---|
| `submissions/AlanHN/jingzhang-commons-line` | exit 1, rejected |
| `submissions` (the root itself) | exit 1, rejected |
| `<pkg>/../../../submissions/AlanHN/x` (traversal) | exit 1, rejected |
| symlink outside the repo pointing into `submissions/` | exit 1, rejected |
| default `.maintainer-review/...` | exit 0, works |
| `/tmp/extpacket` (external) | exit 0, works |

The traversal and symlink cases are worth mentioning in the PR body — they are the two a reviewer would most want reassurance about, and they pass.

One gap, low severity: **`--out .` is still accepted** and writes `review-packet.md`, `review-packet.html` and `packet-manifest.json` into the repository root (`relative.parts` is empty for `.`, so the `parts[0] == "submissions"` test is never reached). That is not a participant-package violation, but it is still a maintainer-only artifact landing in a tracked directory where it can be committed by accident — which is the same failure mode the guard exists to prevent. See the next point for a one-line way to close it.

### 4. There is already a `validate_output_dir` in this repo, and it is stricter

`scripts/ai_review_submission.py:124-134`:

```python
def validate_output_dir(repo_root: Path, out_dir: Path) -> None:
    try:
        relative = out_dir.resolve().relative_to(repo_root.resolve())
    except ValueError:
        return
    if not relative.parts or relative.parts[0] != DEFAULT_OUTPUT_ROOT:
        raise ReviewError(
            "Review output inside the repository must stay under `.maintainer-review/` to prevent accidental commits"
        )
```

Same name, same signature, same `resolve()/relative_to` skeleton — but an **allowlist** ("must be under `.maintainer-review/`") rather than a denylist ("must not be under `submissions/`"). The allowlist version rejects `--out .` for free, and it also rejects `--out docs/`, `--out data/`, and every other tracked directory, without needing a new rule each time someone finds one.

Both modules already define `DEFAULT_OUTPUT_ROOT = ".maintainer-review"` (`export_review_packet.py:22`, `ai_review_submission.py:34`), so the policies are meant to match. I would suggest either adopting the allowlist form here, or — better — lifting the existing one into a shared helper, since the repo will otherwise have two functions with the same name and different semantics. If the denylist is deliberate (e.g. to keep `--out .` usable for a maintainer scratch run), a sentence in the PR body saying so would settle it.

### 5. Three sibling scripts have the identical defect and are not covered

This is the part I would most want on the record. I ran the same "point `--out` at a participant package" experiment against every maintainer script that takes `--out`, on **your head**, against the same real package:

| script | result on head |
|---|---|
| `scripts/export_review_packet.py` | blocked (this PR) |
| `scripts/ai_review_submission.py` | blocked (pre-existing allowlist guard) |
| `scripts/review_submission.py` | **3 files written into the package** (`advisory-review.md`, `review-input.json`, `review-prompt.md`) |
| `scripts/maintainer_review.py` | **5 files written into the package** |
| `scripts/generate_formal_scorecard.py` | **7 files written into the package** |

All three unguarded ones default to `.maintainer-review/` (`review_submission.py:278`, `maintainer_review.py:246`, `generate_formal_scorecard.py:160`) and then accept a bare `parser.add_argument("--out")` with no validation at all (`review_submission.py:266`, `maintainer_review.py:232`, `generate_formal_scorecard.py:146`). `generate_formal_scorecard.py` is the worst of the three by volume — it writes `formal-scorecard.json`, `formal-scorecard-comment.md`, `maintainer-comment.md`, `advisory-review.md`, `review-input.json`, `review-prompt.md` and `review-summary.json` straight into the participant's directory.

So the boundary this PR documents from `docs/maintainer-workflow.md` is enforced on 2 of 5 relevant entry points after this merges. That is a strict improvement and I would not hold the PR for it — but a shared helper (point 4) applied to all five would close the class rather than one instance, and it is the difference between "this command is safe" and "the boundary is enforced".

### 6. Test coverage: the regression does not cover the reproduction

Bidirectional check:

```
main + new test file:  ERROR — ImportError: cannot import name 'validate_output_dir' from 'export_review_packet'
head:                  2 passed
```

The on-main failure is an import error, which proves the symbol is new rather than that the behaviour was wrong. Both new tests call `validate_output_dir()` directly; neither invokes `export_review_packet()`. So the scenario in your Reproduction section — the exporter actually writing three files into a submission — has no regression test guarding it, and the guard could be silently detached from the call site (e.g. moved below `out_dir.mkdir`) with both tests still green.

A single extra test would close that: call `export_review_packet(repo_root, [submission_dir], repo_root / "submissions" / owner / slug, "...")`, assert it raises, and assert `(submission_dir / "review-packet.md").exists()` is `False`. Your #2212 already does exactly this shape in `test_existing_invalid_json_aborts_before_creating_output` (asserting `out_dir.exists()` is `False` after the raise), so the pattern is right there.

Minor: `self.assertFalse(out_dir.exists())` in `test_submission_output_is_rejected_before_writes` sits **outside** the `with tempfile.TemporaryDirectory()` block, so by the time it runs the whole temp tree is already gone and the assertion is vacuous. Moving it inside the `with` makes it mean what it says.

### 7. Regression, and interaction with your #2212

- Full suite, identical environment: main `9 failed, 395 passed` → head `9 failed, 397 passed`; the two failure sets diff empty. Zero regression. (The 9 are artifacts of my checkout not materialising `submissions/`, not repository breakage.)
- #2212 edits the same file from the same pre-image blob. I merged the two heads to check: `git merge-file` returns 0 with no conflict markers, the combined file compiles, all 7 tests pass (`test_review_packet_output_boundary` 2 + `test_review_packet_json_integrity` 2 + `test_export_review_packet` 3), and I confirmed end-to-end that the combined build rejects both `--out` inside `submissions/` (exit 1) and a corrupt `sources.json` (exit 1). Your "this is separate from #2209" note is accurate and they can land in either order.

### 8. Verification I ran

```bash
git ls-tree 4bb9b75e90cfdbcd85836882550dcbb1fa33fae9 -- scripts/export_review_packet.py tests/test_review_packet_output_boundary.py
#   7cb659fef73f4fd8a04868753b47f3f6cd414ea3  scripts/export_review_packet.py
#   4076c1be5619bc8943300897743981d1294be5ad  tests/test_review_packet_output_boundary.py

python3 -m unittest tests.test_review_packet_output_boundary -v   # main: ImportError; head: 2 pass
python3 -m pytest tests -q                                        # main 9F/395P -> head 9F/397P, same failure set
# plus the --out matrix in §3 and the sibling-script matrix in §5, both against a real merged package
```

Thanks for taking the destination-path half separately from #2212 — the split made both easy to test in isolation, and the guard held up against traversal and symlinks, which is more than the description claims.

PR #2211, opened 2026-08-12
fix: publish valid source drafts transactionally

jiangmuran:
## Problem

`prepare_source_registry_draft.py` wrote `--out` before validating the generated registry. If a discovery row produced an invalid record, such as a `public_url` with an `ftp://` URL, the command correctly returned 1 but had already replaced an existing maintainer draft with invalid output.

```bash
printf '%s\n' 'id,title,url,topic' \
  'bad,Bad source,ftp://example.com/source,test' > seed.csv
printf '%s\n' '{"existing":"maintainer draft"}' > draft.json
python3 scripts/prepare_source_registry_draft.py \
  --input seed.csv --out draft.json --json
# exits 1, but draft.json now contains the invalid generated registry
```

## Change

- serialize the candidate into a temporary file beside the requested output;
- validate that exact serialized file;
- atomically replace `--out` only after validation succeeds;
- remove the temporary file on validation failure while preserving any existing output.

The successful generation path and JSON summary remain unchanged.

## Verification

- `python3 -m unittest tests.test_source_registry_draft tests.test_data_registry -v` - 8 passed
- `python3 -m py_compile scripts/prepare_source_registry_draft.py tests/test_source_registry_draft.py`
- `git diff --check`

### DIFF
--- a/scripts/prepare_source_registry_draft.py
+++ b/scripts/prepare_source_registry_draft.py
@@ -13,7 +13,9 @@ import csv
 import datetime as dt
 import hashlib
 import json
+import os
 import re
+import tempfile
 import urllib.parse
 from pathlib import Path
 from typing import Any
@@ -24,7 +26,7 @@ from discover_public_sources import (
     content_type_kind,
     parse_dateish,
 )
-from validate_data_registry import validate_registry
+from validate_data_registry import RegistryReport, validate_registry
 
 
 DEFAULT_DISCOVERY_CSV = "brief/discovery/candidate-sources.csv"
@@ -252,10 +254,35 @@ def build_draft(
     }
 
 
-def write_registry_json(data: dict[str, Any], path: Path) -> None:
+def serialize_registry_json(data: dict[str, Any]) -> str:
     serializable = {key: value for key, value in data.items() if not key.startswith("_")}
+    return json.dumps(serializable, ensure_ascii=False, indent=2) + "\n"
+
+
+def validate_and_publish_draft(
+    data: dict[str, Any], repo_root: Path, path: Path
+) -> RegistryReport:
     path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(json.dumps(serializable, ensure_ascii=False, indent=2) + "\n", encoding="utf-8")
+    temporary: Path | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            mode="w",
+            encoding="utf-8",
+            dir=path.parent,
+            prefix=f".{path.name}.",
+            suffix=".tmp",
+            delete=False,
+        ) as stream:
+            stream.write(serialize_registry_json(data))
+            temporary = Path(stream.name)
+        report = validate_registry(repo_root, temporary)
+        if report.ok:
+            os.replace(temporary, path)
+            temporary = None
+        return report
+    finally:
+        if temporary is not None:
+            temporary.unlink(missing_ok=True)
 
 
 def main() -> int:
@@ -293,8 +320,7 @@ def main() -> int:
         limit=args.limit,
         include_existing=args.include_existing,
     )
-    write_registry_json(draft, out_path)
-    report = validate_registry(repo_root, out_path)
+    report = validate_and_publish_draft(draft, repo_root, out_path)
     summary = {
         "ok": report.ok,
         "input": input_path.relative_to(repo_root).as_posix() if input_path.is_relative_to(repo_root) else str(input_path),
--- a/tests/test_source_registry_draft.py
+++ b/tests/test_source_registry_draft.py
@@ -20,6 +20,49 @@ def write_csv(path: Path, rows: list[dict[str, str]], fields: list[str]) -> None
 
 
 class SourceRegistryDraftTests(unittest.TestCase):
+    def test_invalid_draft_does_not_replace_existing_output(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            (root / "data").mkdir()
+            existing_path = root / "data" / "source_registry.json"
+            existing_path.write_text(
+                json.dumps({"schema_version": "0.1.0", "updated_date": "2026-08-12", "sources": []}),
+                encoding="utf-8",
+            )
+            seed_path = root / "seed.csv"
+            write_csv(
+                seed_path,
+                [{"id": "bad", "title": "Bad source", "url": "ftp://example.com/source", "topic": "test"}],
+                ["id", "title", "url", "topic"],
+            )
+            out_path = root / "draft.json"
+            original = b'{"existing":"maintainer draft"}\n'
+            out_path.write_bytes(original)
+
+            completed = subprocess.run(
+                [
+                    sys.executable,
+                    str(REPO_ROOT / "scripts" / "prepare_source_registry_draft.py"),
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2211 review draft (English; COMMENT)

## Draft body to publish

- checked exact head: `47b56a09dcc9cbba2f98f0ab6cac9bbe23bee424`

`submission-validation.yml` is `pull_request_target` with one checkout step pinned to `ref: ${{ github.event.repository.default_branch }}` — main's scripts, never your head. The only `python3 -m unittest` in any workflow is `gallery-snapshot-maintenance.yml:68`, which names three modules explicitly and runs on push-to-main.

So the new `test_invalid_draft_does_not_replace_existing_output` is never executed by any workflow, before or after merge. Everything below is a local run.

### 2. The reproduction is exact

Your command, verbatim:

```
# main d05f30a4
$ python3 scripts/prepare_source_registry_draft.py --input seed.csv --out draft.json --json
exit=1
$ head -c 90 draft.json
{
  "schema_version": "0.1.0",
  "updated_date": "2026-08-13",
  "sources": [
    {
$ stat -c %s draft.json
1428          # was 31 — the maintainer draft is gone

# head 47b56a09, identical command
exit=1
$ cat draft.json
{"existing":"maintainer draft"}     # preserved
$ ls -a | grep tmp
(none)
```

The command still exits 1, the `--json` summary is unchanged, and the validation errors are keyed by source id (`DRAFT-BAD: public URL sources must use an http(s) url`) rather than by path, so the temporary filename never leaks into the output a maintainer reads. On the success path, main and head produce **byte-identical** output (1453 B, `diff` empty). All three of those are things I would have wanted to check before trusting a write-path change, and they hold.

I also confirmed the approach is sound in the way that matters: `validate_registry(repo_root, registry_path)` (`scripts/validate_data_registry.py:190-211`) derives everything from file *content* and never inspects the filename, so validating `.draft.json.xxxxx.tmp` is equivalent to validating `draft.json`. There is no behaviour that changes just because the path is temporary.

### 3. One unclaimed behaviour change: the published file's mode drops to 0600

This is the only thing I found that is not in the description. With `umask 0022`:

| | resulting mode of `--out` |
|---|---|
| main, new file | `0644` |
| head, new file | **`0600`** |
| head, `--out` already existed at `0644` | **`0600`** (downgraded) |

`tempfile.NamedTemporaryFile` creates its file with mode `0600` by design, and `os.replace()` preserves the source file's mode, so the mode of the temp file becomes the mode of the published draft — including on a file that already existed with broader permissions.

Scoping it honestly: `--out` defaults to `data/source_registry.draft.json`, which is gitignored (`.gitignore:9`), and git does not track read/write bits anyway, so **this cannot reach the repository**. It matters only for a local or CI checkout where another user or a static server has to read the generated file. That makes it a nit, not a blocker — but it is a silent difference between "before" and "after" on the success path, and the fix is one line before the `os.replace`:

```python
os.chmod(temporary, 0o666 & ~current_umask)     # or: shutil.copymode(path, temporary) when path exists
```

Worth noting you already have better prior art in this repo for exactly this: `scripts/ai_review_submission.py:99-104`

```python
def atomic_write_text(path: Path, text: str) -> None:
    path.parent.mkdir(parents=True, exist_ok=True)
    temporary = path.with_name(f".{path.name}.tmp-{os.getpid()}")
    temporary.write_text(text, encoding="utf-8")
    temporary.replace(path)
```

`write_text` creates under the normal umask, so that version has no mode change. It also has no `fsync`, which brings me to the last small point.

### 4. Same-class: the pattern you copied already has this side effect elsewhere

`scripts/refresh_submission_manifest.py:65-70` uses the identical `NamedTemporaryFile(dir=..., delete=False)` + `os.replace` shape, and it has the same consequence on a **participant-owned** file. Measured on main, no patch applied:

```
$ stat -c %a submissions/AlanHN/jingzhang-commons-line/manifest.json
644
$ python3 scripts/refresh_submission_manifest.py submissions/AlanHN/jingzhang-commons-line
exit=0
$ stat -c %a submissions/AlanHN/jingzhang-commons-line/manifest.json
600
```

So this is a pre-existing repo pattern, not something this PR introduces — which I think is the fairer framing, and it also means a single shared `atomic_write_json(path, text)` helper would fix both sites at once instead of adding a third copy of the idiom. There are now three separate atomic-write implementations in `scripts/` (`ai_review_submission.py:99`, `refresh_submission_manifest.py:65`, and this one) with three different sets of properties.

Durability nit while in the area: `refresh_submission_manifest.py` does `handle.flush()` + `os.fsync(handle.fileno())` before `os.replace`; the new `validate_and_publish_draft()` does neither. `os.replace` is atomic with respect to the rename, so a crash cannot leave a half-replaced `--out` — but without the fsync a crash between write and rename can leave a truncated temp file on disk. Given the `finally` block only unlinks the temp when the function returns normally, that stale file would survive. Very low severity for a draft generator; mentioning it because you explicitly describe the change as atomic.

### 5. Test quality: this one is properly anchored

```
main + head's test file:  1 FAIL, 2 pass
head:                     3 pass
```

The on-main failure is the real symptom rather than an import error:

```
AssertionError: b'{\n  "schema_version": "0.1.0",\n  "update[1904 chars]n}\n' != b'{"existing":"maintainer draft"}\n'
```

The test also drives the real CLI through `subprocess.run` rather than calling the function directly, asserts the exit code, asserts the JSON summary's `ok` is `False`, asserts the output bytes are unchanged, **and** asserts no `.draft.json.*.tmp` leftovers. That last assertion is the one that would catch a regression in the `finally` block, and it is easy to leave out. Nice.

The one case it does not cover is the success path — that a valid seed still replaces `--out` and produces the same bytes as before. `test_seed_csv_generates_needs_review_draft_and_skips_existing_urls` covers generation but predates the rewrite, so nothing asserts that `os.replace` actually fires. I checked it by hand (byte-identical output, §2) and it is fine; an assertion on `out_path.read_bytes()` after a *successful* run would lock it in, and would also be the natural place to assert the file mode if you decide to address §3.

### 6. Regression

Full suite, identical environment: main `9 failed, 395 passed` → head `9 failed, 396 passed`. The two failure sets diff empty, so the +1 is exactly your new test. (The 9 pre-existing failures are artifacts of my checkout not materialising `submissions/`, not repository breakage.)

No other open PR touches `scripts/prepare_source_registry_draft.py` or `scripts/validate_data_registry.py`, so there is nothing to coordinate here.

### 7. Verification I ran

```bash
git ls-tree 47b56a09dcc9cbba2f98f0ab6cac9bbe23bee424 -- scripts/prepare_source_registry_draft.py tests/test_source_registry_draft.py
#   edfe33f5c72e1c96b1ae86d9b582c634820e2529  scripts/prepare_source_registry_draft.py
#   b4f368f61dc94874a7947715d9ce8c4125e23cce  tests/test_source_registry_draft.py

python3 -m unittest tests.test_source_registry_draft -v   # main: 1 FAIL / 2 pass; head: 3 pass
python3 -m pytest tests -q                                # main 9F/395P -> head 9F/396P, same failure set
# plus the clobber repro, the success-path byte diff, and the mode matrix in §3
```

Thanks — write-then-validate is a genuinely easy thing to miss, and the `.tmp` leftover assertion in the new test is the detail that makes the regression worth having. The mode point is minor and gitignored, so treat it as a note rather than a request.

PR #2212, opened 2026-08-12
fix: reject corrupt review packet evidence

jiangmuran:
## Problem

`export_review_packet.py` treated an existing JSON file that was malformed or not UTF-8 exactly like an optional file that was absent. A corrupt `sources.json`, `assumptions.json`, `metrics.json`, `risk.json`, manifest, self-check, or agent record was silently rendered as empty data, and the command still emitted a complete-looking review packet and manifest.

For example, replacing `sources.json` with `{"sources": [` currently exits successfully and renders the public-source section as `暂无`, hiding the evidence integrity failure from reviewers.

## Change

- keep missing optional JSON files optional;
- report existing non-UTF-8, unreadable, malformed, or non-object JSON with the affected path;
- abort while loading submissions, before the packet output directory is created.

## Verification

- `python3 -m unittest tests.test_review_packet_json_integrity tests.test_export_review_packet -v` - 5 passed
- regressions cover malformed JSON, invalid UTF-8, non-object roots, output preflight, and a genuinely absent optional `risk.json`
- `python3 -m py_compile scripts/export_review_packet.py tests/test_review_packet_json_integrity.py`
- `git diff --check`

This is separate from #2209: that PR constrains the destination path, while this change validates the source evidence before rendering.

### DIFF
--- a/scripts/export_review_packet.py
+++ b/scripts/export_review_packet.py
@@ -70,13 +70,20 @@ class SubmissionPacket:
     key_files: list[dict[str, Any]]
 
 
-def read_json(path: Path) -> Any:
+def read_json(path: Path) -> dict[str, Any] | None:
     if not path.exists():
         return None
     try:
-        return json.loads(path.read_text(encoding="utf-8"))
-    except (UnicodeDecodeError, json.JSONDecodeError):
-        return None
+        data = json.loads(path.read_text(encoding="utf-8"))
+    except UnicodeDecodeError as exc:
+        raise ReviewPacketError(f"{path}: JSON file must be UTF-8") from exc
+    except json.JSONDecodeError as exc:
+        raise ReviewPacketError(f"{path}: invalid JSON: {exc}") from exc
+    except OSError as exc:
+        raise ReviewPacketError(f"{path}: could not read JSON: {exc}") from exc
+    if not isinstance(data, dict):
+        raise ReviewPacketError(f"{path}: JSON root must be an object")
+    return data
 
 
 def as_dict(value: Any) -> dict[str, Any]:
--- a/tests/test_review_packet_json_integrity.py
+++ b/tests/test_review_packet_json_integrity.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+
+from tests.test_export_review_packet import make_submission
+
+from scripts.export_review_packet import ReviewPacketError, export_review_packet
+
+
+class ReviewPacketJsonIntegrityTests(unittest.TestCase):
+    def test_existing_invalid_json_aborts_before_creating_output(self) -> None:
+        invalid_files = {
+            "sources.json": b'{"sources": [',
+            "risk.json": b"\xff\xfe",
+            "manifest.json": b"[]",
+        }
+        for filename, content in invalid_files.items():
+            with self.subTest(filename=filename), tempfile.TemporaryDirectory() as tmp:
+                repo_root = Path(tmp)
+                submission_dir = make_submission(repo_root, "alice", "proposal-a", "AI network")
+                (submission_dir / filename).write_bytes(content)
+                out_dir = repo_root / "packet"
+
+                with self.assertRaisesRegex(ReviewPacketError, filename):
+                    export_review_packet(repo_root, [submission_dir], out_dir, "Review packet")
+
+                self.assertFalse(out_dir.exists())
+
+    def test_missing_optional_risk_json_still_exports(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            repo_root = Path(tmp)
+            submission_dir = make_submission(repo_root, "alice", "proposal-a", "AI network")
+            (submission_dir / "risk.json").unlink()
+            out_dir = repo_root / "packet"
+
+            files = export_review_packet(repo_root, [submission_dir], out_dir, "Review packet")
+
+            self.assertTrue(Path(files["manifest"]).is_file())
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2212 review draft (English; COMMENT)

## Draft body to publish

- checked exact head: `278e193bf9f235b0ffe98418934668b872ea16a9`

`submission-validation.yml` is `pull_request_target` with one checkout step pinned to `ref: ${{ github.event.repository.default_branch }}` — it runs main's scripts against your changed files and never checks out your head. The only `python3 -m unittest` in any workflow is `gallery-snapshot-maintenance.yml:68`, which names three modules explicitly and runs on push-to-main.

So `tests/test_review_packet_json_integrity.py` is never executed by any workflow, before or after merge. Everything below is a local run. (This one matters more than usual — see §4, where the file has an invocation mode that is already broken and nothing would have told you.)

### 2. The reproduction holds on a real merged package

Not a fixture — `submissions/AlanHN/jingzhang-commons-line`, pulled at the pinned main sha and blob-verified, with `sources.json` replaced by `{"sources": [`:

```
# main d05f30a4
exit=0
  review-packet.md, review-packet.html, packet-manifest.json  -> all produced
  review-packet.md:62   _暂无。_          # the public-source section

# head 278e193b, identical corruption
error: .../sources.json: invalid JSON: Expecting value: line 1 column 14 (char 13)
exit=1
  output directory: not created
```

Both halves of your claim check out: main renders a complete-looking packet with an empty evidence section and exit 0, and head aborts *before* the output directory exists. A clean run of the same real package on head still exits 0 and produces all three artifacts, so the happy path is untouched.

### 3. Blast radius on the merged corpus is exactly zero

This is the number I would want before merging a change from "silently degrade" to "abort", so I measured it rather than assuming. Across **all 713 merged packages** that carry `self_check.json`, for every one of the seven files `read_json()` touches:

| file | present | absent | non-UTF-8 | malformed | non-object root |
|---|---|---|---|---|---|
| `sources.json` | 713 | 0 | 0 | 0 | 0 |
| `assumptions.json` | 713 | 0 | 0 | 0 | 0 |
| `metrics.json` | 713 | 0 | 0 | 0 | 0 |
| `risk.json` | 110 | 603 | 0 | 0 | 0 |
| `manifest.json` | 713 | 0 | 0 | 0 | 0 |
| `self_check.json` | 713 | 0 | 0 | 0 | 0 |
| `agent.json` | 713 | 0 | 0 | 0 | 0 |

**0 of 713 packages would newly fail.** The one file with meaningful absence is `risk.json` (603 packages do not ship it), which is precisely the case your `test_missing_optional_risk_json_still_exports` pins — so the "keep missing optional JSON files optional" half of the change is the half that actually carries the corpus, and it is covered. That combination is what makes this a safe tightening rather than a gamble, and it seems worth putting the number in the PR body.

### 4. The new test file is the only one in `tests/` that imports as a package, and it breaks its own entry point

`tests/test_review_packet_json_integrity.py:1-9`:

```python
from tests.test_export_review_packet import make_submission
from scripts.export_review_packet import ReviewPacketError, export_review_packet
```

It is the only one of 38 files in `tests/` that does this. The other 26 that need `scripts/` use the repo's established shape (`sys.path.insert(0, str(ROOT / "scripts"))` then a bare import). Two measured consequences:

**a) The file's own `__main__` block does not work.** It ends with `if __name__ == "__main__": unittest.main()`, but:

```
$ python3 tests/test_review_packet_json_integrity.py
ModuleNotFoundError: No module named 'tests'

# controls, same invocation, same tree:
$ python3 tests/test_export_review_packet.py     -> Ran 3 tests ... OK
$ python3 tests/test_data_registry.py            -> Ran 5 tests ... OK
```

Under `python3 -m unittest tests.…` and under pytest it works fine (repo root ends up on `sys.path`, and the `from tests.…` line happens to insert `scripts/` as a side effect before the `from scripts.…` line is reached), which is why your Verification section passes. But direct execution is a mode the file advertises and every sibling supports.

**b) Two copies of the module end up loaded.** Because `tests/test_export_review_packet.py` imports it as `export_review_packet` and this file imports it as `scripts.export_review_packet`, a full run has both in `sys.modules` as distinct objects:

```
module objects identical?    False
ReviewPacketError identical? False
raise B.ReviewPacketError -> except A.ReviewPacketError:   NOT caught
```

Nothing breaks today — each test catches the class from the module it imported, and all 5 pass — but it means the exception type in this file is not the same type the rest of the suite uses, and the next test that mixes the two import styles will get an uncaught `ReviewPacketError` with a confusing traceback. Switching to the repo convention removes both problems and the `from tests.…` line at the same time:

```python
ROOT = Path(__file__).resolve().parents[1]
sys.path.insert(0, str(ROOT / "scripts"))
sys.path.insert(0, str(ROOT))
from test_export_review_packet import make_submission   # or keep tests. after inserting ROOT
from export_review_packet import ReviewPacketError, export_review_packet
```

### 5. Test quality otherwise: well anchored

```
main + new test file:  test_existing_invalid_json_aborts_before_creating_output
                       -> 3 subtest FAILs (sources.json, risk.json, manifest.json), "ReviewPacketError not raised"
                       test_missing_optional_risk_json_still_exports -> pass
head:                  2 pass, 3 subtests pass
```

The three subtests cover the three genuinely distinct branches you added (malformed JSON, invalid UTF-8 via `b"\xff\xfe"`, non-object root via `b"[]"`), each fails on main for the right reason, and `assertFalse(out_dir.exists())` inside the temp-dir scope pins the "abort before the output directory is created" ordering that the description leads with. That last assertion is the one that would catch someone later moving `out_dir.mkdir()` above the load loop, and it is placed correctly.

Two small gaps: the `OSError` branch you added (`could not read JSON`) has no test — a `chmod 000` file, or a directory where a JSON file is expected, would exercise it. And `self_check.json` / `agent.json` are in the corrupted-file rotation only implicitly; `manifest.json` is covered, the other two are not. Neither is a correctness concern, just uncovered lines.

### 6. Same-class: five other call sites still swallow corrupt JSON

Since the argument for this change is "a corrupt evidence file must not read as an empty one", the same pattern elsewhere is worth listing:

| site | on malformed JSON |
|---|---|
| `scripts/score_submission.py:189-193, 208-212` | returns `[]` |
| `scripts/generate_submissions_data.py:86-90` | returns `None` |
| `scripts/spatial_review.py:568-571` | returns `"formal"` |
| `scripts/self_check_submission.py:78-79` | returns `"unknown"` |
| `scripts/github_pr_validation.py:461` | swallowed |

`score_submission.py` is the one I would look at next: a corrupt `sources.json` or `metrics.json` returns an empty list there, which is the scoring-path equivalent of the `暂无` you are fixing here — same silent "no evidence" outcome, but attached to a score rather than a rendered section. `spatial_review.py:571` defaulting to `"formal"` on a decode failure is the other one worth a look, since it fails toward the *more* permissive value.

The repo already contains both of the better idioms, if you want precedent: `scripts/review_submission.py:75` and `scripts/source_registry_utils.py:18` return `{"_error": f"invalid JSON: {exc}"}` (surface it, keep going), and `scripts/professional_review.py:63-68` re-raises (what you did here). Three different conventions for the same situation across `scripts/` is probably the real underlying issue; this PR moves one call site to the strictest of them, which for a maintainer-facing review packet seems like the right pick.

### 7. Regression, and interaction with your #2209

- Full suite, identical environment: main `9 failed, 395 passed` → head `9 failed, 397 passed` (+3 subtests). The two failure sets diff empty. Zero regression. (The 9 are artifacts of my checkout not materialising `submissions/`, not repository breakage.)
- Your "This is separate from #2209" note is accurate, and I verified it rather than taking it on faith: both PRs modify `scripts/export_review_packet.py` from the same pre-image blob `85bf74cc64`. `git merge-file` returns 0 with no conflict markers, the combined file compiles, all 7 tests pass (2 + 2 + 3), and I confirmed end-to-end on the real package that the combined build rejects both `--out` inside `submissions/` (exit 1) and a corrupt `sources.json` (exit 1). They can land in either order.

### 8. Verification I ran

```bash
git ls-tree 278e193bf9f235b0ffe98418934668b872ea16a9 -- scripts/export_review_packet.py tests/test_review_packet_json_integrity.py
#   d6e2692031e919caf464258ebaa4379bb9d62e95  scripts/export_review_packet.py
#   243660ededb829881a7c296edf8a248d5449b5f5  tests/test_review_packet_json_integrity.py

python3 -m unittest tests.test_review_packet_json_integrity -v   # main: 3 subtest FAILs; head: 2 pass
python3 -m pytest tests -q                                       # main 9F/395P -> head 9F/397P, same failure set
python3 tests/test_review_packet_json_integrity.py               # ModuleNotFoundError (see §4a)
# plus the real-package corruption repro (§2) and the 713 x 7 corpus scan (§3)
```

Thanks for keeping the missing-vs-corrupt distinction explicit — that is the part that makes this safe, and the corpus scan bears it out: 603 of 713 packages rely on `risk.json` staying optional, and none of them regress.

CocoSgt (approved):
Approved on exact head `278e193bf9f235b0ffe98418934668b872ea16a9` against current main `ee22cbf9b0e1facb5302e59236187b33140d2fd4`.

Independent maintainer verification in an isolated worktree:
- real no-commit merge: clean
- Python 3.9.6: `tests.test_review_packet_json_integrity` + `tests.test_export_review_packet` — 6/6 PASS
- `py_compile` and `git diff --check` — PASS
- corrupt/non-UTF-8/non-object existing JSON fails closed before output creation; absent optional `risk.json` remains supported

This is a narrow review-packet integrity fix and does not alter participant scope, scoring, or merge policy. The test file’s direct-script import style is a non-blocking cleanup; the project-standard module invocation passes.

PR #2213, opened 2026-08-12
fix: match portal collections by path identity

jiangmuran:
## Problem

`render_portal.py` stores each card's `source_path` exactly as its proposal argument. The documented CLI accepts paths, so an absolute proposal path produces an absolute card key. Collection items use validated repository-relative references such as `examples/agent-civic-loop`, and their string lookup therefore misses the loaded card.

The main card still renders, masking the defect, while the collection section shows an empty title, `未载入 portal`, and no detail link.

```bash
python3 scripts/render_portal.py \
  --output /tmp/portal.html \
  --collections-dir collections \
  "$PWD/examples/agent-civic-loop"
```

## Change

Resolve both loaded card paths and validated collection proposal references against the collection directory's repository root, then match their filesystem identities. Relative proposal arguments continue to work, while absolute arguments now populate the same collection entry and detail URL.

## Verification

- `python3 -m unittest tests.test_collections tests.test_scenarios -v` - 3 passed
- the collection regression uses an absolute proposal path and now checks loaded state, detail URL, rendered link, and absence of the unloaded placeholder
- `python3 -m py_compile scripts/render_portal.py tests/test_collections.py`
- `git diff --check`

A broader gallery run reached unrelated partial-clone baseline failures because this lightweight checkout does not contain the hundreds of submission blobs referenced by `submissions-data.js`; the portal collection and scenario tests are dependency-free and pass.

### DIFF
--- a/scripts/render_portal.py
+++ b/scripts/render_portal.py
@@ -266,7 +266,16 @@ def load_collections(collections_dir: Path | None, cards: list[dict[str, Any]]) 
     if not collections_dir.is_dir():
         raise PortalError(f"{collections_dir}: collections path must be a directory")
 
-    card_lookup = {card["source_path"]: card for card in cards}
+    repo_root = collections_dir.resolve().parent
+
+    def proposal_identity(raw_path: str) -> Path:
+        path = Path(raw_path)
+        return (path if path.is_absolute() else repo_root / path).resolve()
+
+    card_lookup = {
+        proposal_identity(text(card["source_path"])): card
+        for card in cards
+    }
     collections: list[dict[str, Any]] = []
     for path in sorted(collections_dir.glob("*.json")):
         try:
@@ -286,7 +295,7 @@ def load_collections(collections_dir: Path | None, cards: list[dict[str, Any]]) 
             if not isinstance(item, dict):
                 raise PortalError(f"{path}: each collection item must be an object")
             proposal_ref = safe_proposal_ref(item.get("proposal"))
-            matched = card_lookup.get(proposal_ref)
+            matched = card_lookup.get(proposal_identity(proposal_ref))
             rendered_items.append(
                 {
                     "proposal": proposal_ref,
--- a/tests/test_collections.py
+++ b/tests/test_collections.py
@@ -82,10 +82,18 @@ tracks: ["civic-agent-governance"]
             collections = load_collections(collections_dir, [card])
             html = render_portal([card], "Collection Portal", collections=collections)
 
+            collection_item = collections[0]["items"][0]
+            self.assertTrue(collection_item["loaded"])
+            self.assertEqual(
+                collection_item["url"],
+                "../examples/agent-civic-loop/index.html",
+            )
             self.assertIn("精选专题", html)
             self.assertIn("最佳 AI 治理", html)
             self.assertIn("治理闭环", html)
             self.assertIn("Collection Test", html)
+            self.assertIn('../examples/agent-civic-loop/index.html', html)
+            self.assertNotIn("未载入 portal", html)
 
 
 if __name__ == "__main__":


anselasimov-web (commented):
# PR #2213 review draft

- 仓库 / PR：open-city-ai/haidian#2213「fix: match portal collections by path identity」
- 作者：`jiangmuran`
- checked exact head（全长）：`90a1aed25a75c2bfa39a915aec2cb2c1b4826813`
- merge-base（全长）：`3a3884766a28f67de49460f0709302b03d3e8bbd`
- 判级：**COMMENT**

---

## 拟发表正文（English）

Hi @jiangmuran — I am a **fellow participant in this competition, not a maintainer**. I have no review or merge authority; everything below is just what I measured locally, offered for your consideration.

**Exact head reviewed: `90a1aed25a75c2bfa39a915aec2cb2c1b4826813`** (merge-base `3a3884766a28f67de49460f0709302b03d3e8bbd`, baseline main `f992bdb02ebd4e0dd0009d82ee7847ba4a99eec4`).

Short version: **the defect is real, the fix works, and I found no regression.** Two things I would still like to raise — the choice of repository-root anchor, and one path shape the test does not cover.

### 1. Reproduction on `main` — confirmed, and the failure is exactly what you described

Running the documented command from `README.md`, once with a relative proposal argument and once with an absolute one, and diffing the two portal outputs on `main`:

```
diff portal_relative.html portal_absolute.html
```

Exactly **one line** differs — the collection card:

```
relative: <li><div><a href=".../examples/agent-civic-loop/index.html">京张城市智能体环</a></div>...
absolute: <li><div><span></span><small>未载入 portal</small></div>...
```

So the empty title, the `未载入 portal` placeholder, and the missing detail link all land on that one item, and the main card section still renders — which is exactly the masking effect your description calls out. On the head commit, the absolute-argument output is **byte-identical** to the relative-argument output.

I also ran the whole thing from a working directory outside the repository, passing an absolute `--collections-dir` and an absolute proposal directory (the shape a maintainer script would most plausibly use): `main` leaves the item unloaded, the head commit loads it.

### 2. Test does discriminate in both directions

- The updated `tests/test_collections.py` run against **`main`'s** `scripts/render_portal.py`: **FAIL** — `AssertionError: False is not true` at `self.assertTrue(collection_item["loaded"])`.
- The same test against your head: **PASS**.
- `main`'s original `tests/test_collections.py` against your head: still **PASS** (no assertion was weakened, only added).

### 3. Full-suite baseline: zero regression

`python3 -m unittest` over all 39 `tests/test_*.py` modules:

| | tests | failures/errors |
|---|---|---|
| `main` f992bdb0 | 404 | 9 |
| head 90a1aed2 | 404 | 9 |

The failure sets are **identical line for line**. (My checkout is a partial clone without the `submissions/` blobs, so those 9 are the gallery / prelaunch / site-package / bilingual-backfill tests failing on both sides for the same reason you hit — they are baseline noise, not signal.)

### 4. One thing I would reconsider: which directory is "the repository root"

```python
repo_root = collections_dir.resolve().parent
```

This infers the repository root from `--collections-dir`. But a few lines further down in the same `main()`, the run is already anchored somewhere else:

```python
track_registry = load_track_registry(Path.cwd())
scenario_registry = load_scenario_registry(Path.cwd())
```

So the file now carries two different notions of "root". They agree only when the collections directory happens to be a direct child of the repository root. Measured, with `--collections-dir sub/collections`:

| anchor | relative proposal arg | absolute proposal arg |
|---|---|---|
| `collections_dir.parent` (this PR) | matches | **does not match** |
| `Path.cwd()` | matches | matches |

I patched the one line to `repo_root = Path.cwd().resolve()` locally and re-ran three configurations (standard layout + absolute arg, nested collections + relative arg, nested collections + absolute arg) — all three matched. It is also one fewer concept in the file. Note this is **not a regression** in your PR: `main` fails the nested case too, because the string comparison never matched there either. It is just that the fix would generalize for free.

### 5. Coverage gap: `submissions/<author>/<slug>` refs are never exercised

`safe_proposal_ref()` accepts two shapes, `examples/<slug>` **and** `submissions/<author>/<slug>`, and `docs/maintainer-workflow.md` documents the portal command with the `submissions/` form. The regression only covers the `examples/` shape. The `submissions/` path is the one maintainers actually run, and it is the one whose card `source_path` is built from a deeper `proposal_dir`, so it would be worth a second item in the collection fixture.

### 6. Heads-up: CI will not run this test

Worth flagging because it changes what the diff is worth. Neither PR-triggered workflow executes `tests/`:

- `submission-validation.yml` is `pull_request_target` and checks out `ref: ${{ github.event.repository.default_branch }}` — trusted `main`, never the PR head — then runs only `scripts/github_pr_validation.py`.
- `gallery-snapshot-maintenance.yml` is the only workflow that calls `python3 -m unittest`, but it triggers on `push` to `main` / `workflow_dispatch` and runs exactly three modules: `tests.test_submissions_gallery`, `tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`.

So `tests/test_collections.py` is never executed by CI — not on this PR, and not after merge. Manual execution is the only gate this regression will ever get, which is why I ran it in both directions rather than trusting the green check.

---

Thanks for the careful write-up — the reproduction command in the description worked verbatim on the first try, and the one-line diff between the two outputs made it very easy to confirm exactly what breaks. If any of the above is already on your list, please ignore.

---

## 复核方法备注（内部，不发布）

1. 工作目录 `$HOME/z4_18714`（唯一目录）。`git clone --filter=blob:none` + sparse-checkout 排除 `submissions/`（29992 个 blob，磁盘不足以全量检出；作者 PR 正文也提到同一限制）。
2. main 快照：`tar` 复制工作树 → 对 `git ls-tree -r main -z` 逐条比对 blob sha1（含 mode 120000 的 symlink 按 readlink 内容算）：**checked=235，missing=0，mismatch=0**，跳过 submissions/ 30013 条。
3. PR 快照：由已核对的 main 快照复制，再用 `git cat-file blob pr2213:<path>` 覆盖两个文件，写盘后二次 `hash-object` 复核：`75a3a8a8…` / `02d4bac0…` 均 OK。
4. 复现命令（README.md:335-343 原样）：`python3 scripts/render_portal.py --output <out> --collections-dir collections <proposal_dir>`，分别传相对与绝对参数；用 `sed` 归一化快照目录名后 `diff`。
5. 双向测试：把 head 版 `tests/test_collections.py` 拷到 main 快照副本上跑（FAIL），把 main 版拷到 head 快照副本上跑（PASS）。
6. 全量基线：`python3 -m unittest $(ls tests/test_*.py → tests.<mod>)`，两侧 `grep -E "^(FAIL|ERROR):"` 排序后 `diff` 为空。
7. 锚点实验：本地把 `repo_root = collections_dir.resolve().parent` 改成 `Path.cwd().resolve()`，在 `sub/collections` 布局下重跑绝对/相对参数，`grep -c '未载入 portal'` 均为 0。
8. CI 结论来自读 `.github/workflows/*.yml` 全文：`grep -rn "unittest\|pytest" .github/workflows/` 仅命中 gallery-snapshot-maintenance.yml:68。
9. 未发布任何 GitHub 内容；gh 仅用于 `api GET` / `pr view`。

CocoSgt (approved):
Reviewed exact head `90a1aed25a75c2bfa39a915aec2cb2c1b4826813` against current main `426c05b4e3f5cba81364a8c68e4291bfd7d67ae8`.

Confirmed the documented absolute-proposal-path failure and the resolved path-identity fix. `safe_proposal_ref()` still rejects absolute/URL/parent-traversal collection references before identity resolution, so this does not expand the collection read boundary. A real no-commit merge was clean.

Independent verification on macOS Python 3.9.6:
- focused/neighbor portal suites: 14/14 PASS;
- `py_compile` for the changed script/test: PASS;
- `git diff --check`: PASS.

The required check is green but follows the non-submission short-circuit, so the manual focused tests above are the code-test evidence. Nested or repository-external custom collection roots remain a separate pre-existing extension case, not a blocker for the documented root-level `collections/` layout.

PR #2116, opened 2026-08-12
fix: align public source schema path safety

jiangmuran:
## Summary
- reject traversal, absolute, and protocol-style local paths in the public source schema
- align schema acceptance with the existing dependency-free runtime path normalizer
- add a focused parity matrix covering safe relative paths and unsafe slash/backslash variants

## Validation
- `python3 -m unittest tests.test_source_schema_contract tests.test_public_sources`
- validated the current `sources/public-sources.json` with `Draft202012Validator`
- `Draft202012Validator.check_schema(schema)`
- `git diff --check`

PR #2215, opened 2026-08-12
fix: contain resolved public source paths

jiangmuran:
## Problem

The public-source index validator rejects absolute paths and lexical `..`, but checks `(repo_root / path).is_file()` without verifying the resolved target remains inside the repository. A repository-local symlink can therefore make an indexed `path` point to an arbitrary file outside the public repository while validation reports `ok: true`.

Minimal reproduction on current `main`:

```bash
ln -s /outside/location local
# sources/public-sources.json contains path: local/evidence.txt
python3 scripts/validate_sources.py --json
# reports PASS when /outside/location/evidence.txt exists
```

## Change

Resolve each normalized local source path, require the result to remain under the resolved repository root, then perform the existing regular-file check. Ordinary repository files and links whose targets remain inside the repository continue to validate.

## Verification

- `python3 -m unittest tests.test_public_sources -v` - 6 passed
- regression covers a directory symlink resolving to an existing file outside the temporary repository
- `python3 -m py_compile scripts/validate_sources.py tests/test_public_sources.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_sources.py
+++ b/scripts/validate_sources.py
@@ -153,8 +153,25 @@ def validate_source_item(
             except ValueError as exc:
                 report.add_error(f"{prefix}: {exc}")
             else:
-                if not (repo_root / local_path).is_file():
-                    report.add_error(f"{prefix}: referenced path is missing: {local_path}")
+                try:
+                    resolved = (repo_root / local_path).resolve(strict=True)
+                except FileNotFoundError:
+                    report.add_error(
+                        f"{prefix}: referenced path is missing: {local_path}"
+                    )
+                except (OSError, RuntimeError):
+                    report.add_error(
+                        f"{prefix}: referenced path could not be resolved: {local_path}"
+                    )
+                else:
+                    if not resolved.is_relative_to(repo_root):
+                        report.add_error(
+                            f"{prefix}: referenced path escapes repo: {local_path}"
+                        )
+                    elif not resolved.is_file():
+                        report.add_error(
+                            f"{prefix}: referenced path is missing: {local_path}"
+                        )
 
     if url_value is not None:
         if not isinstance(url_value, str) or not url_value.startswith("https://"):
--- a/tests/test_public_sources.py
+++ b/tests/test_public_sources.py
@@ -136,6 +136,44 @@ class PublicSourcesTests(unittest.TestCase):
             report = validate_source_index(root)
             self.assertTrue(report.ok, report.errors)
 
+    def test_symlinked_local_path_cannot_escape_repo(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            workspace = Path(tmp)
+            root = workspace / "repo"
+            outside = workspace / "outside"
+            outside.mkdir()
+            (outside / "public-brief.md").write_text(
+                "outside evidence", encoding="utf-8"
+            )
+            root.mkdir()
+            try:
+                (root / "brief").symlink_to(outside, target_is_directory=True)
+            except OSError as exc:
+                self.skipTest(f"symlinks unavailable: {exc}")
+            self.write_json(root, "sources/public-sources.json", VALID_INDEX)
+
+            report = validate_source_index(root)
+
+            self.assertFalse(report.ok)
+            self.assertIn("referenced path escapes repo", "\n".join(report.errors))
+
+    def test_symlink_loop_returns_validation_error(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            try:
+                (root / "loop-a").symlink_to("loop-b", target_is_directory=True)
+                (root / "loop-b").symlink_to("loop-a", target_is_directory=True)
+            except OSError as exc:
+                self.skipTest(f"symlinks unavailable: {exc}")
+            bad_index = copy.deepcopy(VALID_INDEX)
+            bad_index["sources"][0]["path"] = "loop-a/evidence.md"
+            self.write_json(root, "sources/public-sources.json", bad_index)
+
+            report = validate_source_index(root)
+
+            self.assertFalse(report.ok)
+            self.assertIn("referenced path could not be resolved", "\n".join(report.errors))
+
 
 if __name__ == "__main__":
     unittest.main()


anselasimov-web (commented):
# PR #2215 review draft

- 仓库 / PR：open-city-ai/haidian#2215「fix: contain resolved public source paths」
- 作者：`jiangmuran`
- checked exact head（全长）：`7311be6a0abd43e1e36904888c9c401ccb8ea444`
- merge-base（全长）：`3a3884766a28f67de49460f0709302b03d3e8bbd`
- 判级：**COMMENT**

---

## 拟发表正文（English）

Hi @jiangmuran — I am a **fellow participant, not a maintainer**, with no review or merge authority. Everything below is measured locally and offered for your consideration.

**Exact head reviewed: `7311be6a0abd43e1e36904888c9c401ccb8ea444`** (merge-base `3a3884766a28f67de49460f0709302b03d3e8bbd`, baseline main `f992bdb02ebd4e0dd0009d82ee7847ba4a99eec4`).

The vulnerability is real and I reproduced it exactly as described. I have two findings on the change itself that I think are worth acting on before this lands: **the new regression test does not actually exercise the vulnerability**, and **the fix introduces a crash path that `main` does not have**.

### 1. Reproduction on `main` — confirmed

```
repo/
  local -> /…/outside            (repository-local directory symlink)
  sources/public-sources.json    (path: "local/evidence.txt")
outside/
  evidence.txt
```

```
main  →  {"ok": true,  "errors": [], "source_count": 1}   rc=0
head  →  {"ok": false, "errors": ["sources[0]: referenced path escapes repo: local/evidence.txt"]}   rc=1
```

So `main` really does report `PASS` for an index entry whose content lives outside the public repository. Containment is the right property to add.

I also checked the fix does not over-reach: an in-repo directory symlink and an in-repo relative file symlink, both resolving inside the repository, still validate `ok: true` on the head commit, and the real `sources/public-sources.json` (2 entries under `brief/`) passes identically on both sides.

### 2. The new test does not reproduce the vulnerability

This is the one I would most like you to look at. In `test_symlinked_local_path_cannot_escape_repo`, the fixture writes:

```python
(outside / "private.md").write_text("private", encoding="utf-8")
```

but `VALID_INDEX` references `brief/public-brief.md`. So `repo/brief -> outside`, and the resolved target is `outside/public-brief.md` — **which the fixture never creates**. Measured, with the shipped fixture exactly as written:

| scenario | `main` | head |
|---|---|---|
| `outside/public-brief.md` **absent** (shipped fixture) | `ok=False`, `"referenced path is missing: brief/public-brief.md"` | `ok=False`, `"referenced path escapes repo: …"` |
| `outside/public-brief.md` **present** (the actual vuln) | **`ok=True`, `errors=[]`** | `ok=False`, `"referenced path escapes repo: …"` |

On `main`, the shipped fixture is already rejected — for the wrong reason. `report.ok` is `False` on both sides, so `self.assertFalse(report.ok)` proves nothing there, and the test is discriminating purely on the error **string**. The security-relevant state — a validator returning `ok: true` for an escaping path — is only reachable in the second row, which the test never sets up. The description says the regression "covers a directory symlink resolving to **an existing file** outside the temporary repository", and that is the row that is missing.

Practical consequence: the test would still catch a straight revert (the message differs), so it is not useless — but it would also pass against a change that merely renamed the missing-path error, and it does not document the property you are actually protecting. One line fixes it:

```python
(outside / "public-brief.md").write_text("outside evidence", encoding="utf-8")
```

I verified that with that line added, the test fails on `main` for the right reason (`ok=True`) and passes on your head.

### 3. New crash path: symlink loop raises out of `validate_source_index()`

`Path.resolve()` is not total. With a repository containing `loopa -> loopb -> loopa` and an index entry `loopa/evidence.txt`:

```
main  →  {"ok": false, "errors": ["sources[0]: referenced path is missing: loopa/evidence.txt"]}   rc=1
head  →  Traceback (most recent call last):
           File ".../scripts/validate_sources.py", line 119, in validate_source_item
             resolved = (repo_root / local_path).resolve()
           File "/usr/lib/python3.10/pathlib.py", line 1074, in check_eloop
             raise RuntimeError("Symlink loop from '…/loopa/evidence.txt'")
```

`main`'s `is_file()` swallows `ELOOP` and returns `False`, so it degrades into a normal report line. The head commit propagates a `RuntimeError` through `validate_source_item` → `validate_source_index` → `main()`, none of which catch it. The exit code is still non-zero, so a pure pass/fail gate survives, but `--json` emits **no JSON at all**, and any caller using `validate_source_index()` as a library (as `tests/test_public_sources.py` does) gets an exception instead of a report.

This matters here specifically because the threat model of this PR is a repository that contains attacker-chosen symlinks — a loop is in exactly the same class as the escape you are fixing. Wrapping the resolve, or checking `is_symlink()` first, keeps the failure inside the report:

```python
try:
    resolved = (repo_root / local_path).resolve()
except (OSError, RuntimeError):
    report.add_error(f"{prefix}: referenced path could not be resolved: {local_path}")
else:
    ...
```

### 4. Two smaller notes

**Prior art in the repository.** `scripts/refresh_submission_manifest.py:52-53` already does the same containment with the shorter spelling:

```python
target = (root / Path(*pure.parts)).resolve()
if not target.is_relative_to(root) or not target.is_file():
```

`is_relative_to()` (3.9+) is already used in `validate_manifest_schema.py`, `score_submission.py` and `self_check_submission.py`, so the `try/except ValueError` block could collapse to one condition and match the house style. (Same file also shares the `ELOOP` exposure from §3, if you want to sweep both.)

**Different policy in the intake path.** For untrusted submission packages, `scripts/validate_submission.py` does not contain symlinks — it rejects them outright (`first_symbolic_link()` / `submission_directory_is_safe()`, "symbolic links are not allowed in submission packages"). Containment is the right call for `sources/`, which is maintainer-curated, but it is worth knowing that the two validators now answer the same question differently on purpose.

**Blast radius.** For calibration: `scripts/validate_sources.py` has no caller anywhere in the repository — no workflow, no other script. Only `docs/public-sources.md`, `README.md:280` and `tests/test_public_sources.py` reference it. So this is hardening a documented manual command, not a CI gate.

**Related open PR.** #2116 ("align public source schema path safety") tightens the `path` pattern in `schema/source.schema.json`. Different file, no merge conflict — the two are complementary (schema-level shape check vs. runtime containment), and neither subsumes the other, since a schema pattern cannot see symlinks.

### 5. Regression baseline: clean

| | tests | failures/errors |
|---|---|---|
| `main` f992bdb0 | 404 | 9 |
| head 7311be6a | 405 | 9 |

Identical failure sets line for line (the 9 are the `submissions/`-dependent gallery/prelaunch tests, failing on both sides in my partial checkout). The 6 tests in `tests/test_public_sources.py` pass on the head commit; the new one fails on `main`.

### 6. Heads-up: CI will not run this test

`submission-validation.yml` is `pull_request_target` and checks out `ref: ${{ github.event.repository.default_branch }}` — trusted `main`, never the PR head — and runs only `scripts/github_pr_validation.py`. The only workflow calling `python3 -m unittest` is `gallery-snapshot-maintenance.yml`, which triggers on `push` to `main` / `workflow_dispatch` and runs exactly three fixed modules. `tests/test_public_sources.py` is not among them, so this regression will never be executed by CI, before or after merge. That is the main reason §2 seemed worth writing up at length: manual execution is the only thing standing behind it.

---

Thanks for finding this one — the symlink-escape reasoning is exactly right, and the reproduction sketch in the description mapped cleanly onto a working repro. If §2 and §3 are already on your list, please ignore.

---

## 复核方法备注（内部，不发布）

1. 工作目录 `$HOME/z4_18714`；main 快照对 `git ls-tree -r main` 逐 blob 核对 235/235 一致、0 missing、0 mismatch（submissions/ 30013 条 sparse 跳过）。
2. PR 快照由已核对的 main 快照复制 + `git cat-file blob pr2215:<path>` 覆盖，写盘后二次 `hash-object`：`ed1dbb4c…` / `655e38ce…` OK。
3. 复现用真实 CLI：`python3 scripts/validate_sources.py --repo-root <tmp> --json`，symlink 用 `ln -s "$PWD/outside" repo/local`。
4. §2 的两行表格用一个探针脚本产生：直接 `import validate_sources.validate_source_index`，从 head 版 `tests/test_public_sources.py` 里 `ast.literal_eval` 取出真实的 `VALID_INDEX`（path=`brief/public-brief.md`），分别在 outside 文件存在/不存在两种 fixture 下跑 main 与 head 四种组合。
5. §3 的 loop：`ln -s loopb loopa; ln -s loopa loopb`，Python 3.10.12，`pathlib.py:1074 check_eloop`。
6. 误伤检查：repo 内 `inside_link -> real`（目录）与 `sources/rel_link.txt -> ../real/evidence.txt`（相对文件）各一条 + 一条普通文件，两侧均 ok:true。
7. 全量基线：39 个 `tests.test_*` 模块，两侧 FAIL/ERROR 排序后 `diff` 为空。
8. 未发布任何 GitHub 内容；gh 仅用于 `api GET` / `pr view` / `pr list`。

jiangmuran:
Rebased onto current `main@f1a58798c` as exact head `06534e6e7ef549df86d6a53fb46a7cffc0e17ac3` and addressed both review findings.

- The escape regression now creates the exact referenced `outside/public-brief.md`, so current main passes the vulnerable fixture and this head rejects it for the containment reason.
- Resolution uses `resolve(strict=True)`: ordinary absent targets retain the existing missing-path message, while `OSError`/`RuntimeError` (including a symlink loop) become a structured `referenced path could not be resolved` error. The loop CLI returns rc=1 with parseable JSON and empty stderr instead of raising.
- A repository-internal symlink resolving to an in-repo file remains valid.

Verification:
- `python3 -m unittest tests.test_public_sources -v` (11 passed)
- `python3 scripts/validate_sources.py --json` (current index PASS, 2 sources)
- dedicated loop CLI JSON and in-repo symlink probes
- `python3 -m py_compile scripts/validate_sources.py tests/test_public_sources.py`
- `git diff --check upstream/main...HEAD`

anselasimov-web (commented):
- Repo / PR: open-city-ai/haidian#2215 "fix: contain resolved public source paths"
- Author: `jiangmuran`
- checked exact head (full 40): `06534e6e7ef549df86d6a53fb46a7cffc0e17ac3`
- my previous review was on head (full 40): `7311be6a0abd43e1e36904888c9c401ccb8ea444` (COMMENTED, 2026-08-12T19:09:43Z)
- baseline main used for every measurement below (full 40): `233853bfdfdd7309f1f6e8de616a94c052f89235`
  (re-checked at 2026-08-16T11:22Z: main had advanced to `9e63df8c5865601d429cbd9d5e79b18bce7c29c6`,
   but `scripts/validate_sources.py` and `tests/test_public_sources.py` are byte-identical between the two,
   so all numbers below still apply)
- PR state at check time: open, `mergeable=true`, `submission-validation` = success (started 2026-08-16T10:32:39Z)
- Env: Python 3.10.12, GNU/Linux. Every `sha256` below is the SHA-256 hex digest.
- suggested_state: **COMMENT**

---

## Body to post (English)

**Checked exact head: `06534e6e7ef549df86d6a53fb46a7cffc0e17ac3`**, against main `233853bfdfdd7309f1f6e8de616a94c052f89235`. My previous comment was on `7311be6a0abd43e1e36904888c9c401ccb8ea444`.

Short version: **all three of my previous points are resolved**, and I could not reproduce any of them on this head. One measured detail and one scope note follow.

### 1. Previous §2 — the regression test now actually reaches the vulnerable state. Confirmed.

The fixture now creates the file the index points at:

```python
(outside / "public-brief.md").write_text("outside evidence", encoding="utf-8")
```

I ran your `tests/test_public_sources.py` from this head against **main's** `scripts/validate_sources.py`:

```
FAIL: test_symlinked_local_path_cannot_escape_repo
  AssertionError: True is not false
```

That is the right failure: on main the validator returns `ok=True` for an index entry whose resolved target is outside the repository. Previously the shipped fixture failed on main for the wrong reason ("referenced path is missing"), so `assertFalse(report.ok)` proved nothing. It now discriminates on the security-relevant state rather than on an error string.

### 2. Previous §3 — the symlink-loop crash path is gone. Confirmed.

`resolve()` is now wrapped, and `FileNotFoundError` is separated from the rest:

```
repo with loop-a -> loop-b -> loop-a, index entry "loop-a/evidence.txt"

main   -> rc=1  {"ok": false, "errors": ["sources[0]: referenced path is missing: loop-a/evidence.txt"]}
head   -> rc=1  {"ok": false, "errors": ["sources[0]: referenced path could not be resolved: loop-a/evidence.txt"]}
```

No traceback, valid JSON on `--json`, and `validate_source_index()` still returns a report to library callers. The new `test_symlink_loop_returns_validation_error` fails on main for the right reason (message differs, no exception), so it is a genuine regression test.

### 3. Previous §4 — house style adopted.

The `try/except ValueError` around `relative_to()` collapsed into `resolved.is_relative_to(repo_root)`, matching `refresh_submission_manifest.py:53`, `validate_manifest_schema.py`, `finalize_submission.py` and 7 other call sites on main. I checked the one thing that could have gone wrong here: `validate_source_index()` does `repo_root = repo_root.resolve()` on entry, so `resolved` and `repo_root` are both fully resolved and the comparison is sound even when `--repo-root` is relative or reached through a symlink. Measured:

```
repo root passed as a symlink to the real directory
  main -> rc=0 ok:true      head -> rc=0 ok:true    (no false positive)
```

### 4. Full behaviour matrix I ran (real CLI, `--json`)

| scenario | main `233853bfdfdd7309f1f6e8de616a94c052f89235` | head `06534e6e7ef549df86d6a53fb46a7cffc0e17ac3` |
|---|---|---|
| escaping directory symlink, target file **present** | rc=0 `ok:true` **(the bug)** | rc=1 `referenced path escapes repo` |
| escaping directory symlink, target file **absent** | rc=1 `referenced path is missing` | rc=1 `referenced path is missing` |
| symlink loop `loop-a -> loop-b -> loop-a` | rc=1 `referenced path is missing` | rc=1 `referenced path could not be resolved` |
| in-repo dir symlink + in-repo relative file symlink + plain file | rc=0 `ok:true` | rc=0 `ok:true` |
| repo root itself reached via symlink | rc=0 `ok:true` | rc=0 `ok:true` |
| real `sources/public-sources.json` (2 entries under `brief/`) | rc=0 `ok:true` | rc=0 `ok:true` |

No over-reach that I could find.

### 5. Regression baseline: clean

43 `tests.test_*` modules, same tree, same interpreter, only the two files swapped:

| | tests | failures | errors |
|---|---|---|---|
| main `233853bfdfdd7309f1f6e8de616a94c052f89235` | 481 | 19 | 170 |
| head `06534e6e7ef549df86d6a53fb46a7cffc0e17ac3` | 483 | 19 | 170 |

The failure/error sets are line-for-line identical (`diff` empty); they are all artefacts of my partial checkout (no `submissions/` corpus), present on both sides. The delta is exactly your two new tests. Within `tests/test_public_sources.py` alone: main 9 tests OK, head 11 tests OK.

### 6. Scope note: main absorbed part of this PR already

Since my last comment, main picked up the module docstring, the `argparse` help strings, the `copy.deepcopy` refactor and four of the extra cases (`test_invalid_type_fails`, `test_invalid_public_status_fails`, `test_source_without_url_or_path_fails`, `test_https_url_passes_without_local_path`). Diffed against **current** main, this PR is now exactly:

- `scripts/validate_sources.py`: the 17-line resolve/containment block
- `tests/test_public_sources.py`: the two symlink tests

which is a noticeably tighter diff than what GitHub shows against the recorded base. Nothing conflicts — I applied both files onto current main and everything above ran from that tree.

### 7. One small residual (message precision only, not a hole)

Because `resolve(strict=True)` raises before containment is evaluated, an escaping symlink whose target does **not** exist is reported as `referenced path is missing` rather than `referenced path escapes repo` (row 2 of the matrix). Both are `ok:false`, rc=1, so there is no gap in the property you are protecting — only the diagnostic is less specific in that one case. Mentioning it in case you want the message to name the actual reason; ignoring it is entirely reasonable.

### 8. Fact on record: CI still does not execute this test

Re-verified on current main. `submission-validation.yml` is `pull_request_target`, checks out `ref: ${{ github.event.repository.default_branch }}` and runs only `scripts/github_pr_validation.py`. The only workflow invoking `python3 -m unittest` is `gallery-snapshot-maintenance.yml` (on `push` to `main` / `workflow_dispatch`), and its module list is fixed:

```
tests.test_submissions_gallery
tests.test_prelaunch_check
tests.test_gallery_snapshot_maintenance_workflow
```

`tests/test_public_sources.py` is not among them, and no workflow references `validate_sources` at all. The green check on this head does not cover these tests.

### Reproduction

```bash
git fetch origin pull/2215/head:pr2215
git checkout pr2215

# 1. the vulnerability, on main's validator
mkdir -p /tmp/esc/outside /tmp/esc/repo/sources
echo "secret" > /tmp/esc/outside/evidence.txt
ln -s /tmp/esc/outside /tmp/esc/repo/local
cat > /tmp/esc/repo/sources/public-sources.json <<'EOF'
{"version":1,"sources":[{"id":"esc-001","title":"t","type":"brief","path":"local/evidence.txt",
"publisher":"p","published_at":"2026-05-15","public_status":"public-draft","citation":"c",
"usage_note":"u","risk_note":"r"}]}
EOF
python3 scripts/validate_sources.py --repo-root /tmp/esc/repo --json ; echo rc=$?

# 2. symlink loop
mkdir -p /tmp/loop/sources
ln -s loop-b /tmp/loop/loop-a ; ln -s loop-a /tmp/loop/loop-b
sed 's/esc-001/loop-001/; s#local/evidence.txt#loop-a/evidence.txt#' \
  /tmp/esc/repo/sources/public-sources.json > /tmp/loop/sources/public-sources.json
python3 scripts/validate_sources.py --repo-root /tmp/loop --json ; echo rc=$?

# 3. do the new tests fail on main? (swap in main's validator, keep the PR tests)
git show main:scripts/validate_sources.py > scripts/validate_sources.py
python3 -m unittest tests.test_public_sources
```

---

Thanks for taking all three points — the one-line fixture fix, the `except (OSError, RuntimeError)` branch and the `is_relative_to()` cleanup are exactly what I had measured, and the two new tests both fail on main for the correct reason, which is the part that actually matters. I have no comment on design trade-offs here. If §7 is already on your list, please ignore this.

PR #2216, opened 2026-08-12
fix: protect exhibit renderer inputs

jiangmuran:
## Problem

`render_exhibit.py` accepts three paths but does not prevent the output from being the same file as `proposal` or `exhibit`. Because rendering completes before the final write, the command exits successfully and then replaces its Markdown or JSON source with generated HTML.

```bash
python3 scripts/render_exhibit.py \
  examples/agent-civic-loop/proposal.md \
  examples/agent-civic-loop/exhibit.json \
  examples/agent-civic-loop/proposal.md
# exits 0 and proposal.md now starts with <!doctype html>
```

## Change

Resolve the output and both input paths before rendering, reject either identity overlap, and leave the source bytes untouched. Normal output paths retain the existing rendering behavior.

## Verification

- `python3 -m unittest tests.test_render_exhibit_output_boundary -v` - passed with proposal and exhibit overlap subtests
- normal smoke render of `examples/agent-civic-loop` produced a non-empty external HTML file
- `python3 -m py_compile scripts/render_exhibit.py tests/test_render_exhibit_output_boundary.py`
- `git diff --check`

### DIFF
--- a/scripts/render_exhibit.py
+++ b/scripts/render_exhibit.py
@@ -718,6 +718,10 @@ def main() -> int:
     args = parser.parse_args()
 
     try:
+        output = args.output.resolve()
+        for label, input_path in (("proposal", args.proposal), ("exhibit", args.exhibit)):
+            if output == input_path.resolve():
+                raise ExhibitError(f"output must not overwrite the {label} input: {input_path}")
         rendered = render_html(args.proposal, args.exhibit)
     except ExhibitError as exc:
         parser.error(str(exc))
--- a/tests/test_render_exhibit_output_boundary.py
+++ b/tests/test_render_exhibit_output_boundary.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SCRIPT = REPO_ROOT / "scripts" / "render_exhibit.py"
+
+
+class RenderExhibitOutputBoundaryTests(unittest.TestCase):
+    def test_output_cannot_overwrite_an_input(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            proposal = root / "proposal.md"
+            exhibit = root / "exhibit.json"
+            proposal.write_text("---\ntitle: Test\n---\n\n# Test\n", encoding="utf-8")
+            exhibit.write_text(json.dumps({"version": 1, "modules": []}), encoding="utf-8")
+
+            for label, output in (("proposal", proposal), ("exhibit", exhibit)):
+                with self.subTest(label=label):
+                    original = output.read_bytes()
+                    completed = subprocess.run(
+                        [sys.executable, str(SCRIPT), str(proposal), str(exhibit), str(output)],
+                        capture_output=True,
+                        text=True,
+                        check=False,
+                    )
+
+                    self.assertNotEqual(completed.returncode, 0)
+                    self.assertIn(f"output must not overwrite the {label} input", completed.stderr)
+                    self.assertEqual(output.read_bytes(), original)
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2216 review draft

- 仓库 / PR：open-city-ai/haidian#2216「fix: protect exhibit renderer inputs」
- 作者：`jiangmuran`
- checked exact head（全长）：`8a3cdab57d644d6ddd23be198e534d29ac284516`
- merge-base（全长）：`e2d2a9bac5d053ee760d6b8fa811049321d9bb63`
- 判级：**COMMENT**

---

## 拟发表正文（English）

**Exact head reviewed: `8a3cdab57d644d6ddd23be198e534d29ac284516`** (merge-base `e2d2a9bac5d053ee760d6b8fa811049321d9bb63`, baseline main `f992bdb02ebd4e0dd0009d82ee7847ba4a99eec4`).

This is the nastiest of the failure modes in this batch — it exits `0` while destroying a tracked source file — and the fix works. I have three additions: a **third input in the same script** that is still unprotected, one identity case the guard cannot see, and **the same defect in the sibling renderer**.

### 1. Reproduction on `main` — confirmed, on the real example package

```
main:  rc=0
  examples/agent-civic-loop/proposal.md
    before c7f690f1f411d0ad9151a7a0706e2a4bd8109fe26582184cbf23cc33a1b48877
    after  171050b96883665be37761b57800e7bbe53d1687cc0cf7b6db0a554ed4729166
  (the "after" digest is byte-for-byte the rendered HTML — the source is gone)

head:  rc=2
  render_exhibit.py: error: output must not overwrite the proposal input: …/proposal.md
  digest unchanged
```

Because the guard runs before `render_html()`, there is no partial write and no created output directory. Normal renders are unaffected: `main` and head produce a byte-identical 23310-byte HTML (`sha256 171050b9…`) for the same inputs.

Both directions on the test: the new `tests/test_render_exhibit_output_boundary.py` fails **both** subTests on `main` (`AssertionError: 0 == 0` — `main` really does exit 0) and passes on your head.

Full-suite baseline is clean:

| | tests | failures/errors |
|---|---|---|
| `main` f992bdb0 | 404 | 9 |
| head 8a3cdab5 | 405 | 9 |

Identical failure sets (the 9 are `submissions/`-dependent tests failing on both sides in my partial checkout).

### 2. `render_exhibit.py` has a third input, and it is a tracked repository file

`render_html()` reads more than `proposal` and `exhibit`:

```python
# scripts/render_exhibit.py:364
track_registry = load_track_registry(Path.cwd())   # → <cwd>/tracks.json
```

Measured from the repository root, on **your head commit**:

```
python3 scripts/render_exhibit.py \
  examples/agent-civic-loop/proposal.md \
  examples/agent-civic-loop/exhibit.json \
  tracks.json

rc=0
tracks.json  d6515b0ef561f930d2db9962345d1c6be858129a41c038d3161fe02532dde932
          →  171050b96883665be37761b57800e7bbe53d1687cc0cf7b6db0a554ed4729166
```

`tracks.json` is committed and is the track registry every validator reads, so this is arguably a worse outcome than overwriting one submission's `proposal.md`. The loop in the fix already has the right shape — it just needs the third entry:

```python
for label, input_path in (
    ("proposal", args.proposal),
    ("exhibit", args.exhibit),
    ("track registry", Path.cwd() / "tracks.json"),
):
```

Your own **#2230** ("protect exhibit generator inputs") already does exactly this for `generate_exhibit.py` — `generation_inputs()` enumerates `repo_root / "gallery-publication.json"`, `repo_root / "schema" / "exhibit.schema.json"`, the manifest-declared artifacts and the figure glob. Lifting that shape into this PR would make the two consistent.

### 3. `resolve()` compares paths, not inodes

Two aliasing cases, both measured against your head:

| alias for `proposal.md` | result |
|---|---|
| symlink (`alias.md -> proposal.md`) | **rc=2, source intact** — `resolve()` handles it |
| hard link (`ln proposal.md hard.md`) | **rc=0, source destroyed** |

Hard links are not representable in git, so this is not reachable from a checkout — but the description frames the guard as rejecting "identity overlap", and `Path.resolve()` gives path identity, not file identity. If you want the stronger property, `os.stat()` + `(st_dev, st_ino)` comparison covers both, at the cost of requiring the output to already exist. I would personally just soften the wording to "path identity" rather than add the stat call; noting it so the claim and the code line up.

### 4. Same defect, sibling script: `render_portal.py`

`scripts/render_portal.py` takes `--output` and writes to it unconditionally, with no relationship check against the proposal directories it reads. Measured on your head commit (i.e. with #2216 applied):

```
python3 scripts/render_portal.py \
  --output examples/agent-civic-loop/proposal.md \
  --collections-dir collections \
  examples/agent-civic-loop

rc=0
proposal.md  c7f690f1…  →  73bcaedd…      (now starts with <!doctype html>)
```

This is not hypothetical typo-space: `README.md:335-343` documents the two commands back to back, over the same `examples/agent-civic-loop` paths, with `render_exhibit.py`'s third positional argument sitting in the same column as `render_portal.py`'s `--output` value. `render_portal.py` also reads `tracks.json` via `load_track_registry(Path.cwd())` (line 1419), so it has both exposures. Whether that belongs in this PR or a follow-up is your call — but with #2216 and #2230 landed, `render_portal.py` would be the last renderer in the family without the guard.

### 5. Heads-up: CI will not run this test

`submission-validation.yml` is `pull_request_target` and checks out `ref: ${{ github.event.repository.default_branch }}` — trusted `main`, never the PR head — then runs only `scripts/github_pr_validation.py`. The only workflow calling `python3 -m unittest` is `gallery-snapshot-maintenance.yml`, on `push` to `main` / `workflow_dispatch`, and it runs exactly three fixed modules (`test_submissions_gallery`, `test_prelaunch_check`, `test_gallery_snapshot_maintenance_workflow`). `tests/test_render_exhibit_output_boundary.py` is never executed by CI, before or after merge — which is why I ran it against both trees by hand rather than trusting the check mark.

---

Thanks for this one — a renderer that exits `0` after eating its own input is exactly the kind of thing that only gets found by someone deliberately looking. The `subprocess` + byte-comparison shape of the test is also the right call here; it catches the exit code, the message and the side effect in one assertion block. If §2 through §4 are already queued, please ignore.

---

## 复核方法备注（内部，不发布）

1. 工作目录 `$HOME/z4_18714`；main 快照对 `git ls-tree -r main` 逐 blob 核对 235/235 一致、0 missing、0 mismatch。
2. PR 快照由已核对的 main 快照复制 + `git cat-file blob pr2216:<path>` 覆盖，写盘后二次 `hash-object`：`8ed3aced…` / `9742dd9c…` OK。
3. ①④ 用真实语料 `examples/agent-civic-loop`（每次先复制到临时目录，避免污染快照），前后 `sha256sum` 取全长。
4. ② 把 head 版新测试文件拷到 main 快照副本上跑：两个 subTest 均 `assertNotEqual(completed.returncode, 0)` → `0 == 0`。
5. ② 反向在 head 快照原地跑：OK。
6. ⑤ 在 head 快照的完整副本内、以 repo 根为 cwd 执行，确保 `Path.cwd()/tracks.json` 命中真实文件。
7. ⑥ symlink：`ln -s proposal.md alias.md`；hardlink：`ln proposal.md hard.md`。
8. ⑦ 在 head 快照完整副本内跑 render_portal.py（即"打了 #2216 之后仍然存在"）。
9. 全量基线：39 个 `tests.test_*` 模块，两侧 FAIL/ERROR 排序后 `diff` 为空。
10. 未发布任何 GitHub 内容；gh 仅用于 `api GET` / `pr view` / `pr list`。

PR #2217, opened 2026-08-12
fix: keep unreviewed sources in review queue

jiangmuran:
## Problem

`source_registry_utils.summarize_source_registry()` assigns each source to one display/review group, but checks `usable_for_formal=background_only` before `review_status=needs_review`. A newly generated A0/A1 draft has exactly that combination, so it is exposed as an ordinary background source and disappears from `needs_review_sources` even though the registry usage rule says it must not be used until reviewed.

```python
summarize_source_registry({"sources": [{
    "source_id": "DRAFT-A0",
    "review_status": "needs_review",
    "usable_for_formal": "background_only",
    "authority_level": "A0",
}]})
# background_sources: [DRAFT-A0]
# needs_review_sources: []
```

This summary feeds maintainer review input, scaffold source guidance, and generated frontend registry data.

## Change

Give `review_status=needs_review` precedence over use buckets when building the mutually exclusive source lists. Approved background sources and provisional sources keep their existing groups, and the independent count dimensions remain unchanged.

## Verification

- `python3 -m unittest tests.test_source_registry_summary tests.test_source_registry_frontend tests.test_data_registry -v` - 9 passed
- regression covers both `background_only` and `no` unreviewed sources plus approved background and provisional controls
- `python3 -m py_compile scripts/source_registry_utils.py tests/test_source_registry_summary.py`
- `git diff --check`

### DIFF
--- a/scripts/source_registry_utils.py
+++ b/scripts/source_registry_utils.py
@@ -68,14 +68,14 @@ def summarize_source_registry(registry: dict[str, Any], limit: int = 8) -> dict[
         _increment(counts["by_usable_for_formal"], formal_use)
         _increment(counts["by_authority_level"], authority)
         compact = _compact_source(source)
-        if review_status == "approved" and formal_use == "yes":
+        if review_status == "needs_review":
+            needs_review.append(compact)
+        elif review_status == "approved" and formal_use == "yes":
             approved_formal.append(compact)
         elif formal_use == "provisional_only":
             provisional.append(compact)
         elif formal_use == "background_only":
             background.append(compact)
-        elif review_status == "needs_review":
-            needs_review.append(compact)
     return {
         "registry_path": DEFAULT_SOURCE_REGISTRY_PATH,
         "updated_date": registry.get("updated_date") if isinstance(registry, dict) else None,
--- a/tests/test_source_registry_summary.py
+++ b/tests/test_source_registry_summary.py
@@ -1,0 +1,59 @@
+from __future__ import annotations
+
+import sys
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+from source_registry_utils import summarize_source_registry  # noqa: E402
+
+
+def source(source_id: str, review_status: str, formal_use: str) -> dict[str, object]:
+    return {
+        "source_id": source_id,
+        "title": source_id,
+        "publisher": "Publisher",
+        "authority_level": "A0",
+        "review_status": review_status,
+        "usable_for_formal": formal_use,
+        "topics": ["test"],
+        "allowed_uses": ["test"],
+        "prohibited_uses": ["formal use before review"],
+        "url": "https://example.com/source",
+    }
+
+
+class SourceRegistrySummaryTests(unittest.TestCase):
+    def test_needs_review_takes_precedence_over_use_bucket(self) -> None:
+        registry = {
+            "sources": [
+                source("DRAFT-BACKGROUND", "needs_review", "background_only"),
+                source("DRAFT-NO", "needs_review", "no"),
+                source("APPROVED-BACKGROUND", "approved", "background_only"),
+                source("PROVISIONAL", "provisional", "provisional_only"),
+            ]
+        }
+
+        summary = summarize_source_registry(registry)
+
+        self.assertEqual(
+            [item["source_id"] for item in summary["needs_review_sources"]],
+            ["DRAFT-BACKGROUND", "DRAFT-NO"],
+        )
+        self.assertEqual(
+            [item["source_id"] for item in summary["background_sources"]],
+            ["APPROVED-BACKGROUND"],
+        )
+        self.assertEqual(
+            [item["source_id"] for item in summary["provisional_sources"]],
+            ["PROVISIONAL"],
+        )
+        self.assertEqual(summary["counts"]["by_review_status"]["needs_review"], 2)
+        self.assertEqual(summary["counts"]["by_usable_for_formal"]["background_only"], 2)
+
+
+if __name__ == "__main__":
+    unittest.main()


anselasimov-web (commented):
# PR #2217 review draft

- 仓库 / PR：open-city-ai/haidian#2217「fix: keep unreviewed sources in review queue」
- 作者：`jiangmuran`
- checked exact head（全长）：`cb6b057bb140acbc151d1480de8030b6eed5f1b3`
- merge-base（全长）：`e2d2a9bac5d053ee760d6b8fa811049321d9bb63`
- 判级：**COMMENT**

---

## 拟发表正文（English）

**Exact head reviewed: `cb6b057bb140acbc151d1480de8030b6eed5f1b3`** (merge-base `e2d2a9bac5d053ee760d6b8fa811049321d9bb63`, baseline main `f992bdb02ebd4e0dd0009d82ee7847ba4a99eec4`).

The precedence bug is real, and I was able to confirm your premise directly in the draft generator rather than only in the abstract. Two things to add: **one bucket move the description does not mention and the test does not cover**, and a note on why this fix is currently invisible on the committed registry.

### 1. The premise checks out in `prepare_source_registry_draft.py`

Your description says "a newly generated A0/A1 draft has exactly that combination". That is not just plausible — it is forced:

```python
# scripts/prepare_source_registry_draft.py:188
"review_status": "needs_review",
"usable_for_formal": derive_formal_use(authority, access_status),

# lines 130-137
def derive_formal_use(authority, access_status):
    if access_status == "restricted_or_unknown": return "no"
    if authority in {"A0", "A1"}:                return "background_only"
    if authority == "A3":                        return "background_only"
    return "no"
```

`review_status` is hardcoded, and `derive_formal_use()` returns `background_only` for every accessible A0/A1/A3 source. So on `main`, **every** accessible A0/A1 draft the generator produces lands in `background_sources` and vanishes from `needs_review_sources` — the module docstring of that very file says these drafts "cannot be used as formal evidence until a maintainer reviews", so the summary was contradicting its own producer. Worth noting the same function never returns `provisional_only`, which is relevant to §2.

Differential on a registry seeded with three realistic drafts:

| source | `review_status` / `usable_for_formal` | `main` bucket | head bucket |
|---|---|---|---|
| `DRAFT-A0-BG` | `needs_review` / `background_only` | `background_sources` | `needs_review_sources` |
| `DRAFT-A2-NO` | `needs_review` / `no` | `needs_review_sources` | `needs_review_sources` |
| `HAND-PROV` | `needs_review` / `provisional_only` | `provisional_sources` | `needs_review_sources` |

Row 1 is your fix. Row 3 is §2.

Both count dimensions are byte-identical across the change, as you claim — `by_review_status` and `by_usable_for_formal` both read `{approved: 8, provisional: 1, needs_review: 3}` / `{yes: 7, provisional_only: 2, background_only: 2, no: 1}` on either side.

Both directions on the test: the new test fails on `main` (`needs_review_sources` comes back as `['DRAFT-NO']`, expected `['DRAFT-BACKGROUND', 'DRAFT-NO']`) and passes on your head.

### 2. Unstated bucket move: `needs_review` + `provisional_only`

The description says "Approved background sources and **provisional sources keep their existing groups**". That holds for `provisional`/`approved` review statuses, but not across the board: a source with `review_status: needs_review` **and** `usable_for_formal: provisional_only` moves out of `provisional_sources` into `needs_review_sources` (row 3 above, measured). On `main` it was in `provisional`; after this change it is not.

I think the new behaviour is **correct** — the module's own `usage_rule` string says "needs_review sources must not be used until reviewed", which is a stronger constraint than "provisional_only sources for intake/visualization only", so precedence should win there too. But it is a real change to a bucket that feeds `generate_source_registry_data.py`'s frontend `provisional` list and `scaffold_ai_submission.py`'s guidance, and the new test does not exercise it — its `PROVISIONAL` fixture uses `review_status: "provisional"`, which is the case that does *not* move. Two small suggestions:

- add a fifth fixture `source("DRAFT-PROVISIONAL", "needs_review", "provisional_only")` and assert it lands in `needs_review_sources`, so the precedence rule is pinned for all three use buckets rather than two;
- reword the description line to something like "approved and provisional-status sources keep their existing groups", so a reader diffing buckets is not surprised.

As §1 notes, `derive_formal_use()` cannot emit this combination today, so it only arises from a hand-edited registry entry — which is exactly the case a maintainer would hand-edit and then not expect to disappear from the provisional list.

### 3. The fix is currently latent on the committed registry

For calibration when reading the green/red of any local run: `data/source_registry.json` currently holds 9 sources — 7 `approved`/`yes`, 1 `approved`/`background_only`, 1 `provisional`/`provisional_only` — and **zero** `needs_review`. So:

```
main  →  source-registry-data.js  sha256 6abfde36…   == committed blob 510b86fefdd3c664941cb68ab3b79b8c7b66bb6e
head  →  source-registry-data.js  sha256 6abfde36…   == committed blob 510b86fefdd3c664941cb68ab3b79b8c7b66bb6e
```

Byte-identical, and both already fresh against the checked-in artifact. That is a good thing (nothing to regenerate, no stale-artifact churn in this PR), but it does mean the end-to-end behaviour change is invisible until the first draft is merged into the registry — so a reviewer looking only at generated output would conclude the PR does nothing. That is worth one sentence in the description.

### 4. Full-suite baseline: zero regression

| | tests | failures/errors |
|---|---|---|
| `main` f992bdb0 | 404 | 9 |
| head cb6b057b | 405 | 9 |

Identical failure sets line for line (the 9 are `submissions/`-dependent gallery/prelaunch tests failing on both sides in my partial checkout). I confirmed the three consumers you listed are the real ones: `review_submission.py:150`, `scaffold_ai_submission.py:958`, and `generate_source_registry_data.py:20-29` which maps the four lists onto the frontend `formal`/`background`/`provisional`/`needsReview` buckets.

One minor thing while you are in this function: all four lists are truncated by `[:limit]` (default 8, 20 from the frontend generator). With this change `needs_review_sources` absorbs entries from two more buckets, so it is the list most likely to hit the cap first once drafts start landing — the counts dimension is unaffected, but a maintainer reading `needs_review_sources` may not realise it is a head slice.

### 5. Heads-up: CI will not run this test

`submission-validation.yml` is `pull_request_target` and checks out `ref: ${{ github.event.repository.default_branch }}` — trusted `main`, never the PR head — then runs only `scripts/github_pr_validation.py`. The only workflow that calls `python3 -m unittest` is `gallery-snapshot-maintenance.yml`, on `push` to `main` / `workflow_dispatch`, running exactly `tests.test_submissions_gallery`, `tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`. `tests/test_source_registry_summary.py` is not among them and never will be, so this regression gets no CI coverage before or after merge — hence running it against both trees by hand.

Also worth knowing: your #2233 (`generate_source_registry_data.py`) and #2211 (`prepare_source_registry_draft.py`) sit on either side of this function in the same pipeline but touch different files, so I see no merge conflict between the three.

---

Thanks — the "one source, one bucket, and the order of the `elif` chain silently encodes a policy" framing is a good catch, and pointing at the exact producer that generates the bad combination is what made it easy to verify rather than take on faith. If §2 is already on your list, please ignore.

---

## 复核方法备注（内部，不发布）

1. 工作目录 `$HOME/z4_18714`；main 快照对 `git ls-tree -r main` 逐 blob 核对 235/235 一致、0 missing、0 mismatch。
2. PR 快照由已核对的 main 快照复制 + `git cat-file blob pr2217:<path>` 覆盖，写盘后二次 `hash-object`：`84f2f2df…` / `c9928334…` OK。
3. ① 的表：探针脚本直接 `import source_registry_utils.summarize_source_registry`，用真实 `data/source_registry.json` 深拷贝后追加三条构造草稿（字段按 `_compact_source` 读取的键补全），main / head 各跑一次，`limit=20`。
4. ④ `python3 scripts/generate_source_registry_data.py` 在两个快照内各跑一次，对产出 `source-registry-data.js` 取 sha256，并用 `git rev-parse main:source-registry-data.js` 与 `git hash-object` 比对已提交 blob。
5. 消费方用 `grep -rn "summarize_source_registry\|needs_review_sources\|background_sources\|provisional_sources\|approved_formal_sources"` 全仓确认。
6. 全量基线：39 个 `tests.test_*` 模块，两侧 FAIL/ERROR 排序后 `diff` 为空。
7. 未发布任何 GitHub 内容；gh 仅用于 `api GET` / `pr view` / `pr list`。

CocoSgt (approved):
Reviewed exact head `cb6b057bb140acbc151d1480de8030b6eed5f1b3` against current `origin/main@4f965b4784a30557b4603bc7ce0abc99d4181425`. The precedence defect is real: `needs_review` must remain in the maintainer queue regardless of a background/provisional use annotation. The change is fail-closed, does not modify registry data or policy gates, and independent count dimensions remain intact.\n\nIndependent verification: clean merge-tree; `tests.test_source_registry_summary`, `tests.test_source_registry_frontend`, and `tests.test_data_registry` 9/9 PASS; Python compile and `git diff --check` PASS. An additional `needs_review + provisional_only` probe lands only in `needs_review_sources` while preserving both count dimensions, matching the existing usage rule. No blocking correctness, security, or compatibility issue found.

PR #2218, opened 2026-08-12
fix: replace stale submission skill installs

jiangmuran:
## Problem

`install_submission_skill.py` overlays the repository skill onto an existing target with `dirs_exist_ok=True`. Files removed or renamed in the source therefore remain installed indefinitely. The resulting tree has a different digest, but both install mode and `--check` still return exit code 0 because `ok` only checks for the two `SKILL.md` files.

```text
install fresh skill
add skills/urban-design-ai-submission/removed-reference.md
--check => exit 0, ok=true, up_to_date=false
reinstall => exit 0, ok=true, up_to_date=false, stale file remains
```

That makes `--check` unsuitable for automation and can leave agents following obsolete references even after reinstalling.

## Change

- copy the source skill into a clean sibling staging directory;
- replace the installed skill directory with that exact copy so deleted source files disappear;
- make report `ok` require matching source and target tree digests;
- cover stale detection and cleanup through the public CLI.

The replacement remains scoped to this named skill directory; unrelated skills and Codex home contents are untouched.

## Verification

- `python3 -m unittest tests.test_site_package_contract.SitePackageContractTests.test_submission_skill_installer_copies_skill tests.test_site_package_contract.SitePackageContractTests.test_submission_skill_installer_replaces_stale_installation` (2 passed)
- `python3 -m py_compile scripts/install_submission_skill.py tests/test_site_package_contract.py`
- `git diff --check`

The complete `tests.test_site_package_contract` module ran 11 tests: the two installer tests passed, one optional dependency test skipped, and the current default branch's unrelated `test_skill_exists` assertion failed because it expects `machine-readable urban design submission` while `SKILL.md` currently says `machine-readable urban design package`.

### DIFF
--- a/scripts/install_submission_skill.py
+++ b/scripts/install_submission_skill.py
@@ -46,19 +46,23 @@ def tree_digest(path: Path) -> str:
 
 def copy_skill(source: Path, target: Path) -> None:
     target.parent.mkdir(parents=True, exist_ok=True)
+    temporary_target = target.parent / f".{target.name}.installing"
+    if temporary_target.exists():
+        shutil.rmtree(temporary_target)
     shutil.copytree(
         source,
-        target,
-        dirs_exist_ok=True,
+        temporary_target,
         ignore=shutil.ignore_patterns("__pycache__", "*.pyc", ".DS_Store"),
     )
+    if target.exists():
+        shutil.rmtree(target)
+    temporary_target.replace(target)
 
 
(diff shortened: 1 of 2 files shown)


jiangmuran:
Closing as a duplicate of #1054. A local ledger review after publishing found that #1054 already covers stale skill detection and cleanup; no maintainer action is needed here.

PR #808, opened 2026-08-09
ci: run Python tests on code pull requests

147228:
## What changed

Adds a dedicated `python-tests` workflow for pull requests that change Python code, tests, review dependencies, or Actions workflows.

## Why

The existing `submission-validation` workflow intentionally checks out trusted `main` and validates PR metadata. It does not execute PR Python code, so code changes can receive that structural green check without remote unit-test evidence.

## Security boundary

The new workflow uses the `pull_request` event, explicitly grants only `contents: read`, and has no secret-dependent steps. It therefore executes fork PR code without the write-capable `pull_request_target` context used for trusted metadata handling.

## Verification

- `python3 -m unittest discover -s tests` — 269 passed
- Workflow YAML parsed successfully
- `git diff --check` passed

The first remote run will provide the GitHub Actions environment verification.

PR #773, opened 2026-08-08
fix: keep bilingual backfill manifests current

147228:
## What changed

- keep `manifest.json` synchronized when `backfill_proposal()` creates or refreshes `proposal.<lang>.md`
- register the localized proposal with its language and `translation_of` metadata
- refresh SHA-256 values for both proposal files
- invalidate `validation_claim.self_checked` after the content change
- write the manifest atomically while preserving its existing file mode
- add a regression test covering the generated translation, hashes, metadata, and self-check invalidation

## Why

The merged bilingual backfill path rewrote `proposal.md` and created a localized sibling, but did not update the package manifest. On current `main` (`2c6ae754`), a read-only audit finds 133 of 262 packages with a stale `proposal.md` hash, and some translated proposals are not registered in their manifest. This can make deterministic validation reject otherwise generated packages; preserving an old `self_checked: true` would also overstate freshness.

This patch fixes the write path for future backfills. The already-stale historical packages still need a maintainer-controlled batch refresh followed by the full self-check gates; this PR does not silently rewrite those submission artifacts.

## Validation

- `python3 -m pytest -q` (268 passed)
- `python3 -m pytest -q tests/test_bilingual_backfill.py` (23 passed)
- `python3 scripts/generate_submissions_data.py --check`
- `node --check submissions-data.js`
- `git diff --check`

PR #2221, opened 2026-08-12
fix: invalidate reviews after artifact backfill

jiangmuran:
## Problem

`backfill_bilingual_artifacts.py manifests` registers newly generated HTML, PDF, and figure counterparts and refreshes their manifest hashes, but it preserves `validation_claim.self_checked=true` even when those records or bytes changed.

That leaves the package claiming a fresh four-gate review after the display artifacts covered by that review have changed. It also assigns generic roles when a display artifact was absent from the historical manifest: report and visual HTML become `narrative`, and drawings become `figure`, despite the published manifest contract defining specific roles for each.

## Change

- compare the manifest file inventory before and after artifact registration/hash refresh;
- set `validation_claim.self_checked=false` when the inventory actually changes;
- preserve `self_checked` on an idempotent rerun with no manifest changes;
- assign `rendered_proposal_html`, `visualization`, `drawing`, and `proposal_figure` to newly registered display artifacts;
- add dependency-free regression coverage in a separate test module.

This complements #773, which invalidates review freshness when the proposal translation stage changes Markdown. This PR covers the subsequent HTML/PDF/figure artifact stage.

## Verification

- `python3 -m unittest tests.test_bilingual_artifact_manifest_freshness tests.test_bilingual_backfill.BilingualBackfillTests.test_manifest_backfill_registers_language_counterparts_and_hashes` (3 passed)
- `python3 -m py_compile scripts/backfill_bilingual_artifacts.py tests/test_bilingual_artifact_manifest_freshness.py`
- `git diff --check`

The complete bilingual suites ran 24 tests: 23 passed; the existing localized-figure test requires Arial Unicode, which is absent in this Linux environment and is already addressed by open PR #808's Noto CJK fallback.

### DIFF
--- a/scripts/backfill_bilingual_artifacts.py
+++ b/scripts/backfill_bilingual_artifacts.py
@@ -482,6 +482,7 @@ def backfill_manifests(dirs: list[Path]) -> int:
             continue
         manifest = json.loads(manifest_path.read_text(encoding="utf-8"))
         items = manifest.setdefault("files", [])
+        original_items = json.dumps(items, ensure_ascii=False, sort_keys=True)
         by_path = {
             str(item.get("path")): item
             for item in items
@@ -504,9 +505,20 @@ def backfill_manifests(dirs: list[Path]) -> int:
                 continue
             primary_item = by_path.get(primary)
             if primary_item is None:
+                role = (
+                    "rendered_proposal_html"
+                    if primary == "report/proposal.html"
+                    else "visualization"
+                    if primary == "visual/index.html"
+                    else "drawing"
+                    if primary.startswith("drawings/")
+                    else "proposal_figure"
+                    if primary.startswith("assets/figures/")
+                    else "narrative"
+                )
                 primary_item = {
                     "path": primary,
-                    "role": "narrative" if primary.endswith((".md", ".html")) else "figure",
+                    "role": role,
                     "required": primary in {"proposal.md", "report/proposal.html", "visual/index.html"},
                 }
                 items.append(primary_item)
@@ -532,6 +544,10 @@ def backfill_manifests(dirs: list[Path]) -> int:
             rel = item.get("path")
             if rel and rel != "manifest.json" and (directory / rel).is_file():
                 item["sha256"] = sha256(directory / rel)
+        if json.dumps(items, ensure_ascii=False, sort_keys=True) != original_items:
+            validation_claim = manifest.get("validation_claim")
+            if isinstance(validation_claim, dict):
+                validation_claim["self_checked"] = False
         manifest_path.write_text(
             json.dumps(manifest, ensure_ascii=False, indent=2) + "\n", encoding="utf-8"
         )
--- a/tests/test_bilingual_artifact_manifest_freshness.py
+++ b/tests/test_bilingual_artifact_manifest_freshness.py
@@ -1,0 +1,84 @@
+import hashlib
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+from backfill_bilingual_artifacts import backfill_manifests  # noqa: E402
+
+
+class BilingualArtifactManifestFreshnessTests(unittest.TestCase):
+    def write_package(self, root: Path) -> Path:
+        directory = root / "submissions" / "alice" / "example"
+        for relative, content in {
+            "proposal.md": '---\ntitle: "Example"\nlanguage: "en"\n---\nBody\n',
+            "proposal.zh.md": '---\ntitle: "示例"\nlanguage: "zh"\ntranslation_of: "proposal.md"\n---\n正文\n',
+            "report/proposal.html": "primary report",
+            "report/proposal.zh.html": "translated report",
+            "visual/index.html": "primary visual",
+            "visual/index.zh.html": "translated visual",
+            "drawings/a3-booklet.pdf": "primary drawing",
+            "drawings/a3-booklet.zh.pdf": "translated drawing",
+        }.items():
+            path = directory / relative
+            path.parent.mkdir(parents=True, exist_ok=True)
+            path.write_text(content, encoding="utf-8")
+        (directory / "manifest.json").write_text(
+            json.dumps(
+                {
+                    "files": [
+                        {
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):

**Exact head reviewed: `fdcd8270bf215b40a002449d64f279e2101e9312`**, against baseline main `8bbaa84cfc7c683db188736f436e4ab98b76babc`. `scripts/backfill_bilingual_artifacts.py` is byte-identical on your PR's parent commit and on current main (blob `1f558b15e07d63c52b823419dd5c98d44858d199`), so every "before" result below applies to main as it stands.

The defect is real, your reproduction holds in both directions, and there is no regression. I have one measured addition: the change-detection window is wider than what the PR describes, and on the current package corpus that difference is 57 packages.

```
# new test on unpatched main
git checkout 8bbaa84cfc7c683db188736f436e4ab98b76babc -- scripts/backfill_bilingual_artifacts.py
python3 -m unittest tests.test_bilingual_artifact_manifest_freshness
  -> FAILED (failures=1)
     AssertionError: True is not false      # main really does keep self_checked=true

# same test on your head
git checkout fdcd8270bf215b40a002449d64f279e2101e9312 -- scripts/backfill_bilingual_artifacts.py
python3 -m unittest tests.test_bilingual_artifact_manifest_freshness
  -> OK (Ran 2 tests)
```

Full-suite baseline is clean:

|---|---|---|
| head `fdcd8270bf215b40a002449d64f279e2101e9312` | 406 | 25 |

The failure sets are identical (`diff` is empty). All 25 are environment-dependent cases that need `data/`, `assets/` or `collections/`, which my partial checkout omits — none of them touch this script.

Your disclosed Arial Unicode failure in `tests.test_bilingual_backfill` reproduces identically on **unpatched main**, so it is pre-existing and not introduced here.

One mechanical note worth stating explicitly, since it decides how much weight the green check carries: the workflow runs only `python3 scripts/github_pr_validation.py`, and that script short-circuits at `is_non_submission_pr` (`scripts/github_pr_validation.py:611`) for a `scripts/` + `tests/` PR. The bot comment on this PR says so itself — "non-submission code/docs/test PR; participant package validation was not applicable". **The new test module is never executed by CI**, on this PR or after merge, so a green `submission-validation` is not evidence that it passed. Manual runs are the only check it gets.

### 2. The change-detection window also fires on pure metadata annotation

The snapshot compares the whole serialized `files[]` list:

```python
original_items = json.dumps(items, ensure_ascii=False, sort_keys=True)
...
if json.dumps(items, ensure_ascii=False, sort_keys=True) != original_items:
    validation_claim["self_checked"] = False
```

Between those two points the loop also stamps `language` and `translation_of` onto entries that already existed — and main already did that stamping. So an annotation that changes no artifact and no `sha256` now counts as an inventory change.

Minimal end-to-end reproduction on your head. Every declared file is present, every `sha256` is already correct, and no translated counterpart exists on disk, so nothing is registered:

```python
# manifest.json before: files[] = proposal.md, report/proposal.html, visual/index.html
#   each with a correct sha256; validation_claim.self_checked = true
backfill_manifests([Path(".../submissions/alice/demo")])
```

```
BEFORE self_checked = True
AFTER  self_checked = False
delta in files[]: only the language key was added -> [('proposal.md','zh'),
                  ('report/proposal.html','zh'), ('visual/index.html','zh')]
files[] length: 3 -> 3   (nothing registered)
```

No artifact was registered, no artifact byte changed, no `sha256` changed. The four-gate claim is dropped anyway.

### 3. Measured effect on the current corpus

`submission_dirs()` returns every package when `--only` is absent, so `python3 scripts/backfill_bilingual_artifacts.py manifests` is a whole-repository operation. I simulated the `files[]` transition for all packages on main `8bbaa84cfc7c683db188736f436e4ab98b76babc` using the git tree for file existence:

```
packages scanned: 725      currently self_checked=true: 508
self_checked true -> false: 138        (lower bound; sha256 refresh not simulated)
   59  register_counterpart + set_language + set_translation_meta
   57  set_language only                      <-- no file registered, no byte changed
    8  register_primary + set_language
    7  register_counterpart + set_translation_meta
    6  all four
    1  set_language + set_translation_meta
```

So **57 of the 138 flips are caused by the metadata stamping alone**. The number is a lower bound because I could not hash artifact bytes for the whole corpus, so the true total is at least this.

This matters downstream because `scripts/validate_submission.py:1721` treats the flag as mandatory:

```python
if not isinstance(claim, dict) or claim.get("self_checked") is not True:
    ... "must set validation_claim.self_checked=true after running self_check"
```

Each flipped package therefore hard-fails validation until its owner re-runs the self-check — including the 57 whose artifacts were never touched.

### 4. Suggested repair, verified end to end

Compare an artifact inventory rather than the serialized entries, so registration and content changes still trigger while annotation does not:

```python
def artifact_inventory(entries: list) -> list:
    return sorted(
        (str(entry.get("path")), entry.get("sha256"))
        for entry in entries
        if isinstance(entry, dict) and entry.get("path")
    )

original_inventory = artifact_inventory(items)
...
if artifact_inventory(items) != original_inventory:
    validation_claim["self_checked"] = False
```

Measured on your head with only that substitution:

```
your two tests                     -> OK (Ran 2 tests)          # both still pass
tests.test_bilingual_backfill      -> unchanged (only the pre-existing font error)
metadata-only reproduction (§2)    -> self_checked stays True
corpus flips                       -> 138 -> 80
metadata-only flips                -> 57 -> 0
```

The path-plus-`sha256` pair keeps exactly the two signals your PR is about: a newly registered artifact changes the path set, and changed bytes change the `sha256`.

- The four new role tokens (`rendered_proposal_html`, `visualization`, `drawing`, `proposal_figure`) are all present in the role enum of `brief/site-package/schemas/manifest.schema.json`.
- The `else "narrative"` fallback is reachable only for `proposal.md` given the fixed `primary_paths` list, and main already mapped `.md` to `narrative` — so no silent role change hides in that branch.
- Existing manifest entries keep their recorded role, since the counterpart inherits `primary_item.get("role", ...)` from the entry already in the manifest.
- Idempotency holds: a second run with nothing changed preserves the flag, as your second test asserts.
- No permission, grouping, guard-direction, or symlink failure-path change found in this diff.

Thanks for splitting this out from the translation stage — the artifact stage genuinely was uncovered, and the two-directional test you wrote is the right shape. I have no view on the design direction; the only thing I am reporting is the measured width of the trigger. **If you are already on this, please ignore.**

PR #2222, opened 2026-08-12
fix: keep review inputs out of submissions

jiangmuran:
## Summary

- reject lexical or resolved review output paths containing an exact `submissions` component, independent of `--repo-root`
- bind the guard to the selected submission as an additional fail-closed anchor
- publish the three review packet files through same-directory temporary files and `os.replace`
- preserve existing output modes while preventing file-level symlink and hardlink aliases from modifying package inputs

## Reproduction

The previous head could be bypassed by pointing `--repo-root` at `submissions/alice` and writing beneath the selected package. An otherwise external output directory could also contain `review-input.json`, `review-prompt.md`, or `advisory-review.md` as a symlink or hardlink to `proposal.md`; ordinary `write_text()` then replaced the package input bytes.

The updated guard checks both lexical and resolved output paths before creating the directory. Atomic replacement changes only the requested output directory entry, so linked victims retain their original inode and bytes.

## Validation

- `python3 -m unittest tests.test_review_submission_output_boundary tests.test_review_submission -v` (8 passed)
- same 8 tests under Python 3.9
- `python3 -m py_compile scripts/review_submission.py tests/test_review_submission_output_boundary.py`
- Python 3.9 `py_compile`
- `git diff --check upstream/main...HEAD`

The regression matrix covers nested repository roots, direct and dangling submission paths, all three fixed output names through symlink and hardlink aliases, normal maintainer output, and external output.

### DIFF
--- a/scripts/review_submission.py
+++ b/scripts/review_submission.py
@@ -34,8 +34,10 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from typing import Any
 
@@ -161,13 +163,48 @@ def script_path(repo_root: Path, name: str) -> Path:
     return SCRIPT_DIR / name
 
 
-def validate_output_dir(repo_root: Path, out_dir: Path) -> None:
+def validate_output_dir(repo_root: Path, out_dir: Path, submission_dir: Path) -> None:
     try:
-        relative = out_dir.resolve().relative_to(repo_root.resolve())
-    except ValueError:
-        return
-    if relative.parts and relative.parts[0] == "submissions":
-        raise ReviewOutputError("Review output must not be written inside `submissions/`")
+        candidates = {
+            Path(os.path.abspath(out_dir)),
+            out_dir.resolve(),
+        }
+        protected_roots = {
+            repo_root.resolve() / "submissions",
+            submission_dir.resolve(),
+        }
+    except (OSError, RuntimeError) as exc:
+        raise ReviewOutputError(f"Review output path cannot be resolved safely: {exc}") from exc
+
+    for candidate in candidates:
+        if "submissions" in candidate.parts or any(
+            candidate == root or candidate.is_relative_to(root) for root in protected_roots
+        ):
+            raise ReviewOutputError("Review output must not be written inside `submissions/`")
+
+
+def write_text_atomically(path: Path, text: str) -> None:
+    mode = path.stat().st_mode & 0o777 if path.exists() else 0o644
+    temporary: Path | None = None
+    try:
+        with tempfile.NamedTemporaryFile(
+            mode="w",
+            encoding="utf-8",
+            dir=path.parent,
+            prefix=f".{path.name}.",
+            suffix=".tmp",
+            delete=False,
+        ) as handle:
+            temporary = Path(handle.name)
+            os.fchmod(handle.fileno(), mode)
+            handle.write(text)
+            handle.flush()
+            os.fsync(handle.fileno())
+        os.replace(temporary, path)
+    except BaseException:
+        if temporary is not None:
+            temporary.unlink(missing_ok=True)
+        raise
 
 
 def run_json_command(command: list[str]) -> dict:
@@ -413,17 +450,18 @@ def main() -> int:
     else:
         out_dir = repo_root / DEFAULT_OUTPUT_ROOT / submission_dir.name / "review-packet"
     try:
-        validate_output_dir(repo_root, out_dir)
+        validate_output_dir(repo_root, out_dir, submission_dir)
     except ReviewOutputError as exc:
         parser.error(str(exc))
     out_dir.mkdir(parents=True, exist_ok=True)
 
     review_input = build_review_input(repo_root, submission_dir)
-    (out_dir / "review-input.json").write_text(
-        json.dumps(review_input, ensure_ascii=False, indent=2), encoding="utf-8"
+    write_text_atomically(
+        out_dir / "review-input.json",
+        json.dumps(review_input, ensure_ascii=False, indent=2),
     )
-    (out_dir / "review-prompt.md").write_text(build_prompt(review_input), encoding="utf-8")
-    (out_dir / "advisory-review.md").write_text(build_template(review_input), encoding="utf-8")
+    write_text_atomically(out_dir / "review-prompt.md", build_prompt(review_input))
+    write_text_atomically(out_dir / "advisory-review.md", build_template(review_input))
     print(out_dir)
     return 0
 
--- a/scripts/review_submission.py
+++ b/scripts/review_submission.py
@@ -119,6 +119,10 @@ PACKAGE_FILES_INCLUDED_AS_PARTIAL_PREVIEW = {
 }
 
 
+class ReviewOutputError(ValueError):
+    pass
+
+
 def read_text(path: Path) -> str:
     if not path.exists():
         return ""
@@ -155,6 +159,15 @@ def script_path(repo_root: Path, name: str) -> Path:
     # participant-controlled scripts), so it must never supply executables to
     # the review process.
     return SCRIPT_DIR / name
+
+
+def validate_output_dir(repo_root: Path, out_dir: Path) -> None:
+    try:
+        relative = out_dir.resolve().relative_to(repo_root.resolve())
+    except ValueError:
+        return
+    if relative.parts and relative.parts[0] == "submissions":
+        raise ReviewOutputError("Review output must not be written inside `submissions/`")
 
 
 def run_json_command(command: list[str]) -> dict:
@@ -399,6 +412,10 @@ def main() -> int:
             out_dir = repo_root / out_dir
     else:
         out_dir = repo_root / DEFAULT_OUTPUT_ROOT / submission_dir.name / "review-packet"
+    try:
+        validate_output_dir(repo_root, out_dir)
+    except ReviewOutputError as exc:
+        parser.error(str(exc))
     out_dir.mkdir(parents=True, exist_ok=True)
 
     review_input = build_review_input(repo_root, submission_dir)
--- a/tests/test_review_submission_output_boundary.py
+++ b/tests/test_review_submission_output_boundary.py
@@ -1,0 +1,69 @@
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+
+
+class ReviewSubmissionOutputBoundaryTests(unittest.TestCase):
+    def test_repository_output_must_stay_under_maintainer_review(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            submission = root / "submissions" / "alice" / "demo"
+            submission.mkdir(parents=True)
+            proposal = submission / "proposal.md"
+            proposal.write_text("original proposal\n", encoding="utf-8")
+            out_dir = submission / "review-inputs"
+
+            completed = subprocess.run(
+                [
+                    sys.executable,
+                    str(REPO_ROOT / "scripts" / "review_submission.py"),
+                    str(submission),
+                    "--repo-root",
+                    str(root),
+                    "--out",
+                    str(out_dir),
+                ],
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+
+            self.assertNotEqual(completed.returncode, 0)
+            self.assertIn("must not be written inside `submissions/`", completed.stderr)
+            self.assertFalse(out_dir.exists())
+            self.assertEqual("original proposal\n", proposal.read_text(encoding="utf-8"))
+
+    def test_repository_maintainer_review_output_remains_allowed(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            submission = root / "submissions" / "alice" / "demo"
+            submission.mkdir(parents=True)
+            (submission / "proposal.md").write_text("# Demo\n", encoding="utf-8")
+            out_dir = root / ".maintainer-review" / "demo" / "review-packet"
+
+            completed = subprocess.run(
+                [
+                    sys.executable,
+                    str(REPO_ROOT / "scripts" / "review_submission.py"),
+                    str(submission),
+                    "--repo-root",
+                    str(root),
+                    "--out",
+                    str(out_dir),
+                ],
+                capture_output=True,
+                text=True,
+                check=False,
+            )
+
+            self.assertEqual(completed.returncode, 0, completed.stdout + completed.stderr)
+            self.assertTrue((out_dir / "review-input.json").is_file())
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_review_submission_output_boundary.py
+++ b/tests/test_review_submission_output_boundary.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import sys
 import tempfile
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):

**Exact head reviewed: `5feb2a2b49060d65bd815a8280fbb7850d3793ed`**, against baseline main `8bbaa84cfc7c683db188736f436e4ab98b76babc`. `scripts/review_submission.py` is byte-identical on your PR's parent commit and on current main (blob `a9a2d1cdc0aaae02e058f46654a7b4538597be04`), so the "before" results below apply to main as it stands.

Your reproduction holds, the guard handles the path-normalisation cases correctly, and there is no regression. One addition: there is a call shape in which the stated goal still does not hold, and I have a tested one-line-shaped fix for it.

```
# new test on unpatched main
python3 -m unittest tests.test_review_submission_output_boundary
  -> FAILED (failures=1)
     AssertionError: 0 == 0        # main really does exit 0 and write into the package

# same test on your head
  -> OK (Ran 2 tests)

# existing module, both sides
python3 -m unittest tests.test_review_submission        -> OK (Ran 2 tests)
```

Full-suite baseline:

|---|---|---|
| head `5feb2a2b49060d65bd815a8280fbb7850d3793ed` | 406 | 25 |

Failure sets identical (`diff` empty); all 25 are environment-dependent cases needing `data/`, `assets/` or `collections/`, which my partial checkout omits.

Worth stating explicitly because it changes what the green check means: the workflow runs only `python3 scripts/github_pr_validation.py`, which short-circuits at `is_non_submission_pr` (`scripts/github_pr_validation.py:611`) for a `scripts/` + `tests/` PR. The bot comment here says as much — "non-submission code/docs/test PR; participant package validation was not applicable". **`tests/test_review_submission_output_boundary.py` is never executed by CI**, before or after merge, so the green `submission-validation` is not evidence it passed.

### 2. The guard handles the normalisation cases correctly

Measured on your head, both rejected before `mkdir`, with the package left untouched:

```
--out ./x/../submissions/alice/demo/trav        -> error: Review output must not be written inside `submissions/`
--out <abs>/submissions/alice/demo/abs          -> error: Review output must not be written inside `submissions/`
```

`Path.resolve()` normalises the `..` segment and the absolute form, so neither slips past. The default `.maintainer-review` destination and external output paths still work.

### 3. Residual gap: the guard is relative to `--repo-root`, not to the package

`validate_output_dir()` decides by comparing against `repo_root`:

```python
relative = out_dir.resolve().relative_to(repo_root.resolve())
if relative.parts and relative.parts[0] == "submissions":
    raise ReviewOutputError(...)
```

If `--repo-root` is pointed at a directory that is already inside `submissions/`, the first component of the relative path is the package owner or slug, never the literal `submissions`, so the guard cannot fire. Measured on your head:

```
python3 scripts/review_submission.py demo \
  --repo-root <root>/submissions/alice \
  --out demo/review-inputs

exit = 0
files now inside the participant package:
  submissions/alice/demo/proposal.md
  submissions/alice/demo/review-inputs/review-prompt.md
  submissions/alice/demo/review-inputs/review-input.json
  submissions/alice/demo/review-inputs/advisory-review.md
```

All three review-packet artifacts land in the package, which is the outcome the PR sets out to prevent.

### 4. Suggested repair, verified end to end

Anchor the check on the package being reviewed as well, so it does not depend on where `--repo-root` points, while leaving external destinations alone:

```python
def validate_output_dir(repo_root: Path, out_dir: Path, submission_dir: Path) -> None:
    resolved = out_dir.resolve()
    try:
        resolved.relative_to(submission_dir.resolve())
    except ValueError:
        pass
    else:
        raise ReviewOutputError("Review output must not be written inside `submissions/`")
    try:
        relative = resolved.relative_to(repo_root.resolve())
    except ValueError:
        return
    if relative.parts and relative.parts[0] == "submissions":
        raise ReviewOutputError("Review output must not be written inside `submissions/`")
```

with the call site updated to `validate_output_dir(repo_root, out_dir, submission_dir)`. Measured:

```
your 2 tests + existing 2 tests   -> OK (Ran 4 tests)
the §3 call shape                 -> error: Review output must not be written inside `submissions/`
                                     package contents afterwards: proposal.md only
external --out <tmp>/ext/out      -> exit 0, all three artifacts written as before
```

`submission_dir` is already computed a few lines above the call site, so nothing new has to be threaded through.

- No permission-mode change, no grouping change, and the guard direction is fail-closed in both the original and the fixed form — an unresolvable comparison falls through to "allowed" only when the destination is genuinely outside both roots.
- The script writes to exactly one destination directory (`review-input.json`, `review-prompt.md`, `advisory-review.md` all under `out_dir`), so there is no second or third output path left unguarded.
- Rejection happens before `out_dir.mkdir(...)`, so no directory is created and no partial write occurs, which your test already asserts.
- A symlink loop in `--out` raises an uncaught exception on both main and your head (`OSError [Errno 40]` from `mkdir` before, `RuntimeError` from `resolve()` after). The exception type moves but both sides fail loudly with a non-zero exit, so this is not a behaviour regression.

Thanks for catching this one — writing advisory output into the package under review is the kind of thing that only shows up once it has already happened. I have no view on the design direction; the only thing I am reporting is the one call shape the current anchor cannot see. **If you are already on this, please ignore.**

jiangmuran:
Updated exact head: `4ab286c64e8d820b974e6151d2d69b7a16e3fbeb` on main `233853bfdfdd7309f1f6e8de616a94c052f89235`.

This addresses the outstanding nested `--repo-root` review case and the remaining file-level alias combinations. Lexical and resolved output paths now fail closed on an exact `submissions` component, and each review packet file is published with a same-directory temporary file, flush/fsync, and `os.replace`. Existing symlink/hardlink victims remain byte-identical while the output entry becomes independent.

Verification: 8/8 focused and neighboring tests on current Python and Python 3.9; both `py_compile` runs and `git diff --check` pass. Please re-review the new exact head.

anselasimov-web (commented):
Hi @jiangmuran — I am a **fellow participant, not a maintainer**, with no review or merge authority. Everything below is measured locally and reported as reproducible facts only. All `sha256` values are SHA-256 digests.

**Checked exact head: `4ab286c64e8d820b974e6151d2d69b7a16e3fbeb`** (previous round I reviewed `5feb2a2b49060d65bd815a8280fbb7850d3793ed`), against main `88c6f6ee0bf96d0e4dc17edcf93fd739b38440aa`. `scripts/review_submission.py` is byte-identical on main `550d6db83914e4298862b30705dcdeef3de38450` and on `88c6f6ee0b`, so the "before" numbers below apply to main as it stands (main blob `f30f4485417809466ab0507c38662b32a7455a68`; your head blob `79193cf88a1a3c44ad03ccd9288fb2d5e018c833`). Python 3.10.12.

Short version: **the §3 residual gap from last round is closed and I could not re-open it**, the new alias tests are a real regression gate, and the atomic-write rewrite does what the PR body says. I found two behaviour changes the new commit introduces that main did not have — one of them makes the script unusable for a whole class of checkouts — and both have a tested repair below.

---

### 1. Last round's item, re-checked

| Last round | Status | One line |
|---|---|---|
| §3 guard was anchored on `--repo-root` only, so `--repo-root <root>/submissions/alice` wrote the whole review packet into the package | **Closed** | rc=2, package still contains only `proposal.md` |
| §4 my suggested repair (add `submission_dir` as a second anchor) | **Adopted, and extended** | you also added a lexical candidate and atomic writes |

Replaying last round's §3 call shape on this head:

```
python3 scripts/review_submission.py <T>/A/submissions/alice/demo \
  --repo-root <T>/A/submissions/alice \
  --out <T>/A/submissions/alice/demo/review-inputs
-> rc=2
   review_submission.py: error: Review output must not be written inside `submissions/`
   package afterwards: proposal.md
```

I also checked that the `submission_dir` anchor stands on its own, i.e. that the rejection is not just the new literal-`submissions` rule firing on the temp path. With a layout that contains no `submissions` component anywhere:

```
--repo-root <T>/P/pkgs/alice  --out <T>/P/pkgs/alice/demo/review-inputs
-> rc=2, package afterwards: proposal.md
```

So the anchor added at `scripts/review_submission.py:166-183` is the thing doing the work there.

### 2. The tests are a real gate, and CI still never runs them

```
head 4ab286c64e:                       Ran 5 tests — OK
main script + your test file kept:     Ran 5 tests — FAILED (failures=8)
                                       (4 of the 5 cases red; only the
                                        `.maintainer-review` control stays green)
tests.test_review_submission            3 OK on both sides
```

Test surface I declared for myself: `grep -ln review_submission tests/*.py` → `test_agent_scaffold_and_self_check`, `test_ai_review_submission`, `test_review_submission`, `test_spatial_review`, `test_trusted_review_scripts`, plus your new module. On this head:

```
tests.test_ai_review_submission            Ran 20  OK
tests.test_spatial_review                  Ran 14  OK (skipped=14)
tests.test_trusted_review_scripts          Ran  5  OK
tests.test_agent_scaffold_and_self_check   Ran 33  OK (skipped=27)
python3 -m py_compile scripts/review_submission.py tests/test_review_submission_output_boundary.py   # OK
```

(The skips are my partial checkout, not your change.)

Repeating last round's point because it decides what the green check means, and it is still literally true on today's main: `.github/workflows/submission-validation.yml` is a `pull_request_target` workflow whose only step is `python3 scripts/github_pr_validation.py`, and that script short-circuits at `is_non_submission_pr` (defined `scripts/github_pr_validation.py:511`, used at `:732`) for a `scripts/` + `tests/` PR. The only workflow in the repo that runs `unittest` is `gallery-snapshot-maintenance.yml`, which triggers on push to main / `workflow_dispatch` and runs exactly `tests.test_submissions_gallery`, `tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`. **`tests/test_review_submission_output_boundary.py` is executed by no workflow, before or after merge.** The `submission-validation` check-run at this head is `success` (started 2026-08-16T11:33:55Z, completed 11:36:16Z), and separately the bot's upserted comment still has `updated_at=2026-08-12T20:58:20Z`, i.e. the comment text was not refreshed for this head.

---

### 3. New: a checkout whose path contains a directory named `submissions` can no longer run the script at all

`scripts/review_submission.py:180` tests `"submissions" in candidate.parts` on the *absolute* candidate, so a `submissions` component that belongs to the user's directory layout **above** the repository root counts the same as one inside it. That includes the default destination, which no `--out` flag is involved in:

```
repo checked out at <T>/submissions/mycheckout
python3 scripts/review_submission.py <repo>/submissions/alice/demo --repo-root <repo>
   (no --out at all; default is <repo>/.maintainer-review/demo/review-packet)

main         -> rc=0, wrote advisory-review.md  review-input.json  review-prompt.md
this head    -> rc=2  error: Review output must not be written inside `submissions/`
```

`~/submissions/haidian` is not an exotic layout for this event, and in that layout the script has no working invocation left. A second, milder case is an external destination that happens to have such a component (`<T>/ext/submissions/reviewpacket` → rc=2 on this head, rc=0 on main); that one is a deliberate cost of the rule and I am not asking you to change it.

### 4. New: `write_text_atomically` fixes the mode of new files at `0o644`, ignoring umask

`scripts/review_submission.py:187` uses `path.stat().st_mode & 0o777 if path.exists() else 0o644`. Preserving the mode of an existing file is exactly right; the `else` branch is where the change is, because `Path.write_text()` previously created files as `0o666 & ~umask`:

```
umask 077, default .maintainer-review destination
main       -> 600 600 600   (advisory-review.md, review-input.json, review-prompt.md)
this head  -> 644 644 644
```

`.maintainer-review` is local-only review material, so on a shared machine this widens it from owner-only to world-readable. The PR body's "preserve existing output modes" is accurate for existing files; it is the new-file path that moved.

### 5. Repair I measured end to end

Two independent edits; both keep all five of your tests green.

**(a) judge the `submissions` component relative to the repository root when the destination is inside it.**

```python
    root_resolved = repo_root.resolve()
    for candidate in candidates:
        try:
            parts = candidate.relative_to(root_resolved).parts
        except ValueError:
            parts = candidate.parts
        if "submissions" in parts or any(
            candidate == root or candidate.is_relative_to(root) for root in protected_roots
        ):
            raise ReviewOutputError("Review output must not be written inside `submissions/`")
```

**(b) keep the previous default mode for files that do not exist yet.**

```python
    if path.exists():
        mode = path.stat().st_mode & 0o777
    else:
        current_umask = os.umask(0)
        os.umask(current_umask)
        mode = 0o666 & ~current_umask
```

Measured on top of your head:

| Scenario | your head | head + (a)+(b) |
|---|---|---|
| §3 `--repo-root` inside `submissions/` | rc=2, package intact | rc=2, package intact |
| `submission_dir` anchor with no literal `submissions` component | rc=2, package intact | rc=2, package intact |
| checkout under `<T>/submissions/mycheckout`, default destination | **rc=2, unusable** | rc=0, three files written |
| external `--out <T>/ext/submissions/reviewpacket` | rc=2, nothing created | rc=2, nothing created (unchanged) |
| external `--out <T>/ext2/out` | rc=0 | rc=0 |
| default `.maintainer-review` in an ordinary repo | rc=0 | rc=0 |
| umask 077 / umask 022 new-file modes | 644 / 644 | **600 / 644** |
| `tests.test_review_submission_output_boundary` | Ran 5, OK | Ran 5, **OK** |
| `tests.test_review_submission` | Ran 3, OK | Ran 3, OK |

(a) deliberately leaves the out-of-repo case alone, so `test_dangling_submission_output_symlink_is_rejected_lexically` still fails closed: its lexical candidate is under the repository's own `submissions/`, which the relative form still sees.

### 6. Things I checked that came back clean

- All three artifacts (`review-input.json`, `review-prompt.md`, `advisory-review.md`) go through `write_text_atomically`, so there is no fourth write site left on the old path.
- Rejection still happens before `out_dir.mkdir(...)`; no directory is created and no partial write occurs.
- `os.replace` really does detach the alias: with `--out` pointing at an external directory in which `review-input.json` is a hardlink or symlink to the package's `proposal.md`, the proposal keeps its original bytes and the output stops being a link. That is what your `test_external_output_file_aliases_do_not_overwrite_submission` asserts, and it fails on main.
- No `.review-*.tmp` leftovers after either a success or a rejection.
- A symlink loop in `--out` is now caught by the `except (OSError, RuntimeError)` wrapper and reported as `Review output path cannot be resolved safely: ...` instead of escaping as a bare `RuntimeError` — that is an improvement over both main and the previous head.

### 7. Reproduction

```bash
git fetch origin pull/2222/head:pr2222 && git checkout pr2222
python3 -m unittest tests.test_review_submission_output_boundary

# §3, closed
T=$(mktemp -d); mkdir -p $T/A/submissions/alice/demo
printf 'original proposal\n' > $T/A/submissions/alice/demo/proposal.md
python3 scripts/review_submission.py $T/A/submissions/alice/demo \
  --repo-root $T/A/submissions/alice --out $T/A/submissions/alice/demo/review-inputs; echo rc=$?

# §3 of this comment: a checkout under a directory named submissions
mkdir -p $T/submissions/mycheckout/submissions/alice/demo
printf 'original proposal\n' > $T/submissions/mycheckout/submissions/alice/demo/proposal.md
python3 scripts/review_submission.py $T/submissions/mycheckout/submissions/alice/demo \
  --repo-root $T/submissions/mycheckout; echo rc=$?

# §4 of this comment
mkdir -p $T/U/submissions/alice/demo; printf 'x\n' > $T/U/submissions/alice/demo/proposal.md
( umask 077; python3 scripts/review_submission.py $T/U/submissions/alice/demo --repo-root $T/U )
stat -c '%a %n' $T/U/.maintainer-review/demo/review-packet/*
```

---

Thanks for taking the `submission_dir` anchor and then going further than I suggested — the atomic-replace rewrite closes an alias class I had not tested, and your `test_external_output_file_aliases_do_not_overwrite_submission` proves it in both directions. The two items above are side effects of the new commit rather than anything in the original goal, and both have a repair that keeps your five tests green. I have no view on the design direction. **If you are already on any of this, please ignore.**

CocoSgt (changes requested):
Blocking review for exact head `4ab286c64e8d820b974e6151d2d69b7a16e3fbeb` against current main `0d88f498b6ece0f43a4ef5378acd4de443a09ce1`. The alias protection is valuable and its focused tests pass, but three regressions remain:\n\n1. `validate_output_dir()` searches `"submissions" in candidate.parts` on the absolute path. A valid checkout at `<tmp>/submissions/mycheckout` rejects its default `<repo>/.maintainer-review/...` output, making the command unusable solely because an ancestor outside the repo has that name. Please judge repository-internal components relative to `repo_root`; keep the separate resolved `submission_dir` anchor and the intended external-output policy explicit.\n2. New files are forced to mode `0644`. Under `umask 077`, I reproduced three review outputs becoming world-readable instead of the prior `0600`. Preserve existing modes, but derive new-file mode from `0666 & ~umask` (or use an equivalent creation path) and add `077`/`022` regressions.\n3. `os.fchmod` is unavailable on Windows Python 3.9–3.12 (Windows support was added only in 3.13). Simulating that capability surface produces an uncaught `AttributeError`. Use a Python 3.9+ cross-platform mode-restoration path and add a regression; temporary cleanup already works.\n\nIndependent checks: current focused and neighboring suites 8/8 PASS; `py_compile`, latest-main merge-tree, and `git diff --check` PASS. Reproductions returned ancestor-path rejection, mode `0o644` under umask `077`, and `AttributeError: module os has no attribute fchmod`. Please update all three before re-review.

PR #2223, opened 2026-08-12
fix: reject nonstandard manifest JSON constants

jiangmuran:
## Summary

- reject `NaN`, `Infinity`, and `-Infinity` while decoding manifests
- keep strict audit output machine-readable and include the offending token in its error
- preserve support for standard scalar and structured extension data

## Reproduction

Python JSON decoding accepts nonstandard numeric constants by default. A valid `0.2.0` manifest with `validation_claim.extensions.x-test.data: NaN` therefore completed `validate_manifest_schema.py --strict --json` with exit code 0 and reported `valid: 1`, even though the file is not portable JSON. `Infinity` and `-Infinity` behaved identically.

## Root cause

The audit used the default `json.loads()` constant parser. The extension `data` schema intentionally accepts any JSON value, so schema validation had no reason to reject Python's non-finite float result.

## Validation

- `python3 -m unittest tests.test_manifest_schema_json tests.test_manifest_schema` (17 passed)
- `python3 -m py_compile scripts/validate_manifest_schema.py tests/test_manifest_schema_json.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_manifest_schema.py
+++ b/scripts/validate_manifest_schema.py
@@ -15,6 +15,10 @@ from pathlib import Path
 from manifest_schema import legacy_role_findings, manifest_paths, schema_errors
 
 
+def reject_nonstandard_constant(value: str) -> None:
+    raise ValueError(f"invalid JSON constant: {value}")
+
+
 def main() -> int:
     parser = argparse.ArgumentParser()
     parser.add_argument("--repo-root", default=".")
@@ -34,8 +38,11 @@ def main() -> int:
     for path in paths:
         manifest = None
         try:
-            manifest = json.loads(path.read_text(encoding="utf-8"))
-        except (OSError, UnicodeDecodeError, json.JSONDecodeError) as exc:
+            manifest = json.loads(
+                path.read_text(encoding="utf-8"),
+                parse_constant=reject_nonstandard_constant,
+            )
+        except (OSError, UnicodeDecodeError, json.JSONDecodeError, ValueError) as exc:
             errors = [f"unable to read manifest: {exc}"]
         else:
             if not isinstance(manifest, dict):
--- a/tests/test_manifest_schema_json.py
+++ b/tests/test_manifest_schema_json.py
@@ -1,0 +1,91 @@
+import json
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+SCRIPT = ROOT / "scripts" / "validate_manifest_schema.py"
+
+
+def manifest(extension_data: object) -> dict:
+    return {
+        "schema_version": "0.2.0",
+        "package_id": "sample-package",
+        "project_id": "centennial-jingzhang-ai-belt",
+        "site_package_version": "0.1.0",
+        "submission_stage": "formal",
+        "submission_type": "ai_agent",
+        "agent": {
+            "agent_id": "sample-agent",
+            "agent_name": "Sample Agent",
+            "model": "Model 1",
+        },
+        "generated_at": "2026-08-10T00:00:00Z",
+        "files": [
+            {"path": "evidence.json", "role": "evidence_data", "required": True}
+        ],
+        "validation_claim": {
+            "self_checked": True,
+            "known_blockers": [],
+            "readiness_contract": "persisted-self-check-v1",
+            "extensions": {
+                "x-test": {"schema_version": "1.0", "data": extension_data}
+            },
+        },
+    }
+
+
+class ManifestSchemaJsonTests(unittest.TestCase):
+    def run_audit(self, raw_manifest: str) -> subprocess.CompletedProcess[str]:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            (root / "manifest.json").write_text(raw_manifest, encoding="utf-8")
+            return subprocess.run(
+                [
+                    sys.executable,
+                    str(SCRIPT),
+                    "--repo-root",
+                    str(root),
+                    "--manifest",
+                    "manifest.json",
+                    "--strict",
+                    "--json",
+                ],
+                cwd=ROOT,
+                check=False,
+                capture_output=True,
+                text=True,
+            )
+
+    def test_nonstandard_numeric_constants_are_invalid_json(self):
+        marker = '"NONSTANDARD_CONSTANT"'
+        base = json.dumps(manifest("NONSTANDARD_CONSTANT"))
+
+        for constant in ("NaN", "Infinity", "-Infinity"):
+            with self.subTest(constant=constant):
+                result = self.run_audit(base.replace(marker, constant))
+                payload = json.loads(result.stdout)
+
+                self.assertEqual(result.returncode, 1, result.stdout)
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):

**Exact head reviewed: `f50a27126b0e80983cfd96093626bd26356e3bcd`**, against baseline main `8bbaa84cfc7c683db188736f436e4ab98b76babc`. `scripts/validate_manifest_schema.py` is byte-identical on your PR's parent commit and on current main (blob `8da75778e5b9fbe448b971cc9482b0578064a9e4`), so the "before" results apply to main as it stands.

Your root-cause statement is exactly right, all three constants reproduce in both directions, and there is no regression. Two additions: the effect on the existing corpus is measurably zero, and the same acceptance still exists on the path that actually gates merges.

### 1. Both directions confirmed, for all three constants

Not just `NaN` — I ran each one separately, using your `manifest()` helper's field values:

```
main  8bbaa84cfc7c683db188736f436e4ab98b76babc
  NaN        -> exit 0   valid=1 invalid=0   errors=[]
  Infinity   -> exit 0   valid=1 invalid=0   errors=[]
  -Infinity  -> exit 0   valid=1 invalid=0   errors=[]

head  f50a27126b0e80983cfd96093626bd26356e3bcd
  NaN        -> exit 1   valid=0 invalid=1   ['unable to read manifest: invalid JSON constant: NaN']
  Infinity   -> exit 1   valid=0 invalid=1   ['unable to read manifest: invalid JSON constant: Infinity']
  -Infinity  -> exit 1   valid=0 invalid=1   ['unable to read manifest: invalid JSON constant: -Infinity']
```

Command form:

```
python3 scripts/validate_manifest_schema.py --repo-root <tmp> --manifest manifest.json --strict --json
```

On the test module: `tests.test_manifest_schema_json` fails on main (3 subTest failures) and passes on your head; `tests.test_manifest_schema` passes on both sides (15 tests). The `--json` output stays machine-readable in the failing case, with the offending token inside `results[0].errors`, as the PR describes.

Full-suite baseline:

|---|---|---|
| head `f50a27126b0e80983cfd96093626bd26356e3bcd` | 406 | 25 |

Failure sets identical (`diff` empty); all 25 are environment-dependent cases needing `data/`, `assets/` or `collections/`, absent from my partial checkout.

One mechanical note, since it decides how much the green check is worth here: the workflow runs only `python3 scripts/github_pr_validation.py`, which short-circuits at `is_non_submission_pr` (`scripts/github_pr_validation.py:611`) for a `scripts/` + `tests/` PR — the bot comment on this PR says "non-submission code/docs/test PR; participant package validation was not applicable". **`tests/test_manifest_schema_json.py` is never executed by CI**, so its green status is not evidence the test passed.

### 2. Effect on the existing corpus is exactly zero

I scanned every `submissions/*/*/manifest.json` on main with a `parse_constant` probe:

```
manifests scanned: 725
containing NaN / Infinity / -Infinity: 0
```

So nothing in the repository flips from valid to invalid. This is purely forward-looking hardening, and no participant has to do anything. Worth stating in the PR body, since a reader would reasonably wonder whether 725 packages are about to start failing.

### 3. The same acceptance remains on the enforcing path

The audit script is now strict, but it is not the reader that gates merges. `validate_manifest_file()` reads `manifest.json` through `load_json_file()`, which still uses the default constant parser:

```python
# scripts/validate_submission.py:900-901
def load_json_file(report: ValidationReport, path: Path, display_path: str) -> object | None:
    try:
        return json.loads(path.read_text(encoding="utf-8"))
```

Measured against the same `NaN` fixture, on your head:

```python
r = ValidationReport(changed_files=[])
d = load_json_file(r, Path("manifest_NaN.json"), "submissions/x/y/manifest.json")
```

```
parsed OK, report.errors = []
extensions["x-test"]["data"] = nan      (math.isnan -> True)
```

`scripts/github_pr_validation.py:355` and `:460` read `manifest.json` the same way. So a manifest carrying a non-finite constant is rejected by the standalone audit and accepted by the CI validator — the reverse of the priority a portability rule usually wants.

Passing the same `parse_constant=reject_nonstandard_constant` into `load_json_file()` would close it in one place, since every manifest read in `validate_submission.py` goes through that helper. Whether that belongs in this PR or a follow-up is your call.

### 4. One narrowing suggestion on the exception tuple

`json.JSONDecodeError` is a subclass of `ValueError`:

```
python3 -c "import json; print(issubclass(json.JSONDecodeError, ValueError))"
True
```

so `except (OSError, UnicodeDecodeError, json.JSONDecodeError, ValueError)` collapses to `(OSError, UnicodeDecodeError, ValueError)`, and any `ValueError` raised anywhere in that block now becomes `unable to read manifest: ...`. A dedicated subclass keeps the intent narrow and self-documenting:

```python
class NonstandardJSONConstant(ValueError):
    pass

def reject_nonstandard_constant(value: str) -> None:
    raise NonstandardJSONConstant(f"invalid JSON constant: {value}")

# except (OSError, UnicodeDecodeError, json.JSONDecodeError, NonstandardJSONConstant) as exc:
```

The user-visible message and your assertions are unchanged by this; it only stops the handler from absorbing unrelated failures.

- Standard scalar and structured extension data still validate: `{"score": 1.5}` gives exit 0, `valid=1`, as your second test asserts.
- No permission, grouping, or guard-direction change in the diff; the script has a single input path and no output path, so there is no second or third input left unguarded.
- The change applies on every run, not only under `--strict`, because the decode sits in the shared per-path loop.

Thanks for writing the root cause down precisely — "the extension `data` schema intentionally accepts any JSON value, so schema validation had no reason to reject Python's non-finite float result" is the part that makes this easy to review. I have no view on the design direction; I am only reporting where the same acceptance still lives. **If you are already on this, please ignore.**

CocoSgt (changes requested):
在 exact head `f50a27126b0e80983cfd96093626bd26356e3bcd` 上复核后，这个修复目前只收紧独立 audit，实际 intake/trusted validator 仍接受同一组非标准 JSON 常量，因此暂不能合并。

合成到 latest main `742cdf23` 后，我通过真实 `scripts.validate_submission.load_json_file()` 分别读取 manifest 中的 `NaN`、`Infinity`、`-Infinity`；三者均为 `accepted=True`、`report.errors=[]`，并得到非有限 float。`scripts/github_pr_validation.py` 的 manifest 读取路径同样仍使用默认 `json.loads()`。这会造成 `validate_manifest_schema.py --strict` 拒绝、而真正 gate 路径接受的相反口径，非标准常量仍能绕过投稿校验。

请把同一个 fail-closed JSON loader（或等价的 `parse_constant` 拒绝逻辑）用于 `validate_submission.py` 和 `github_pr_validation.py` 的 manifest 读取路径，并补一条走真实 submission/trusted gate 的回归；只测试 standalone audit 不足以闭合此缺陷。建议用专用异常类型，避免把整个读取块里未来无关的 `ValueError` 都折叠成 JSON 读取错误。

其余验证正常：与 current main 合并无冲突；Python 3.9 环境下 `tests.test_manifest_schema_json` + `tests.test_manifest_schema` 18/18 PASS，三个相关脚本 `py_compile` PASS，`git diff --check` PASS。exact-head `submission-validation` 虽为 SUCCESS，但其 non-submission 分支没有执行这些专项测试。补齐 enforcing path 后可重新复核。

PR #2225, opened 2026-08-12
fix: require changed files in package manifests

jiangmuran:
## Summary

- require changed files in modern professional submission packages to appear in the manifest inventory
- require their existing manifest integrity contract, including `sha256`
- preserve the documented maintainer-only `FEEDBACK.md` exception and historical package compatibility

## Reproduction

On current main, adding `submissions/alice/sample/visual/assets/runtime.js` to an otherwise valid professional package without listing it in `manifest.json` passes `validate_submission()` with `ok: true`. The same result holds when the script contains `fetch(...)`. Because `package_sha256()` hashes only the manifest and manifest-declared artifacts, changing that undeclared executable asset also leaves the reviewed package digest unchanged. An undeclared figure has the same inventory gap.

## Root cause

Manifest validation checked every declared path against disk, but did not apply the reverse check from changed package paths back to the declared inventory.

## Compatibility

The reverse check only applies to `package_type=professional_design_package` and only to changed files that still exist in the checkout. It therefore does not require full historical package hydration. `proposal.md` and `manifest.json` are already declared by the current scaffold; maintainer-only `FEEDBACK.md` remains outside the participant package by design.

## Validation

- `python3 -m unittest tests.test_manifest_changed_inventory tests.test_submission_workflow` (138 passed)
- `python3 -m py_compile scripts/validate_submission.py tests/test_manifest_changed_inventory.py`
- `git diff --check`

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -1613,6 +1613,20 @@ def validate_manifest_file(report: ValidationReport, repo_root: Path, proposal_d
                 report.add_error(
                     f"{proposal_dir}/manifest.json: required file `{required}` must be listed in files"
                 )
+        if package_type == "professional_design_package":
+            changed_prefix = proposal_dir.rstrip("/") + "/"
+            for changed_path in report.changed_files:
+                if not changed_path.startswith(changed_prefix):
+                    continue
+                rel_path = relative_to_proposal(changed_path, proposal_dir)
+                if rel_path == "FEEDBACK.md" or rel_path in listed_paths:
+                    continue
+                if not (repo_root / changed_path).is_file():
+                    continue
+                report.add_error(
+                    f"{proposal_dir}/manifest.json: changed package file `{rel_path}` "
+                    "must be listed in files with its sha256"
+                )
         validate_media_manifest_entries(report, repo_root, proposal_dir, files, listed_paths)
         cover_image = data.get("cover_image")
         if cover_image not in (None, ""):
--- a/tests/test_manifest_changed_inventory.py
+++ b/tests/test_manifest_changed_inventory.py
@@ -1,0 +1,108 @@
+import hashlib
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(ROOT / "scripts"))
+
+from validate_submission import (  # noqa: E402
+    REQUIRED_AI_PACKAGE_FILES,
+    ValidationReport,
+    validate_manifest_file,
+)
+
+
+class ManifestChangedInventoryTests(unittest.TestCase):
+    def write_package(
+        self,
+        root: Path,
+        *,
+        changed_rel: str,
+        declare_changed: bool,
+    ) -> tuple[str, ValidationReport]:
+        proposal_dir = "submissions/alice/sample-package"
+        base = root / proposal_dir
+        package_paths = sorted(REQUIRED_AI_PACKAGE_FILES | {"proposal.md"})
+        for rel_path in package_paths:
+            if rel_path == "manifest.json":
+                continue
+            path = base / rel_path
+            path.parent.mkdir(parents=True, exist_ok=True)
+            path.write_bytes(f"fixture for {rel_path}\n".encode())
+
+        changed_path = base / changed_rel
+        changed_path.parent.mkdir(parents=True, exist_ok=True)
+        changed_path.write_bytes(b"changed package content\n")
+        if declare_changed:
+            package_paths.append(changed_rel)
+
+        files = []
+        for rel_path in sorted(package_paths):
+            item = {"path": rel_path, "role": "asset", "required": True}
+            if rel_path != "manifest.json":
+                item["sha256"] = hashlib.sha256((base / rel_path).read_bytes()).hexdigest()
+            files.append(item)
+        manifest = {
+            "schema_version": "0.2.0",
+            "package_id": "sample-package",
+            "project_id": "centennial-jingzhang-ai-belt",
+            "site_package_version": "0.1.0",
+            "package_type": "professional_design_package",
+            "package_state": "ready_for_review",
+            "submission_stage": "formal",
+            "submission_type": "ai_agent",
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):

**Exact head reviewed: `952d53fcdbcd140f62f5af21d132b7fb47a0a8fa`**, against baseline main `8bbaa84cfc7c683db188736f436e4ab98b76babc`. `scripts/validate_submission.py` is byte-identical on your PR's parent commit and on current main (blob `7456a0e96b4a970725f2040aa043715de72ae70c`), so the "before" numbers apply to main as it stands.

The inventory gap you describe is real and your test proves it in both directions. My addition is a measurement of how wide the new check reaches on the packages already in the repository: it is considerably wider than the Compatibility section suggests, and I can reproduce a hard failure on a merged package with a routine iteration change.

```
# new test on unpatched main
python3 -m unittest tests.test_manifest_changed_inventory
  -> FAILED (failures=2)
     AssertionError: '... changed package file `visual/assets/runtime.js` must be listed
      in files with its sha256' not found in []          # main produces no error at all
     (same for assets/figures/extra.png)

# same test on your head
  -> OK (Ran 3 tests)
```

Full-suite baseline:

|---|---|---|
| head `952d53fcdbcd140f62f5af21d132b7fb47a0a8fa` | 407 | 25 |

Failure sets identical (`diff` empty); the 25 are environment-dependent cases needing `data/`, `assets/` or `collections/`, which my partial checkout omits.

This one deserves the CI note more than the others in your batch, because it is the only PR here that changes a script CI actually executes. The workflow runs `python3 scripts/github_pr_validation.py`, which short-circuits at `is_non_submission_pr` (`scripts/github_pr_validation.py:611`) for a `scripts/` + `tests/` PR — the bot comment says "non-submission code/docs/test PR; participant package validation was not applicable". So **the tightening will take effect for every future submission PR, while the test that verifies it is never executed by CI at all**. A green `submission-validation` here means only that the short-circuit fired.

### 2. A routine iteration on a merged package now fails

`submissions/12442a/jingzhang-second-line` is in the repository today. It has `changelog.md` on disk and does not declare it in `manifest.json`. Simulating an ordinary iteration PR that updates the proposal and its changelog:

```python
r = ValidationReport(changed_files=[
    "submissions/12442a/jingzhang-second-line/proposal.md",
    "submissions/12442a/jingzhang-second-line/changelog.md",
])
validate_manifest_file(r, Path("."), "submissions/12442a/jingzhang-second-line")
```

```
main  8bbaa84cfc7c683db188736f436e4ab98b76babc   ->  ok = True    errors = 0
head  952d53fcdbcd140f62f5af21d132b7fb47a0a8fa   ->  ok = False   errors = 1
  submissions/12442a/jingzhang-second-line/manifest.json:
    changed package file `changelog.md` must be listed in files with its sha256
```

Nothing about that package changed; the file was already there and already undeclared.

### 3. How many packages carry that trigger today

I compared each package's on-disk file list against its declared `files[]` inventory, over every `package_type=professional_design_package` on main:

```
professional_design_package packages:                        724
carrying at least one undeclared on-disk file:               312   (43.1%)
total undeclared files:                                      656

by top-level name / directory:
   174  proposal.en.md
   156  visual/**
   127  changelog.md
   123  assets/**
    40  risk.json
    18  spatial.json
    12  drawings/**
     3  simulation.json
     2  report/**
     1  proposal.zh.md
```

Two of those rows are the ones I would flag:

- **`changelog.md`, 127 packages.** Updating the changelog is the normal accompaniment to an iteration, so for those packages the first PR after this merges is likely to be the one that trips.
- **`proposal.en.md`, 174 packages.** This is a required file under the bilingual contract — `docs/formal-submission-guide.md:11`, `README.md:203` and `docs/terminology-glossary.md:93` all specify that a Chinese primary proposal must ship `proposal.en.md`. So a participant updating a translation they were required to provide would get a hard error telling them the file must be listed with its `sha256`.

The Compatibility section says the reverse check "does not require full historical package hydration", which is accurate — it only inspects changed files that still exist. But the packages are compatible only for as long as nobody touches them; the check skips deleted files and `FEEDBACK.md`, and nothing else.

### 4. The two directions this can go, with the cost of each measured

I am not going to argue for either one — both are defensible and it is your call and the maintainers'.

**Keep it strict.** The `package_sha256()` argument in your PR body is sound: an undeclared asset is outside the reviewed digest. The cost is that 312 packages need an inventory top-up before their next touch, and the error arrives at the participant rather than at whoever left the file undeclared.

**Widen the non-inventory exception.** Extending the `FEEDBACK.md` carve-out to the files that are evidently not meant to be inventory entries:

```python
NON_INVENTORY_PACKAGE_FILES = {
    "FEEDBACK.md", "changelog.md",
    "proposal.en.md", "proposal.zh.md",
    "risk.json", "spatial.json", "simulation.json",
}
...
if rel_path in NON_INVENTORY_PACKAGE_FILES or rel_path in listed_paths:
    continue
```

Measured over the same 724 packages, this moves the affected set from **312 to 58 (8.0%)**, while still catching the `visual/assets/runtime.js` case your test is built around — that one is under `visual/`, which stays covered. The cost is that those seven names sit outside `package_sha256()` coverage by definition.

A third option, if it fits the validator's structure: emit a warning rather than an error when the changed file already existed in the base commit, and an error only when the PR adds it. That preserves the digest argument for new assets without retroactively invoicing existing packages — though I could not test it, since `report.changed_files` does not currently distinguish added from modified.

- The prefix comparison uses `proposal_dir.rstrip("/") + "/"`, so a package named `sample` does not swallow `sample-extra`.
- The check is correctly scoped to `package_type == "professional_design_package"`; older packages without that type are untouched.
- Deletions are skipped by the `is_file()` guard, so removing a file does not demand that it be declared.
- The `FEEDBACK.md` carve-out works as documented, which your third test asserts.
- No permission, grouping, guard-direction, or symlink failure-path change found in this diff.

Thanks for stating the reverse-check idea so plainly — "manifest validation checked every declared path against disk, but did not apply the reverse check" is exactly the gap, and it is worth closing. The only thing I am adding is the size of the existing surface it lands on. **If you are already on this, please ignore.**

PR #2080, opened 2026-08-12
fix(validation): reject boolean metric values

147228:
## Summary

Closes #2071.

Python treats `bool` as a subclass of `int`, so the metrics deterministic gate, visual gate, and spatial recalculation could interpret JSON `true`/`false` as `1`/`0`. This PR introduces one shared `is_json_number()` predicate and uses it across those three gates.

- known metric booleans now fail deterministic validation;
- visual review no longer compares a boolean to an HTML numeric value;
- spatial review reports a blocking metric type issue instead of silently skipping recalculation;
- geometry area declarations and official key-area values receive the same type guard;
- legal numeric ratios `0` and `1` remain valid.

## Verification

- `python3 -m unittest tests.test_metric_types tests.test_visual_review tests.test_spatial_review -q` — 26 passed
- `python3 -m unittest tests.test_submission_workflow.SubmissionWorkflowTests.test_minimal_ai_package_passes_hard_validation tests.test_submission_workflow.SubmissionWorkflowTests.test_v2_simulation_metrics_must_match_task_records -v` — 2 passed
- Python compile checks passed with a workspace-local bytecode cache
- `git diff --check` passed

This is a validation-integrity fix only. It does not alter scoring weights, historical submissions, gallery data, or the protected first-place package.

PR #2229, opened 2026-08-12
fix(validation): reject boolean metric values

### LINKED ISSUE #2071: [bug] metrics JSON boolean 被当作 0/1：绕过 deterministic/visual，并跳过 spatial 复算
## 问题

当前 Python 校验把 JSON 布尔值 `true` / `false` 当作数值 `1` / `0`。因此一个按发布的 `metrics.schema.json` 无效的指标，可以同时绕过 deterministic metrics gate、visual gate，并在 spatial gate 中跳过几何复算比对。

核查基于当前 `main`：`cb83387ebe1c93e0752f6a894098ed2f25a235a5`，Python 3.11.9。

这不是 `unit` / `confidence` 枚举未执行的问题（#1781 已记录那一类）；这里即使字段齐全、枚举合法，只把数字替换成 JSON boolean 也会产生三门不一致。

## 根因

Python 中 `bool` 是 `int` 的子类：

```python
isinstance(True, (int, float))  # True
float(True)                    # 1.0
```

当前逻辑因此分裂：

1. `scripts/validate_submission.py::validate_metrics_file` 用 `isinstance(value, (int, float))` 判断 known numeric 和 ratio，未排除 `bool`，所以 `true` 被接受为合法 ratio；
2. `scripts/visual_review.py::review_visual` 同样未排除 `bool`，并执行 `float(expected)`，所以 HTML 中 `data-value="1"` 可与 JSON 的 `true` 匹配；
3. `scripts/spatial_review.py::metric_value` 已正确写了 `and not isinstance(value, bool)`，但结果是 bool 返回 `None`，`check_metric_close` 随即直接 return，真实几何复算不再与声明值比较；
4. 发布的 `brief/site-package/schemas/metrics.schema.json` 正确要求 `value` 为 `number | null`；`jsonschema` 明确拒绝 `true`。实际 gate 与发布契约不一致。

## 最小复现

`metrics.json`：

```json
{
  "schema_version": "0.1.0",
  "units": {"length": "m", "area": "sqm"},
  "metrics": {
    "site_area_sqm": {
      "status": "known", "value": 1, "unit": "sqm",
      "source_files": [], "formula": "fixture",
      "confidence": "high", "assumptions": []
    },
    "green_ratio": {
      "status": "known", "value": true, "unit": "ratio",
      "source_files": [], "formula": "fixture",
      "confidence": "high", "assumptions": []
    },
    "public_space_ratio": {
      "status": "known", "value": 0, "unit": "ratio",
      "source_files": [], "formula": "fixture",
      "confidence": "high", "assumptions": []
    }
  }
}
```

`visual/index.html` 保留现行必需文本，并声明：

```html
<span data-metric="site_area_sqm" data-value="1">1</span>
<span data-metric="green_ratio" data-value="1">1</span>
<span data-metric="public_space_ratio" data-value="0">0</span>
```

在当前 main 直接调用三方契约，结果为：

```json
{
  "validate_metrics_file": {"ok": true, "errors": []},
  "visual_review": {
    "ok": true,
    "issues": [],
    "metrics_seen": {
      "site_area_sqm": 1.0,
      "green_ratio": 1.0,
      "public_space_ratio": 0.0
    }
  },
  "published_schema_errors": [
    "True is not of type 'number', 'null'"
  ]
}
```

另用 main 上一份完整 formal 包，只把原 `green_ratio=0.123423` 改为 `true`，不改几何：`spatial_review.py --stage formal --json` 仍 `ok=true`，报告复算 `green_ratio=0.123423`，但没有产生 `METRIC_RECALC_MISMATCH`。这证明 bool 路径会静默关闭该指标的空间一致性检查。

## 影响

- `known` 指标可以不是 JSON number，却通过实际 deterministic gate；
- ratio 的值域检查会把 boolean 解释成 0/1；
- visual 可把 `true` 展示为 1 并通过一致性检查；
- 对有几何复算的核心指标，空间 gate 不报类型错，也不报数值不一致；
- 下游 Python 聚合器若复用同类判断，还可能把布尔值计入数值统计。

目前没有证据表明已有投稿利用了这一点；该 Issue 报告的是确定的契约漏洞和未来误填风险，不对既有作者作推断。

## 建议修复与回归

1. 在 metrics 相关代码复用一个严格 helper，例如：

   ```python
   def is_json_number(value: object) -> bool:
       return isinstance(value, (int, float)) and not isinstance(value, bool)
   ```

2. `validate_metrics_file` 的 known numeric 和 ratio 分支均使用该 helper；bool 应产生明确 error。
3. `visual_review` 在读取 `metrics.json` 期望值时同样排除 bool，不把它转换成 0/1。
4. `spatial_review` 若指标存在但值为 bool，建议报告类型问题，而不是把它当“没有可比数值”静默 return；至少 deterministic gate 修复后应有回归覆盖。
5. 增加 `true` 与 `false` 两个 fixture，验证 published schema、deterministic、visual、spatial 对类型错误给出一致结论；同时保留合法整数 0/1 的通过用例，避免误伤。

修复只收紧 JSON 类型语义，不改变合法指标值、评分口径或历史数据含义。

CocoSgt:
## Summary

This is the maintainer port of contributor PR #2080 onto the current trusted `main`; the original branch became conflicting after unrelated validator imports changed.

- add a shared strict JSON-number predicate that rejects Python `bool`
- apply it to deterministic metrics, visual metric sources, spatial metrics, declared polygon areas, and official key-area areas
- preserve valid integer/float values including numeric `0` and `1`
- retain the original contributor attribution in the commit trailer

## Validation

- `.venv/bin/python -m unittest -q tests.test_metric_types tests.test_spatial_review tests.test_visual_review tests.test_submission_workflow` — 161 passed
- `.venv/bin/python -m py_compile scripts/metric_types.py scripts/spatial_review.py scripts/validate_submission.py scripts/visual_review.py`
- `git diff --check`

Supersedes #2080.
Fixes #2071

### DIFF
--- a/scripts/metric_types.py
+++ b/scripts/metric_types.py
@@ -1,0 +1,10 @@
+"""Shared JSON metric type predicates used by the review gates."""
+
+from __future__ import annotations
+
+
+def is_json_number(value: object) -> bool:
+    """Return true only for JSON numbers, never for Python booleans."""
+
+    # Python makes bool a subclass of int, but JSON booleans are not numbers.
+    return isinstance(value, (int, float)) and not isinstance(value, bool)
--- a/scripts/metric_types.py
+++ b/scripts/metric_types.py
@@ -2,9 +2,15 @@
 
 from __future__ import annotations
 
+import math
+
 
 def is_json_number(value: object) -> bool:
     """Return true only for JSON numbers, never for Python booleans."""
 
     # Python makes bool a subclass of int, but JSON booleans are not numbers.
-    return isinstance(value, (int, float)) and not isinstance(value, bool)
+    return (
+        isinstance(value, (int, float))
+        and not isinstance(value, bool)
+        and math.isfinite(value)
+    )
--- a/scripts/spatial_review.py
+++ b/scripts/spatial_review.py
@@ -358,7 +358,7 @@ def check_key_areas(report: SpatialReport, site: Any, key_items: list[tuple[str,
                 )
             )
         expected_area = props.get("official_area_sqm")
-        if isinstance(expected_area, bool):
+        if is_official and isinstance(expected_area, bool):
             report.add(
                 SpatialIssue(
                     "KEY_AREA_AREA_TYPE",
--- a/scripts/spatial_review.py
+++ b/scripts/spatial_review.py
@@ -15,6 +15,8 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
+from metric_types import is_json_number
+
 
 try:
     from pyproj import Transformer
@@ -172,7 +174,18 @@ def load_feature_geometries(
                 )
             )
         declared = props.get("area_sqm_declared")
-        if isinstance(declared, (int, float)) and geom.geom_type in {"Polygon", "MultiPolygon"}:
+        if isinstance(declared, bool):
+            report.add(
+                SpatialIssue(
+                    "DECLARED_AREA_TYPE",
+                    "major",
+                    str(path),
+                    "Declared area must be a JSON number, not a boolean.",
+                    feature_id,
+                    actual=declared,
+                )
+            )
+        elif is_json_number(declared) and geom.geom_type in {"Polygon", "MultiPolygon"}:
             delta = abs(float(declared) - float(geom.area))
             allowed_delta = max(AREA_TOLERANCE_SQM, abs(float(geom.area)) * RELATIVE_AREA_TOLERANCE)
             if delta > allowed_delta:
@@ -345,9 +358,20 @@ def check_key_areas(report: SpatialReport, site: Any, key_items: list[tuple[str,
                 )
             )
         expected_area = props.get("official_area_sqm")
-        if not isinstance(expected_area, (int, float)) and is_official:
+        if isinstance(expected_area, bool):
+            report.add(
+                SpatialIssue(
+                    "KEY_AREA_AREA_TYPE",
+                    "major",
+                    "geometry/key_areas.geojson",
+                    "Official key-area area must be a JSON number, not a boolean.",
+                    feature_id,
+                    actual=expected_area,
+                )
+            )
+        elif not is_json_number(expected_area) and is_official:
             expected_area = OFFICIAL_KEY_AREA_AREAS.get(area_id)
-        if isinstance(expected_area, (int, float)) and is_official:
+        if is_json_number(expected_area) and is_official:
             delta = abs(float(expected_area) - float(geom.area))
             allowed = max(AREA_TOLERANCE_SQM, float(expected_area) * KEY_AREA_RELATIVE_TOLERANCE)
             if delta > allowed:
@@ -394,7 +418,7 @@ def metric_value(metrics: dict, name: str) -> float | None:
     if not isinstance(metric, dict):
         return None
     value = metric.get("value")
-    if isinstance(value, (int, float)) and not isinstance(value, bool):
+    if is_json_number(value):
         return float(value)
     return None
 
@@ -521,6 +545,17 @@ def review_submission(submission_dir: Path, repo_root: Path, stage: str) -> Spat
     }
 
     metrics = load_metrics(submission_dir)
+    for name, metric in metrics.items():
+        if isinstance(metric, dict) and isinstance(metric.get("value"), bool):
+            report.add(
+                SpatialIssue(
+                    "METRIC_VALUE_TYPE",
+                    "major",
+                    "metrics.json",
+                    f"{name} must use a JSON number, not a boolean.",
+                    actual=metric["value"],
+                )
+            )
     check_metric_close(report, metrics, "site_area_sqm", site_area, "sqm")
     check_metric_close(report, metrics, "green_space_area_sqm", green_area, "sqm")
     check_metric_close(report, metrics, "public_space_area_sqm", public_area, "sqm")
@@ -532,7 +567,7 @@ def review_submission(submission_dir: Path, repo_root: Path, stage: str) -> Spat
     for name, metric in metrics.items():
         if isinstance(metric, dict) and metric.get("unit") == "ratio":
             value = metric.get("value")
-            if isinstance(value, (int, float)) and not 0 <= float(value) <= 1:
+            if is_json_number(value) and not 0 <= float(value) <= 1:
                 report.add(
                     SpatialIssue(
                         "METRIC_RATIO_RANGE",
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -18,6 +18,7 @@ from pathlib import Path, PurePosixPath
 from typing import Iterable
 
 from front_matter import parse_front_matter
+from metric_types import is_json_number
 
 
 POLICY_ROOT = Path(__file__).resolve().parents[1]
@@ -1289,9 +1290,9 @@ def validate_metrics_file(report: ValidationReport, path: Path, display_path: st
                 report.add_error(f"{label}: unknown metric value must be null")
             if not metric.get("reason"):
                 report.add_error(f"{label}: unknown metric needs reason")
-        if status == "known" and not isinstance(metric.get("value"), (int, float)):
+        if status == "known" and not is_json_number(metric.get("value")):
             report.add_error(f"{label}: known metric needs numeric value")
-        if metric.get("unit") == "ratio" and isinstance(metric.get("value"), (int, float)):
+        if metric.get("unit") == "ratio" and is_json_number(metric.get("value")):
             value = metric["value"]
             if value < 0 or value > 1:
                 report.add_error(f"{label}: ratio value must be between 0 and 1")
--- a/scripts/visual_review.py
+++ b/scripts/visual_review.py
@@ -18,6 +18,8 @@ from html.parser import HTMLParser
 from pathlib import Path
 from typing import Any
 
+from metric_types import is_json_number
+
 
 REQUIRED_TEXT_MARKERS = [
     "总览地图",
@@ -190,7 +192,7 @@ def review_visual(submission_dir: Path) -> VisualReport:
             )
             continue
         expected = metric.get("value")
-        if not isinstance(expected, (int, float)):
+        if not is_json_number(expected):
             report.add(
                 "VISUAL_METRIC_SOURCE_MISSING",
                 "major",
--- a/tests/test_metric_types.py
+++ b/tests/test_metric_types.py
@@ -1,0 +1,74 @@
+import json
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT / "scripts"))
+
+from validate_submission import ValidationReport, validate_metrics_file  # noqa: E402
+from metric_types import is_json_number  # noqa: E402
+
+
+class MetricTypeTests(unittest.TestCase):
+    def test_booleans_are_not_json_numbers(self) -> None:
+        self.assertTrue(is_json_number(0))
+        self.assertTrue(is_json_number(1.0))
+        self.assertFalse(is_json_number(True))
+        self.assertFalse(is_json_number(False))
+
+    def test_boolean_known_metric_fails_deterministic_validation(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            path = Path(tmp) / "metrics.json"
+            path.write_text(
+                json.dumps(
+                    {
+                        "schema_version": "0.1.0",
+                        "units": {"length": "m", "area": "sqm"},
+                        "metrics": {
+                            "green_ratio": {
+                                "status": "known",
+                                "value": True,
+                                "unit": "ratio",
+                            }
+                        },
+                    }
+                ),
+                encoding="utf-8",
+            )
+            report = ValidationReport()
+            validate_metrics_file(report, path, "metrics.json")
+
+        self.assertFalse(report.ok)
+        self.assertTrue(any("known metric needs numeric value" in error for error in report.errors))
+
+    def test_numeric_zero_and_one_remain_valid_ratio_values(self) -> None:
+        for value in (0, 1):
+            with self.subTest(value=value), tempfile.TemporaryDirectory() as tmp:
+                path = Path(tmp) / "metrics.json"
+                path.write_text(
+                    json.dumps(
+                        {
+                            "schema_version": "0.1.0",
+                            "units": {"length": "m", "area": "sqm"},
+                            "metrics": {
+                                "green_ratio": {
+                                    "status": "known",
+                                    "value": value,
+                                    "unit": "ratio",
+                                }
+                            },
+                        }
+                    ),
+                    encoding="utf-8",
+                )
+                report = ValidationReport()
+                validate_metrics_file(report, path, "metrics.json")
+
+            self.assertTrue(report.ok, report.errors)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/tests/test_metric_types.py
+++ b/tests/test_metric_types.py
@@ -19,6 +19,11 @@ class MetricTypeTests(unittest.TestCase):
         self.assertFalse(is_json_number(True))
         self.assertFalse(is_json_number(False))
 
+    def test_non_finite_values_are_not_json_numbers(self) -> None:
+        for value in (float("nan"), float("inf"), float("-inf")):
+            with self.subTest(value=value):
+                self.assertFalse(is_json_number(value))
+
     def test_boolean_known_metric_fails_deterministic_validation(self) -> None:
         with tempfile.TemporaryDirectory() as tmp:
             path = Path(tmp) / "metrics.json"
--- a/tests/test_spatial_review.py
+++ b/tests/test_spatial_review.py
@@ -119,6 +119,28 @@ class SpatialReviewTests(unittest.TestCase):
             report = review_submission(root / base, REPO_ROOT, "formal")
             self.assertTrue(report.ok, [issue.__dict__ for issue in report.issues])
 
+    def test_boolean_metric_value_is_rejected_before_spatial_recalculation(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            base = "submissions/alice/spatial-boolean-metric"
+            write_valid_spatial_package(root, base)
+            write_json(
+                root,
+                f"{base}/metrics.json",
+                {
+                    "schema_version": "0.1.0",
+                    "units": {"length": "m", "area": "sqm"},
+                    "metrics": {
+                        "green_ratio": {"status": "known", "value": True, "unit": "ratio"},
+                    },
+                },
+            )
+
+            report = review_submission(root / base, REPO_ROOT, "formal")
+
+        self.assertFalse(report.ok)
+        self.assertIn("METRIC_VALUE_TYPE", {issue.check_id for issue in report.issues})
+
     def test_land_use_gap_fails(self) -> None:
         with tempfile.TemporaryDirectory() as tmp:
             root = Path(tmp)
--- a/tests/test_spatial_review.py
+++ b/tests/test_spatial_review.py
@@ -249,6 +249,29 @@ class SpatialReviewTests(unittest.TestCase):
             self.assertFalse(report.ok)
             self.assertIn("KEY_AREA_AREA_MISMATCH", {issue.check_id for issue in report.issues})
 
+    def test_provisional_key_area_ignores_non_authoritative_area_field(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            root = Path(tmp)
+            base = "submissions/alice/spatial-provisional-key-area"
+            write_valid_spatial_package(root, base)
+            path = root / base / "geometry/key_areas.geojson"
+            data = json.loads(path.read_text(encoding="utf-8"))
+            props = data["features"][0]["properties"]
+            props.update(
+                source_type="agent_generated_design",
+                confidence="medium",
+                geometry_role="design_proposal",
+                official_boundary=False,
+                official_area_sqm=True,
+            )
+            write_json(root, f"{base}/geometry/key_areas.geojson", data)
+
+            report = review_submission(root / base, REPO_ROOT, "formal")
+
+        check_ids = {issue.check_id for issue in report.issues}
+        self.assertNotIn("KEY_AREA_AREA_TYPE", check_ids)
+        self.assertIn("KEY_AREA_NOT_OFFICIAL", check_ids)
+
     def test_absolute_metric_drift_is_reported_without_blocking_legacy_package(self) -> None:
         with tempfile.TemporaryDirectory() as tmp:
             root = Path(tmp)
--- a/tests/test_visual_review.py
+++ b/tests/test_visual_review.py
@@ -95,6 +95,19 @@ class VisualReviewTests(unittest.TestCase):
             self.assertFalse(report.ok)
             self.assertIn("VISUAL_METRIC_MISMATCH", {issue.check_id for issue in report.issues})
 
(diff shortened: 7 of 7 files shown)


anselasimov-web (commented):
Hi @CocoSgt — I am a fellow participant in this open call, not a maintainer. This is a voluntary peer review; nothing here is a gate, and the maintainers' call is the only one that counts.

Checked exact head: `22a3065be31ac20d4fb85b925f978e17f41af5a0` (against `main` at `ce51ece794a66f2a07cc9c9aa2bfe9ad60d77f83`).

**Why a manual review at all.** `scripts/github_pr_validation.py` short-circuits on `is_non_submission_pr(files)` (line 611 on current `main`) and returns after emitting an informational comment, so the package validator never runs and — more importantly — **this PR's own new tests are never executed by CI**. The `submission-validation` check-run reporting `success` on this head is therefore an empty green, not evidence that `tests/test_metric_types.py` passes. Everything below is from actually running the code.

Setup used for every command below:

```bash
git clone --filter=blob:none https://github.com/open-city-ai/haidian.git && cd haidian
git fetch origin pull/2229/head:pr2229
pip install shapely pyproj 'jsonschema>=4.18'
```

---

### 1. The defect is real, and the silent-bypass case is worse than the one your test covers

I reproduced on `main` first. The strongest form is in the visual gate: a package whose `metrics.json` declares `"green_ratio": {"status": "known", "value": true}` and whose `visual/index.html` renders `1` passes `review_visual()` **completely silently** — `ok=True`, zero issues:

```bash
# run on main, then on main+PR
python3 - <<'PROBE'
import json, sys, tempfile
from pathlib import Path
sys.path.insert(0, "scripts")
from visual_review import review_visual
HTML = """<!doctype html>
<html lang="zh-CN"><head><meta charset="utf-8"><title>Visual</title></head>
<body><h1>总览地图</h1>
<p>三层范围 重点区域 用地分区 交通慢行 蓝绿公共空间 建筑 更新项目 AI 场景 核心指标 任务覆盖 自检状态 来源 假设</p>
<span data-metric="site_area_sqm" data-value="100">100</span>
<span data-metric="green_ratio" data-value="1">1</span>
<span data-metric="public_space_ratio" data-value="0.1">0.1</span>
</body></html>
"""
with tempfile.TemporaryDirectory() as tmp:
    sub = Path(tmp)/"submissions"/"alice"/"v"; (sub/"visual").mkdir(parents=True)
    (sub/"visual"/"index.html").write_text(HTML, encoding="utf-8")
    (sub/"metrics.json").write_text(json.dumps({"metrics": {
        "site_area_sqm": {"status": "known", "value": 100},
        "green_ratio": {"status": "known", "value": True},
        "public_space_ratio": {"status": "known", "value": 0.1}}}), encoding="utf-8")
    r = review_visual(sub)
    print("ok=", r.ok, "issues=", [i.check_id for i in r.issues])
PROBE
```

- `main` → `ok= True issues= []`
- `main` + this PR → `ok= False issues= ['VISUAL_METRIC_SOURCE_MISSING']`

Your `test_boolean_metric_value_is_not_coerced_to_one` uses the fixture where the HTML declares `0.2`, so on `main` that case already failed — as `VISUAL_METRIC_MISMATCH`, i.e. wrong reason but still a failure. Only the `assertIn("VISUAL_METRIC_SOURCE_MISSING", ...)` line in that test actually binds; the `assertFalse(report.ok)` above it passes on unpatched `main` too. Suggestion: change the fixture so the HTML value is `1`, which is the case that genuinely escapes today. That makes the test both stronger and self-documenting.

### 2. Test double-verification: passes

- `main` + this PR's 7 files, hunk-applied: `python3 -m unittest -q tests.test_metric_types tests.test_spatial_review tests.test_visual_review` → **26 tests, OK**.
- Same tree with only the three production files reverted to `main` (`scripts/spatial_review.py`, `scripts/validate_submission.py`, `scripts/visual_review.py`), tests kept: **3 failures**, exactly the three new tests:
  - `test_boolean_known_metric_fails_deterministic_validation`
  - `test_boolean_metric_value_is_rejected_before_spatial_recalculation`
  - `test_boolean_metric_value_is_not_coerced_to_one`

So the tests bind to the fix rather than passing vacuously. Also for the record: all five modified files' hunks apply cleanly to current `main` even though `scripts/validate_submission.py` has drifted since your branch point (`main` blob `f7ca6b193a` vs your diff base `7456a0e96b`) — no rebase needed as of `ce51ece7`.

### 3. Existing-package impact: zero

I pulled all **731** `submissions/*/*/metrics.json` at `main@ce51ece794a66f2a07cc9c9aa2bfe9ad60d77f83` (GraphQL blob batch) and ran `validate_metrics_file()` on each, on both sides:

| | main | main + PR |
|---|---|---|
| PASS | 731 | 731 |
| FAIL | 0 | 0 |
| verdict changes | — | **0** |
| error-set changes | — | **0** |

Zero merged packages contain a boolean `"value"`. So this is a pure forward-looking tightening with no retroactive breakage — worth stating in the PR body, since a reviewer's first question on any validator-tightening PR is how many existing packages it breaks.

### 4. Unclaimed behaviour change: provisional key areas now block on a field that branch never reads

This one is not in the PR description. In `check_key_areas()`, `is_official` requires `official_boundary is True and geometry_role == "official_constraint"`. A **provisional** key area (`official_boundary: false`, `geometry_role: "provisional_constraint"`) is a legitimate, non-blocking state today.

The new code tests `isinstance(expected_area, bool)` *before* and *outside* the `is_official` guard. Using `write_valid_spatial_package()` from `tests/test_spatial_review.py` and setting on `key_areas.geojson` feature[0]: `official_boundary=False`, `geometry_role="provisional_constraint"`, `official_area_sqm=True`:

- `main` → `ok=True`, issues `['KEY_AREA_PROVISIONAL']` (informational)
- `main` + PR → `ok=False`, issues `['KEY_AREA_AREA_TYPE', 'KEY_AREA_PROVISIONAL']` — now blocking

The asymmetry that suggests it is unintended: the *same* provisional feature with `official_area_sqm: 999999999.0` — a numerically absurd value — still passes on **both** sides, because the recalculation branch is `is_official`-gated. So after this PR a provisional area is blocked for a boolean in a field that is never consumed on that path, but not for a wildly wrong number in the same field.

Same shape at `load_feature_geometries()`: `DECLARED_AREA_TYPE` now fires for a boolean `area_sqm_declared` on non-polygon geometry where `main` was silent. In practice that one is masked — I measured `JSON_SCHEMA_VALIDATION` firing first on that fixture — so it is cosmetic.

Suggested minimum-surface fix, which I ran:

```python
        expected_area = props.get("official_area_sqm")
        if is_official and isinstance(expected_area, bool):
            report.add(SpatialIssue("KEY_AREA_AREA_TYPE", "major", ...))
        elif not is_json_number(expected_area) and is_official:
            expected_area = OFFICIAL_KEY_AREA_AREAS.get(area_id)
```

With that guard the provisional case returns to `ok=True` and all 26 tests still pass. Either behaviour may be what you want — but if the tightening is deliberate it deserves a line in the PR body and a test, because it changes the verdict for a package shape the repo currently accepts.

### 5. `is_json_number` is broader than its own docstring: NaN / Infinity

`scripts/metric_types.py` says "Return true only for JSON numbers." NaN and Infinity are not JSON numbers under RFC 8259 — they are a CPython `json` extension — and `load_json_file()` uses a bare `json.loads()`, which accepts the literals `NaN`, `Infinity`, `-Infinity`. Measured on both sides via `validate_metrics_file()`:

| `metrics.json` value | unit | main | main + PR |
|---|---|---|---|
| `true` | ratio | PASS | **FAIL** (fixed) |
| `NaN` | ratio | PASS | PASS |
| `NaN` | sqm | PASS | PASS |
| `Infinity` | sqm | PASS | PASS |
| `Infinity` | ratio | FAIL | FAIL |

`NaN` slips the ratio range check because every comparison against NaN is `False`, so `value < 0 or value > 1` never trips. This is the same class of bug as the boolean one, one layer down. It is not a regression from your PR — but the new shared helper is exactly the right place to close it, and its current docstring over-promises. Two options:

```python
# option A — keep the name honest
def is_json_number(value: object) -> bool:
    return (
        isinstance(value, (int, float))
        and not isinstance(value, bool)
        and (not isinstance(value, float) or math.isfinite(value))
    )

# option B — reject at the parse boundary in validate_submission.load_json_file
def _reject_constant(name: str):
    raise ValueError(f"{name} is not a JSON number")
json.loads(text, parse_constant=_reject_constant)
```

I ran option A against the full 731-package corpus: still 731/731 PASS, no verdict change, and all 26 tests pass. Option B is the wider fix but touches the shared loader, so it is probably a separate PR. Worth noting `scripts/github_pr_validation.py` also parses with bare `json.loads` in several places.

### 6. One numeric entry point in the same file is still boolean-permissive

Not claimed by this PR, flagging it only because it is adjacent and in a file you already touch — `coordinate_pair_is_valid()` in `scripts/validate_submission.py`:

```bash
python3 -c "import sys; sys.path.insert(0,'scripts'); from validate_submission import coordinate_pair_is_valid; print(coordinate_pair_is_valid([True, False]))"
# main: True    main+PR: True
```

`isinstance(True, (int, float))` is `True` and `-180 <= True <= 180` holds, so a GeoJSON coordinate pair of `[true, false]` is accepted as a valid lon/lat. The simulation-metrics paths at lines ~1396 / ~1437 / ~1467 / ~1496 already carry explicit `not isinstance(..., bool)` guards, so this and `auto_review_queue.py:129` (`score`) look like the only remaining gaps. Swapping in `is_json_number` there would make the sweep complete — your call whether that belongs here or in a follow-up.

### 7. Import path is safe — verified

`from metric_types import is_json_number` is a bare sibling import, matching the existing `from front_matter import parse_front_matter`. I checked the invocation paths: `run_trusted_review_gates()` builds `trusted_repo_root = Path(__file__).resolve().parent.parent` (not a curated file allowlist) and runs `[sys.executable, str(trusted_repo_root/"scripts"/"spatial_review.py"), ...]`, so `sys.path[0]` is the `scripts/` directory and the new module resolves. There is no `python -m scripts.<module>` usage anywhere in the repo or in the workflows. No packaging gap.

---

Net: the core fix is correct, the tests genuinely bind, and it is retroactively safe across all 731 merged packages. Items 4 and 5 are the two I would want addressed before merge — item 4 because it silently changes the verdict for provisional key areas, item 5 because the helper's docstring currently claims more than it delivers. Items 1, 3 and 6 are suggestions.

Thanks for porting #2080 forward and for keeping the contributor attribution in the trailer — the `bool`-is-`int` trap is easy to wave past, and closing it at a shared predicate rather than at each call site is the right shape.

If you are already on any of this, please ignore.

CocoSgt:
已按复核意见补齐 current head `c290974d`：

- `is_json_number()` 现在除布尔值外，也拒绝 JSON/Python 解码器可能接受的 `NaN`、`Infinity` 与 `-Infinity`，避免非有限数进入评分和空间计算。
- `official_area_sqm` 的布尔类型错误只在该 key area 确实是 official 时阻断；provisional/agent-generated geometry 中这个非权威字段继续被忽略，不会错误降低内容评审资格。
- 新增 finite-number 与 provisional-key-area 回归。现有视觉布尔回归已通过 `VISUAL_METRIC_SOURCE_MISSING` 精确绑定，无需改成更弱断言。

验证：metric/spatial/visual/workflow 共 163/163 PASS；`py_compile` 与 `git diff --check` PASS。感谢指出 Python JSON 数值语义与 official/provisional 顺序问题。当前等待新 head trusted CI 与合规 code-owner review。

anselasimov-web (commented):
# PR #2229 follow-up review

Hi @CocoSgt — I am a fellow participant in this open call, not a maintainer. This is a voluntary peer review; nothing here gates anything, and only the maintainers' decision counts.

Checked exact head: `c290974db8282a66d99cd31cb7bb8828bbb4cfce` (git commit sha, full).
Baseline used for every comparison: `main` at `0f051ddb5b91bf5e1992a1b32b8c3b2763978fc8` (git commit sha, full).
Previous head I reviewed: `22a3065be31ac20d4fb85b925f978e17f41af5a0` (git commit sha, full).

Delta since my last review is one commit, four files: `scripts/metric_types.py` (+7/-1, adds `math.isfinite`), `scripts/spatial_review.py` (+1/-1, the `is_official and` guard), `tests/test_metric_types.py` (+5), `tests/test_spatial_review.py` (+23).

Because the predicate itself changed, I did not carry my earlier backtest forward — I re-ran it against the merged corpus at the current `main`. Details below.

---

## 1. Point-by-point on my earlier review

### (1) Backtest safety — **re-measured at this head, still zero regression**

Corpus: all 732 merged packages under `submissions/` at `main` (tree sha `f685da42c6f53cbcbe0164f827298d38009ce875`, git tree sha, full). I pulled 1,464 blobs — every `metrics.json` and every `geometry/key_areas.geojson` — through `git/trees` plus batched GraphQL blob reads, with no full clone.

```
metrics.json files                     732
metric values examined              16,321
  boolean values                         0
  non-finite values                      0
key_areas.geojson files                732
key-area features                    2,196
  official_constraint                    0
  provisional_constraint             2,196
  boolean official_area_sqm              0
  boolean area_sqm_declared              0
  non-finite area_sqm_declared           3
is_json_number() verdict changes, previous head -> this head:   3
is_json_number() verdict changes, main          -> this head:   3
```

All three verdict changes are the same file: `submissions/shawnshi/jingzhang-common-table/geometry/key_areas.geojson`, which carries `"area_sqm_declared": NaN` on three features. A second package, `submissions/fayewong5406/jingzhang-ai-century-echo`, carries `"area_sqm_calculated": NaN` on three features (no gate reads that key).

So the predicate does change verdicts on real merged data. I therefore ran the full gate end to end on both packages rather than reasoning about it:

```
shawnshi/jingzhang-common-table   main: ok=True  3 issues | prev head: identical | this head: identical
fayewong5406/...-century-echo     main: ok=False 4 issues | prev head: identical | this head: identical
```

Issue lists and the `report.metrics` dict are equal element-for-element across all three trees. The reason is that both branches end in "no issue" for a non-finite declared area: previously `abs(nan - area) > allowed` evaluates `False`, and now the `elif` is skipped entirely.

Scope statement, so you can judge how much this covers: I scanned `metrics.json` and `key_areas.geojson` for all 732 packages. I did **not** download the other geometry layers (about 148 MB). For those layers the only field the changed predicate touches is `area_sqm_declared`, and the branch structure there is identical to the `key_areas` case above — which is exactly what the `shawnshi` end-to-end run demonstrates on real data.

### (2) The real bypass is stronger than the self-test — **re-verified, and there is now a second one**

I rebuilt this on a real merged package (`shawnshi/jingzhang-common-table`) instead of a synthetic fixture, changing one metric and one HTML attribute.

**Boolean, as before.** `metrics.json` `green_ratio.value = true`, `visual/index.html` declares `data-value="1"`:

```
main       review_visual ok=True   0 issues | validate_metrics_file 0 errors
prev head  review_visual ok=False  VISUAL_METRIC_SOURCE_MISSING | 1 error
this head  review_visual ok=False  VISUAL_METRIC_SOURCE_MISSING | 1 error
```

**Non-finite, which is new at this head.** `metrics.json` `site_area_sqm.value = NaN`, `visual/index.html` declares `data-value="999999999"` against a real site area of about 11.4 million sqm:

```
main       review_visual ok=True   0 issues | validate_metrics_file 0 errors
prev head  review_visual ok=True   0 issues | validate_metrics_file 0 errors
this head  review_visual ok=False  VISUAL_METRIC_SOURCE_MISSING | 1 error
```

This one is arguably the stronger of the two. With a boolean the displayed number still has to equal 1; with `NaN`, `abs(displayed - float(nan)) > tolerance` is `False` for *every* displayed number, so the HTML-to-`metrics.json` cross-check passes unconditionally. Adding `math.isfinite` closes that, and it closes it in the same three places as the boolean case rather than as a special case. Worth mentioning in the PR body — right now the description reads as a small hardening, and this is a live bypass on current `main`.

### (3) Provisional `official_area_sqm: true` would turn blocking — **resolved**

Verified end to end on a real merged package, setting `official_area_sqm: true` on its three provisional features (`official_boundary: false`, `geometry_role: "provisional_constraint"`):

```
main       ok=True   {KEY_AREA_PROVISIONAL: 3}
prev head  ok=False  {KEY_AREA_PROVISIONAL: 3, KEY_AREA_AREA_TYPE: 3}   <- would newly block
this head  ok=True   {KEY_AREA_PROVISIONAL: 3}                          <- identical to main
```

The intended tightening for official areas survives. Flipping one feature to `official_boundary: true` + `geometry_role: "official_constraint"` with `official_area_sqm: true`:

```
main       ok=False  {KEY_AREA_AREA_MISMATCH: 1, KEY_AREA_PROVISIONAL: 2}
prev head  ok=False  {KEY_AREA_AREA_TYPE: 1,    KEY_AREA_PROVISIONAL: 2}
this head  ok=False  {KEY_AREA_AREA_TYPE: 1,    KEY_AREA_PROVISIONAL: 2}
```

Note the side benefit: `main` reports this as an area *mismatch*, because it silently compares the polygon against `float(True) == 1.0`. Both PR heads report it as a *type* error, which is the accurate diagnosis.

One corpus fact that makes this guard load-bearing: 2,196 of 2,196 merged key-area features are provisional and zero are official. Under the previous head the new check could only ever have fired on provisional geometry.

---

## 2. New findings at this head

### N1. Non-finite JSON literals are already in the merged corpus, and the predicate only guards values it is applied to

`json.loads` accepts bare `NaN`, `Infinity` and `-Infinity`; the JSON spec does not. Two of 732 merged packages contain them, so this is not hypothetical. `is_json_number()` now catches such a value wherever it is called, but a property that no gate reads — `area_sqm_calculated` in `fayewong5406/jingzhang-ai-century-echo` is the live example — passes straight through. The schema layer does not catch it either: `validate_schema()` runs on those same files today and both packages are merged.

If you or the maintainers ever want to close this at the parse layer, the measured blast radius is exactly those 2 packages. I tried the obvious version and it needs care: adding `parse_constant=` to `spatial_review.load_json()` propagates an uncaught exception out of `review_submission()` via `validate_schema()` (line 127), not only via `load_feature_geometries()`, so every caller of `load_json` needs a handler that turns it into a report issue. I would treat that as a separate change rather than growing this PR.

### N2. `METRIC_VALUE_TYPE` checks booleans only, so a non-finite metric is silent in the spatial report

The new loop in `review_submission()` tests `isinstance(metric.get("value"), bool)`. With `site_area_sqm: NaN`:

```
main / prev head / this head:   spatial ok=True, no METRIC_VALUE_TYPE
```

The package is still blocked overall, by `validate_submission.validate_metrics_file` and by `review_visual` — so this is not a hole in the merged pipeline. But `metric_value()` now returns `None` for a non-finite value, `check_metric_close()` skips silently, and the spatial report gives a reviewer no indication that a declared metric was discarded. Since you already imported `is_json_number` into this module, the guard can express the whole intent for one line. Tested (see S1 below).

### N3. CI gives no signal on this PR either

Same short-circuit as on your other tool PR: `GET /repos/.../commits/c290974db8282a66d99cd31cb7bb8828bbb4cfce/check-runs` returns exactly one check-run, `submission-validation`, and its published body is the `non-submission code/docs/test PR` warning. No workflow runs `tests.test_metric_types`, `tests.test_spatial_review`, `tests.test_visual_review` or `tests.test_submission_workflow` — `gallery-snapshot-maintenance.yml` runs three unrelated modules on push to `main`, and nothing else invokes `unittest`. Everything above is from executing the code locally.

### N4. The base is behind current `main`, in the same file you touch

PR base is `7236317c35f53e9acdd58d7992391247bb979cbc`; `main` was `0f051ddb5b91bf5e1992a1b32b8c3b2763978fc8` when I started and `bc2cb1fbe329a838eda92e0cb21d7a84a8db3b8c` when I finished. Since your base, `main` added a `completeness_limited_by` rule to `validate_design_depth_matrix_file()` in `scripts/validate_submission.py` — a different hunk from yours, and the API currently reports `mergeable: true`, so no action is needed. Flagging it only because diffing this branch against `main` rather than against the merge base makes that rule look like a deletion by this PR, which it is not.

---

## 3. Reproduction

```bash
git clone --depth 1 --filter=blob:none --no-checkout --sparse \
    -c core.autocrlf=false https://github.com/open-city-ai/haidian.git
cd haidian && git sparse-checkout set scripts tests schema brief scenarios data docs && git checkout
git fetch origin pull/2229/head --depth 2 && git checkout FETCH_HEAD

# (2) the non-finite bypass, on a real merged package
#   metrics.json:      site_area_sqm.value -> NaN     (bare literal, python json accepts it)
#   visual/index.html: data-value="999999999"
python3 - <<'PY'
import sys; sys.path.insert(0,'scripts')
from pathlib import Path
from visual_review import review_visual
from validate_submission import ValidationReport, validate_metrics_file
pkg = Path('submissions/shawnshi/jingzhang-common-table')   # local copy, edited as above
rv  = review_visual(pkg)
rep = ValidationReport(); validate_metrics_file(rep, pkg/'metrics.json', 'metrics.json')
print('review_visual ok:', rv.ok, [i.check_id for i in rv.issues])
print('validator errors:', rep.errors)
PY

# (3) provisional key area with a boolean official_area_sqm
python3 -c "import sys; sys.path.insert(0,'scripts'); \
from pathlib import Path; from spatial_review import review_submission; \
r=review_submission(Path('submissions/shawnshi/jingzhang-common-table'), Path('.'), 'formal'); \
print(r.ok, sorted(i.check_id for i in r.issues))"

# (1) corpus scan, trees + blobs only, no checkout of submissions/
#   GET /repos/open-city-ai/haidian/git/trees/f685da42c6f53cbcbe0164f827298d38009ce875?recursive=1
#   then batched GraphQL: repository { object(oid:"<blob sha>") { ... on Blob { text } } }

# tests
PYTHONPATH=scripts:. python3 -m unittest tests.test_metric_types tests.test_spatial_review
```

Test scope note: `tests/` has no `__init__.py`, so `unittest discover` does not work; I ran modules individually. At this head: `test_metric_types` 4, `test_spatial_review` 12, `test_visual_review` 12, `test_submission_workflow` 135 — **163 passed, 0 failed**, matching the 163/163 in your comment exactly.

---

## 4. Suggestion, tested end to end

**S1 — let `METRIC_VALUE_TYPE` cover the whole intent (addresses N2).** One line, reusing the predicate already imported in this module:

```python
    for name, metric in metrics.items():
        if not isinstance(metric, dict):
            continue
        value = metric.get("value")
        if metric.get("status") == "known" and not is_json_number(value):
            report.add(
                SpatialIssue(
                    "METRIC_VALUE_TYPE",
                    "major",
                    "metrics.json",
                    f"{name} must use a finite JSON number.",
                    actual=str(value),
                )
            )
```

Measured against the same fixtures:

```
package                                      this head                    with S1
unmodified shawnshi (merged as-is)           ok=True  {PROVISIONAL:3}     unchanged
provisional key area + official_area_sqm:true ok=True {PROVISIONAL:3}     unchanged
green_ratio: true                            METRIC_VALUE_TYPE            METRIC_VALUE_TYPE
site_area_sqm: NaN                           (silent)                     METRIC_VALUE_TYPE
```

`test_metric_types`, `test_spatial_review` and `test_visual_review` all still pass with S1 applied (4 / 12 / 12). The `status == "known"` condition is what keeps unknown and not_applicable metrics out of it; without that condition a `null` value would trip the check.

This is optional — the package is already blocked by the hard validator either way, so it only affects what a reviewer sees in the spatial report.

---

Thanks for the fast turnaround, and specifically for scoping the `official_area_sqm` check to `is_official` rather than relaxing the check itself. That is the version I was hoping for, and the `math.isfinite` addition turned out to close a bypass I had not looked for.

If you are already on any of this, please ignore.

CocoSgt:
Superseded by merged PR #2574 (`ce282f2b45cef7df91e71de7255320842192661e`). The external port preserves this PR’s reviewed commits and original/co-author attribution, integrates current-main Git-blob hashing, and adds the final fail-closed regression for positive/negative oversized JSON integers. Exact current-main verification passed 174 tests, py_compile, diff-check, direct huge-integer replay, and trusted CI. Closing this conflicting maintainer-authored branch avoids duplicate implementations; issue #2071 is now closed by the merged external port.

PR #2237, opened 2026-08-12
fix(validation): support verified GitHub rename aliases

### LINKED ISSUE #2210: 请求明确 GitHub 改名后的历史投稿维护路径：PR 作者已更新，但目录 owner 被 CI 阻断
## 背景

我的 GitHub 账号为同一账号（GitHub user id: `51290995`），登录名已由历史用户名 `zymk8353` 改为当前用户名 `zyaoii`。GitHub API / 网页现在会把既有 PR 作者显示为 `zyaoii`，但四个已合并投稿仍保留历史 owner 路径与包内身份：

- PR #1851 / #1919：`submissions/zymk8353/jingzhang-safe-return-line/`
- PR #1868 / #1920：`submissions/zymk8353/jingzhang-safe-charge-line/`
- PR #1909 / #1921：`submissions/zymk8353/jingzhang-ready-aed-line/`
- PR #1913：`submissions/zymk8353/jingzhang-level-access-line/`

后续新投稿已使用当前身份，例如：

- PR #2015：`submissions/zyaoii/jingzhang-dignity-supply-line/`
- PR #2058：`submissions/zyaoii/jingzhang-passage-first/`

## 可复现问题

以当前 `main` 的正式校验链运行时：

- 历史四包使用 `--pr-author zymk8353` 可通过 owner/path 一致性检查；
- 若当前账号 `zyaoii` 对历史目录发起维护 PR，校验会因 `PR author != submissions/<owner>/` 以及 `author_github` / agent id 不一致而阻断；
- 本地只读模拟中，归药线、安心充分别出现 41/41、42/42 个 changed-file owner/path 阻断，五分钟与等高线也会确定性失败。

相关规则位于：

- `scripts/participant_preflight.py`
- `scripts/validate_submission.py`
- `schemas/submission-manifest.schema.json`

## 请求维护者明确支持路径

为避免重复投稿、gallery 重复、历史链接损坏或绕过 CI，烦请确认推荐哪一种：

1. **稳定账号 alias（优先）**：在校验层将同一 GitHub user id 的 `zymk8353 -> zyaoii` 视为历史别名，仅用于维护已合并旧路径；历史目录和公开链接不移动。
2. **维护者受控原子迁移**：由维护者统一移动四个目录，并同步 `author_github`、agent id、自检记录及 gallery/publication 映射或重定向。

也请确认在正式机制落地前，是否应保持旧目录只读，不由 `zyaoii` 直接复制、删除或改名。

本 Issue 只请求身份维护规则，不请求改分、重新评审或实施背书；收到明确口径前我不会创建重复投稿目录。

CocoSgt:
## Summary

- add a maintainer-controlled participant owner alias policy keyed by stable GitHub user ID
- verify the live `pull_request_target` user ID and current login before granting any legacy ownership
- authorize only four explicit, already merged historical packages; new packages and other legacy paths remain denied
- preserve historical directories, `author_github`, agent IDs, public URLs, and audit history
- document the verification and read-only migration procedure

## Security boundary

Aliases cannot be supplied by participants. A matching login without the stable user ID, a matching user ID with the wrong current login, a legacy directory owned by a different historical login, and an unlisted package all fail closed. Maintainer bypass, participant scope, and trusted-base execution remain unchanged.

## Validation

- `.venv/bin/python -m unittest -q tests.test_submission_workflow` — 140 passed
- `.venv/bin/python -m py_compile scripts/github_pr_validation.py scripts/validate_submission.py tests/test_submission_workflow.py`
- `.venv/bin/python -m json.tool data/participant_owner_aliases.json`
- `git diff --check`

Fixes #2210

### DIFF
--- a/data/participant_owner_aliases.json
+++ b/data/participant_owner_aliases.json
@@ -1,0 +1,16 @@
+{
+  "schema_version": "0.1.0",
+  "aliases": [
+    {
+      "github_user_id": 51290995,
+      "current_login": "zyaoii",
+      "legacy_login": "zymk8353",
+      "legacy_submission_dirs": [
+        "submissions/zymk8353/jingzhang-safe-return-line",
+        "submissions/zymk8353/jingzhang-safe-charge-line",
+        "submissions/zymk8353/jingzhang-ready-aed-line",
+        "submissions/zymk8353/jingzhang-level-access-line"
+      ]
+    }
+  ]
+}
--- a/docs/participant-owner-aliases.md
+++ b/docs/participant-owner-aliases.md
@@ -11,4 +11,6 @@ Each exception binds all of the following trusted values:
 
 The exception permits the current account to maintain only those listed historical packages. It does not authorize new packages under the old login, other users with a similar name, repository infrastructure, another historical owner, or participant-declared aliases. Historical directories, `author_github` values, public URLs, and agent identifiers remain unchanged so the audit chain is preserved.
 
+GitHub may release an old login for re-registration. Once a legacy login is listed in this trusted policy, a different stable user ID cannot use that login's ordinary path rule to modify its historical packages. The code machine-checks the live user ID and current login; GitHub does not expose rename history, so the claim that the legacy login belonged to that ID remains an explicit maintainer attestation.
+
 Before adding an entry, maintainers must verify the stable user ID against the live GitHub API and the authorship of every listed historical pull request. Until that review is complete, the old directory stays read-only; do not copy, rename, or delete it to work around validation.
--- a/docs/participant-owner-aliases.md
+++ b/docs/participant-owner-aliases.md
@@ -1,0 +1,14 @@
+# GitHub rename aliases for historical submissions
+
+Participant directories normally match the pull-request author's current GitHub login exactly. A GitHub account rename does not move historical submission directories, so maintainers may add a narrow exception to `data/participant_owner_aliases.json`.
+
+Each exception binds all of the following trusted values:
+
+- the stable numeric GitHub user ID from the `pull_request_target` event;
+- the account's current login;
+- one historical login;
+- an explicit list of already merged, three-component proposal directories owned by that historical login.
+
+The exception permits the current account to maintain only those listed historical packages. It does not authorize new packages under the old login, other users with a similar name, repository infrastructure, another historical owner, or participant-declared aliases. Historical directories, `author_github` values, public URLs, and agent identifiers remain unchanged so the audit chain is preserved.
+
+Before adding an entry, maintainers must verify the stable user ID against the live GitHub API and the authorship of every listed historical pull request. Until that review is complete, the old directory stays read-only; do not copy, rename, or delete it to work around validation.
--- a/scripts/github_pr_validation.py
+++ b/scripts/github_pr_validation.py
@@ -403,7 +403,14 @@ def authorized_legacy_submission_dirs(
     if not isinstance(aliases, list):
         return set()
     for item in aliases:
-        if not isinstance(item, dict) or item.get("github_user_id") != github_user_id:
+        if not isinstance(item, dict):
+            continue
+        configured_user_id = item.get("github_user_id")
+        if (
+            not isinstance(configured_user_id, int)
+            or isinstance(configured_user_id, bool)
+            or configured_user_id != github_user_id
+        ):
             continue
         configured_login = item.get("current_login")
         if not isinstance(configured_login, str) or configured_login.casefold() != current_login.casefold():
@@ -429,6 +436,32 @@ def authorized_legacy_submission_dirs(
             result.add(PurePosixPath(value).as_posix())
         return result
     return set()
+
+
+def reserved_legacy_login_user_id(repo_root: Path, login: str) -> int | None:
+    """Return the trusted user ID that owns a released historical login."""
+    try:
+        policy = json.loads(
+            (repo_root / PARTICIPANT_OWNER_ALIASES_PATH).read_text(encoding="utf-8")
+        )
+    except (OSError, UnicodeDecodeError, json.JSONDecodeError):
+        return None
+    aliases = policy.get("aliases") if isinstance(policy, dict) else None
+    if not isinstance(aliases, list):
+        return None
+    for item in aliases:
+        if not isinstance(item, dict):
+            continue
+        legacy_login = item.get("legacy_login")
+        user_id = item.get("github_user_id")
+        if (
+            isinstance(legacy_login, str)
+            and legacy_login.casefold() == login.casefold()
+            and isinstance(user_id, int)
+            and not isinstance(user_id, bool)
+        ):
+            return user_id
+    return None
 
 
 def is_review_queue_candidate(
@@ -660,6 +693,12 @@ def main() -> int:
     authorized_legacy_dirs = authorized_legacy_submission_dirs(
         trusted_repo_root, pr_user_id, pr_author
     )
+    reserved_legacy_user_id = reserved_legacy_login_user_id(trusted_repo_root, pr_author)
+    blocked_submission_owners = (
+        {pr_author}
+        if reserved_legacy_user_id is not None and reserved_legacy_user_id != pr_user_id
+        else set()
+    )
     validation_files = validation_paths_for(files, maintainer_bypass)
     queue_candidate = is_review_queue_candidate(changed_files, pr_author, authorized_legacy_dirs)
 
@@ -748,6 +787,7 @@ def main() -> int:
                 strict_manifest_paths=strict_manifest_paths_for(files),
                 required_readiness_contract_dirs=required_readiness_contract_dirs,
                 authorized_legacy_submission_dirs=authorized_legacy_dirs,
+                blocked_submission_owners=blocked_submission_owners,
             )
             if validation_files and not base_sha:
                 validation.add_error(
--- a/scripts/github_pr_validation.py
+++ b/scripts/github_pr_validation.py
@@ -45,6 +45,7 @@ RETRYABLE_METHODS = frozenset({"GET", "HEAD", "OPTIONS"})
 # with the path and head-specific download error after the retry budget.
 RETRYABLE_STATUS_CODES = frozenset({404, 429, 500, 502, 503, 504})
 TRUSTED_REVIEW_GATE_TIMEOUT_SECONDS = 180
+PARTICIPANT_OWNER_ALIASES_PATH = "data/participant_owner_aliases.json"
 
 
 def _http_error_message(error: urllib.error.HTTPError) -> str:
@@ -387,14 +388,67 @@ def strict_manifest_paths_for(files: list[dict]) -> list[str]:
     return sorted(paths)
 
 
-def is_review_queue_candidate(changed_files: list[str], pr_author: str) -> bool:
+def authorized_legacy_submission_dirs(
+    repo_root: Path, github_user_id: object, current_login: str
+) -> set[str]:
+    """Resolve maintainer-controlled rename aliases from trusted GitHub identity."""
+    if not isinstance(github_user_id, int) or isinstance(github_user_id, bool):
+        return set()
+    policy_path = repo_root / PARTICIPANT_OWNER_ALIASES_PATH
+    try:
+        policy = json.loads(policy_path.read_text(encoding="utf-8"))
+    except (OSError, UnicodeDecodeError, json.JSONDecodeError):
+        return set()
+    aliases = policy.get("aliases") if isinstance(policy, dict) else None
+    if not isinstance(aliases, list):
+        return set()
+    for item in aliases:
+        if not isinstance(item, dict) or item.get("github_user_id") != github_user_id:
+            continue
+        configured_login = item.get("current_login")
+        if not isinstance(configured_login, str) or configured_login.casefold() != current_login.casefold():
+            continue
+        legacy_login = item.get("legacy_login")
+        if not isinstance(legacy_login, str) or not legacy_login:
+            return set()
+        legacy_dirs = item.get("legacy_submission_dirs")
+        if not isinstance(legacy_dirs, list):
+            return set()
+        result: set[str] = set()
+        for value in legacy_dirs:
+            if not isinstance(value, str):
+                return set()
+            parts = PurePosixPath(value).parts
+            if (
+                len(parts) != 3
+                or parts[0] != "submissions"
+                or parts[1] != legacy_login
+                or ".." in parts
+            ):
+                return set()
+            result.add(PurePosixPath(value).as_posix())
+        return result
+    return set()
+
+
+def is_review_queue_candidate(
+    changed_files: list[str],
+    pr_author: str,
+    authorized_legacy_dirs: set[str] | None = None,
+) -> bool:
     """Return true only for a single participant-owned submission directory."""
+    authorized_legacy_dirs = authorized_legacy_dirs or set()
     roots: set[str] = set()
     for filename in changed_files:
         parts = filename.split("/")
-        if len(parts) < 4 or parts[0] != "submissions" or parts[1].casefold() != pr_author.casefold():
+        proposal_dir = "/".join(parts[:3])
+        if (
+            len(parts) < 4
+            or parts[0] != "submissions"
+            or (parts[1].casefold() != pr_author.casefold() and proposal_dir not in authorized_legacy_dirs)
+        ):
             return False
-        roots.add("/".join(parts[:3]))
+        roots.add(proposal_dir)
     return bool(changed_files) and len(roots) == 1
 
 
@@ -573,6 +627,7 @@ def main() -> int:
 
     pr_number = int(pull_request["number"])
     pr_author = pull_request["user"]["login"]
+    pr_user_id = pull_request["user"].get("id")
     head_repo = pull_request["head"]["repo"]["full_name"]
     head_sha = pull_request["head"]["sha"]
     base = pull_request.get("base") or {}
@@ -601,8 +656,12 @@ def main() -> int:
         if item.strip()
     ]
     maintainer_bypass = pr_author.lower() in {item.lower() for item in bypass}
+    trusted_repo_root = Path(__file__).resolve().parent.parent
+    authorized_legacy_dirs = authorized_legacy_submission_dirs(
+        trusted_repo_root, pr_user_id, pr_author
+    )
     validation_files = validation_paths_for(files, maintainer_bypass)
-    queue_candidate = is_review_queue_candidate(changed_files, pr_author)
+    queue_candidate = is_review_queue_candidate(changed_files, pr_author, authorized_legacy_dirs)
 
     # Code/docs/test PRs do not need participant-package hydration.  Decide
     # this immediately after the file listing so a non-submission change
@@ -688,12 +747,12 @@ def main() -> int:
                 bypass,
                 strict_manifest_paths=strict_manifest_paths_for(files),
                 required_readiness_contract_dirs=required_readiness_contract_dirs,
+                authorized_legacy_submission_dirs=authorized_legacy_dirs,
             )
             if validation_files and not base_sha:
                 validation.add_error(
                     "pull_request.base.sha is required to establish the trusted readiness migration boundary"
                 )
-            trusted_repo_root = Path(__file__).resolve().parent.parent
             for proposal_path in sorted(proposal_paths):
                 run_trusted_review_gates(
                     validation,
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -182,6 +182,7 @@ ALLOWED_VISUAL_ASSET_EXTENSIONS = {".css", ".js", ".json", ".svg", ".png", ".jpg
 PARTICIPANT_PROTECTED_GLOBAL_FILES = {
     "submissions-data.js",
     "gallery-publication.json",
+    "data/participant_owner_aliases.json",
 }
 MAINTAINER_CONTROLLED_SUBMISSIONS_ROOT_FILES = {
     "submissions/README.md",
@@ -2801,6 +2802,7 @@ def validate_submission(
     required_readiness_contract_dirs: Iterable[str] = (),
     strict_manifest_paths: Iterable[str] = (),
     authorized_legacy_submission_dirs: Iterable[str] = (),
+    blocked_submission_owners: Iterable[str] = (),
 ) -> ValidationReport:
     report = ValidationReport()
     repo_root = repo_root.resolve()
@@ -2818,6 +2820,11 @@ def validate_submission(
     authorized_legacy_dirs = {
         normalize_changed_path(path).rstrip("/")
         for path in authorized_legacy_submission_dirs
+    }
+    blocked_owners = {
+        str(owner).strip().casefold()
+        for owner in blocked_submission_owners
+        if str(owner).strip()
     }
 
     if not pr_author or not GITHUB_LOGIN_RE.match(pr_author):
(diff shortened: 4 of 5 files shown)


anselasimov-web (commented):
Hi @CocoSgt — I am a fellow participant in this open call, not a maintainer. This is a voluntary peer review; nothing here is a gate, and the maintainers' call is the only one that counts.

Checked exact head: `64820777b7b5afb6c28b1c7f9a401e8b3ccf18aa` (against `main` at `ce51ece794a66f2a07cc9c9aa2bfe9ad60d77f83`).

Because this PR changes an authorization boundary, I spent the review on the trust model rather than on style. **Note that CI gives no signal here at all**: `scripts/github_pr_validation.py` short-circuits on `is_non_submission_pr(files)` (line 611) for a `scripts/`+`data/`+`docs/` PR, so the `submission-validation` check-run showing `success` on this head never ran `tests/test_submission_workflow.py`. Everything below is from executing the code.

---

### 1. The core trust boundary holds — I tried to break it and could not

The question I set out to answer: can a participant PR that *also* modifies `data/participant_owner_aliases.json` self-certify an alias and use it within that same PR? **No, and for two independent reasons.** Both verified:

**(a) The policy file is read from the trusted checkout, never from the PR head.** `.github/workflows/submission-validation.yml` checks out `ref: ${{ github.event.repository.default_branch }}`, and your `main()` change passes `trusted_repo_root = Path(__file__).resolve().parent.parent` into `authorized_legacy_submission_dirs()`. I confirmed the read is root-scoped by planting a forged policy in a throwaway tree:

```python
# forged tree grants attacker -> submissions/victim/prize-package
authorized_legacy_submission_dirs(forged_root, 999999, "attacker")  # -> {'submissions/victim/prize-package'}
authorized_legacy_submission_dirs(trusted_root, 999999, "attacker") # -> set()
```

**(b) Even if (a) were bypassed, the combined PR is rejected on the path rule.** A PR touching `submissions/victim/prize-package/**` plus `data/participant_owner_aliases.json`, authored by `attacker`:

```
ok= False
 - data/participant_owner_aliases.json: participant PRs may only change submissions/attacker/
 - submissions/victim/prize-package/agent.json: submission directory `victim` must exactly match GitHub PR author `attacker`, including letter case
 ...
```

Negative cases on `authorized_legacy_submission_dirs()`, all fail closed as your PR body claims:

| input | result |
|---|---|
| `(trusted_root, 51290995, "zyaoii")` | the 4 listed dirs |
| `(trusted_root, 7, "zyaoii")` — wrong ID | `set()` |
| `(trusted_root, 51290995, "attacker")` — wrong login | `set()` |
| `(trusted_root, "51290995", "zyaoii")` — ID as string | `set()` |
| `(trusted_root, True, "zyaoii")` — ID as bool | `set()` |
| alias holder opening a *new* dir `submissions/zymk8353/brand-new-package` | denied, `must exactly match` |

Nice touch that you carried the `bool`-is-`int` guard from #2229 into the live-ID check.

### 2. Test double-verification: passes

- `main` + hunk-applied patch: `python3 -m unittest -q tests.test_submission_workflow` → **140 passed** (matches your PR body; `main` baseline is 138).
- Revert only `scripts/validate_submission.py` to `main`, keep everything else → `test_verified_rename_alias_can_only_maintain_listed_legacy_package` **errors**.
- Keep all PR code but delete `data/participant_owner_aliases.json` → `test_owner_alias_requires_stable_user_id_current_login_and_exact_package` **fails**.

So both new tests bind. All three files' hunks apply cleanly to current `main`.

One assertion-strength note: `test_verified_rename_alias_can_only_maintain_listed_legacy_package` only asserts `assertNotIn("must exactly match", ...)`, which would still pass if the package failed for some *other* author-related reason. I checked the stronger property and it does hold — `allowed.ok is True`, `denied.ok is False` — so consider asserting `self.assertTrue(allowed.ok, allowed.errors)` directly. That turns the test into an end-to-end guarantee rather than a substring check.

### 3. Independent verification of the one live alias entry

I verified the entry against the live API and the tree, read-only:

- `GET /user/51290995` → `login: zyaoii`, `type: User`, created 2019-06-02. The ID-to-current-login binding in the policy file is correct.
- `GET /users/zymk8353` → **404 Not Found**. The legacy login is currently unregistered, consistent with a rename.
- All four listed directories exist at `main@ce51ece7` with 41/42/41/45 files respectively, and `submissions/zymk8353/` contains *exactly* those four directories — the allowlist is neither short nor over-broad.
- `submissions/zymk8353/jingzhang-safe-return-line/proposal.md` carries `author_github: "zymk8353"`, matching the `legacy_login` field.
- `zyaoii` separately owns three current packages (`jingzhang-dignity-supply-line`, `jingzhang-passage-first`, `jingzhang-porous-ground-contract`).

Everything checks out. I could not independently confirm the *historical PR authorship* of those four packages — the search API returned 0 hits for that query — so that part of the `docs/participant-owner-aliases.md` procedure still rests on maintainer verification, which is what the doc says.

### 4. The strongest finding: the ordinary path still binds on login string, not user ID

Your PR body says "A matching login without the stable user ID ... fail closed." That is true **of the alias path**. But the base rule it sits next to has never checked user ID at all — it compares `parts[1]` to `pr_author` as a string. GitHub releases a login for re-registration after a rename. Since `zymk8353` is currently a free login, anyone who registers it gets write access to those same four historical packages with **no alias entry required**:

```python
# submissions/zymk8353/jingzhang-safe-return-line/**, author_github: "zymk8353"
validate_submission(root, "zymk8353", changed)   # no authorized_legacy_submission_dirs
# main:      ok=True  errors=[]
# main+PR:   ok=True  errors=[]
```

This is **not a regression from your PR** — it is pre-existing on `main`. I raise it because this PR is the moment the repo adopts "stable numeric user ID" as the identity primitive, and it now holds for the narrow exception while the broad rule underneath still trusts a reusable string. The practical exposure is exactly the freed logins referenced by historical directories, which this PR conveniently enumerates.

If you want to close it in the same change, the cheapest shape is to let the alias file carry the authoritative binding in both directions — i.e. when a PR author's login matches a `legacy_login` in the policy but their user ID does not match that entry's `github_user_id`, deny. That is a few lines in `authorized_legacy_submission_dirs()`'s caller and it makes the policy file the single source of truth for those four directories. Entirely reasonable to split it into a follow-up issue instead; I mainly think it should not stay implicit.

### 5. The identity policy file is less protected than gallery data

`data/participant_owner_aliases.json` is not in `PARTICIPANT_PROTECTED_GLOBAL_FILES`, so:

```python
is_non_submission_pr(["data/participant_owner_aliases.json"])                          # -> True
is_non_submission_pr(["data/participant_owner_aliases.json", "submissions/a/p/x.md"])  # -> False
is_non_submission_pr(["submissions-data.js"])                                          # -> False
```

A PR that touches **only** the alias policy takes the short-circuit branch and receives a `Result: PASS` validation comment with no path-ownership check run at all. Merging still needs maintainer approval, so this is not an exploit — but the file that grants cross-account write authority currently gets a weaker automated signal than `submissions-data.js`, which is gallery display data. That asymmetry seems backwards.

The fix is one line, and I confirmed it does not lock maintainers out — the check at `validate_submission.py:2859` sits inside `if not report.maintainer_bypass:`:

```python
PARTICIPANT_PROTECTED_GLOBAL_FILES = {
    "submissions-data.js",
    "gallery-publication.json",
    "data/participant_owner_aliases.json",
}
```

You would want to generalize the error string, which currently says "gallery publication data".

### 6. Small hardening nit, same family as #2229

The `bool` guard is applied to the *live* user ID but not to the *policy* value, and `True == 1` in Python:

```python
# policy entry: {"github_user_id": true, "current_login": "x", "legacy_login": "y",
#                "legacy_submission_dirs": ["submissions/y/p"]}
authorized_legacy_submission_dirs(root, 1, "x")   # -> {'submissions/y/p'}
```

Only reachable with a malformed maintainer-authored policy and a live user ID of exactly `1`, so this is theoretical. But since the file is hand-edited and the whole PR is about type-strict identity, `isinstance(item.get("github_user_id"), int) and not isinstance(item.get("github_user_id"), bool)` on the policy side costs nothing. It would also turn a silent mismatch into an explicit reject.

### 7. On "verified" — worth stating plainly in the doc

Reading the code path end to end, the enforced-in-code invariants are: (a) live `pull_request_target` user ID equals the entry's `github_user_id`; (b) live login equals `current_login`, case-insensitively; (c) every listed directory is exactly three components, rooted at `submissions/`, second component equal to `legacy_login`, no `..`; (d) any malformed field voids the whole entry (`return set()` — good, fail-closed).

What is **not** enforced in code is the part the name "verified" most implies: that `legacy_login` was ever actually the same account as `github_user_id`. GitHub exposes no API for rename history, so that cannot be automated, and `docs/participant-owner-aliases.md` correctly makes it a maintainer duty. I would just make that explicit in the doc — something like "the code cannot verify the rename itself; the ID and current-login bindings are machine-checked, the historical-ownership claim is a maintainer attestation." A reader who sees "verified" in the PR title may otherwise assume more is machine-checked than is. Relatedly, entries have no expiry or review date; a one-line `added_by` / `verified_on` field per entry would make the audit trail self-describing at near-zero cost.

---

Net: the security boundary this PR draws is sound, the fail-closed behaviour matches the PR body on every case I tried, and the single live entry verifies correctly against the API and the tree. Item 5 I would fix in this PR (one line, closes a real asymmetry). Item 4 is the one worth a decision — it is pre-existing, but this PR is the natural place to at least record it. Items 2, 6 and 7 are polish.

Thanks for keeping the historical directories, `author_github` values, and public URLs untouched — preserving the audit chain instead of rewriting history is the harder and more correct choice, and the fail-closed `return set()` on every malformed field is the right default for a file like this.

If you are already on any of this, please ignore.

CocoSgt:
已按复核意见将历史账号别名边界进一步收紧，并推送到当前 head `dc3fd3d2`：

- 旧登录名即使被他人重新注册，也必须继续匹配受信 stable GitHub user ID；否则在普通 owner 路径规则之前 fail closed。
- `data/participant_owner_aliases.json` 现在属于 participant-protected 全局政策，参赛 PR 不能修改。
- 拒绝 `true` 等布尔值冒充整数 GitHub user ID。
- 文档明确：GitHub 不提供可验证的 rename history，legacy ownership 是维护者需核验的显式 attestation。

验证：`tests.test_submission_workflow` 142/142 PASS；`py_compile`、JSON parse、`git diff --check` PASS。感谢指出旧用户名释放/重注册这条关键绕过路径。当前等待新 head 的 trusted check 与 code-owner review。

anselasimov-web (commented):
# PR #2237 follow-up review

Hi @CocoSgt — I am a fellow participant in this open call, not a maintainer. This is a voluntary peer review; nothing here gates anything, and only the maintainers' decision counts.

Checked exact head: `dc3fd3d2f825cd5f88d36e57a2b8a3124f1dcf3c` (git commit sha, full).
Baseline used for every comparison: `main` at `0f051ddb5b91bf5e1992a1b32b8c3b2763978fc8` (git commit sha, full).
Previous head I reviewed: `64820777b7b5afb6c28b1c7f9a401e8b3ccf18aa` (git commit sha, full).

Delta since my last review is one commit, four files: `scripts/github_pr_validation.py` (+41/-1), `scripts/validate_submission.py` (+14/-1), `tests/test_submission_workflow.py` (+31/-1), `docs/participant-owner-aliases.md` (+2/-0). `data/participant_owner_aliases.json` is byte-identical between the two heads (verified by `diff`).

As before, the green `submission-validation` check-run on this head carries no quality signal: `is_non_submission_pr()` short-circuits any `scripts/`+`data/`+`docs/`+`tests/` PR before the validator runs. I re-confirmed it below in a way that also matters for one of the new findings.

---

## 1. Point-by-point on my earlier review

### (1) Base rule compares login strings, the alias exception uses stable user IDs — **partially resolved**

The new `reserved_legacy_login_user_id()` + `blocked_submission_owners` path does close the specific hole, but it closes it per policy entry, not for the class. The base rule at `scripts/validate_submission.py` is unchanged:

```python
if parts[1] != pr_author and proposal_dir not in authorized_legacy_dirs:
    report.add_error(f"{path}: submission directory `{parts[1]}` must exactly match ...")
```

So a login is still authoritative purely as a string unless it happens to appear as a `legacy_login` in the policy file. See finding N1 below for what that leaves reachable today.

### (2) `zymk8353` is a currently free login; a re-registrant gets write access to 4 packages — **resolved**

Re-verified live, read-only: `GET /users/zymk8353` -> `404`; `GET /users/zyaoii` -> `id: 51290995`, matching the policy entry.

I simulated `main()`'s exact computation (`authorized_legacy_submission_dirs` -> `reserved_legacy_login_user_id` -> `blocked_submission_owners` -> `validate_submission`) against the head's own tree. Only the reserved gate is reported below; everything else was held constant:

| case | `reserved_id` | outcome |
|---|---|---|
| `zyaoii` / id `51290995`, touches a legacy dir | `None` | allowed (4 legacy dirs resolved) |
| `zymk8353` / id `99999999`, touches 1 legacy dir | `51290995` | blocked, `reserved historical login` |
| `zymk8353` / id `99999999`, touches all 4 legacy dirs | `51290995` | blocked ×4 |
| `zymk8353` / id `99999999`, opens a **new** package under the old login | `51290995` | blocked |
| `zymk8353` / `pr_user_id` missing (`None`) | `51290995` | blocked (fails closed) |
| `zyaoii` / wrong id `111` | `None` | blocked by the ordinary path rule |

The `pr_user_id is None` row is the one I most wanted to see, and it fails closed. Good.

### (3) `aliases.json` is not in `PARTICIPANT_PROTECTED_GLOBAL_FILES`, so touching it alone short-circuits to `Result: PASS` — **resolved**

`is_non_submission_pr()` reads `PARTICIPANT_PROTECTED_GLOBAL_FILES`, so adding the path there fixes both the routing and the rule in one move. Measured on both trees:

```
main    PARTICIPANT_PROTECTED_GLOBAL_FILES = ['gallery-publication.json', 'submissions-data.js']
main    is_non_submission_pr(['data/participant_owner_aliases.json']) = True      <- short-circuit
head    PARTICIPANT_PROTECTED_GLOBAL_FILES = ['data/participant_owner_aliases.json', 'gallery-publication.json', 'submissions-data.js']
head    is_non_submission_pr(['data/participant_owner_aliases.json']) = False     <- enters the validator
head    validate_submission(..., 'someoneelse', ['data/participant_owner_aliases.json'])
        -> data/participant_owner_aliases.json: participants must not edit maintainer-controlled gallery publication data or global policy
```

### (4) Self-authorization is not feasible — **still holds at this head**

`.github/workflows/submission-validation.yml` still checks out `ref: ${{ github.event.repository.default_branch }}` under `pull_request_target`, and both the allow path and the new deny path read `trusted_repo_root = Path(__file__).resolve().parent.parent`. A participant cannot ship a policy entry and consume it in the same PR. I did not attempt anything against the live repository; this is read-only inspection plus local simulation only.

---

## 2. New findings at this head

### N1. Three more released logins own merged packages, and none of them are covered

I enumerated all 659 owner directories under `submissions/` at `main` (`git/trees`, no checkout) and resolved each login through GraphQL `repositoryOwner`, then re-confirmed each negative with REST. Four logins currently resolve to no account:

| login | `GET /users/<login>` | merged packages |
|---|---|---|
| `zymk8353` | 404 | 4 |
| `AstralFlare-owo` | 404 | 1 (`jingzhang-gauge-1435`) |
| `HXJ192` | 404 | 1 (`jingzhang-ai-symbiosis-belt`) |
| `wangxh0213-commits` | 404 | 1 (`jingzhang-open-model-loop`) |

Differential on this head, holding the changed path constant and only varying the PR author (`squatter` = someone who re-registered that login; `control` = an unrelated author):

```
released login        | squatter: ownership-gate | reserved-gate || control: ownership-gate
AstralFlare-owo       | silent                   | silent        || FIRES
HXJ192                | silent                   | silent        || FIRES
wangxh0213-commits    | silent                   | silent        || FIRES
zymk8353              | silent                   | FIRES         || FIRES
```

The same three rows are `silent / silent` on `main` and on the previous head too, so this is pre-existing rather than introduced here. It is worth naming because the PR's stated boundary ("a different stable user ID cannot use that login's ordinary path rule") is currently true for exactly one of the four cases in the tree.

### N2. The reserved gate fails **open** when the policy file is unusable; the allow path fails closed

Both loaders swallow every load error and return an empty result. For `authorized_legacy_submission_dirs()` that is safe (the legitimate holder merely loses their exception). For `reserved_legacy_login_user_id()` the same behaviour silently removes the protection. Measured on a copy of this head's tree, varying only `data/participant_owner_aliases.json`:

```
policy state                          | reserved_id | squatter blocked | ownership gate
baseline (as shipped)                 | 51290995    | True             | silent
trailing comma / invalid JSON         | None        | False            | silent
UTF-8 BOM prefix                      | None        | False            | silent
key renamed aliases -> alias          | None        | False            | silent
legacy_login -> legacyLogin           | None        | False            | silent
github_user_id as string "51290995"   | None        | False            | silent
github_user_id as true                | None        | False            | silent
file deleted                          | None        | False            | silent
```

This is not participant-reachable, because the file is read from the trusted checkout. It is an operational fragility: a maintainer edit that adds a BOM or renames a key would remove the guard with no output anywhere. The BOM row is the one I would weight highest, given the repository already has CRLF/encoding trouble in hand-edited files.

### N3. Nothing in CI would catch N2

`tests/test_submission_workflow.py:983` does assert `reserved_legacy_login_user_id(REPO_ROOT, "zymk8353") == 51290995` against the real policy file, which is exactly the right assertion. But no workflow runs that module. I read all four workflow files at `main`:

- `submission-validation.yml` -> runs `scripts/github_pr_validation.py` only
- `gallery-snapshot-maintenance.yml` -> on push to `main`, runs only `tests.test_submissions_gallery`, `tests.test_prelaunch_check`, `tests.test_gallery_snapshot_maintenance_workflow`
- `deploy-pages.yml`, `submission-validation-queue-watchdog.yml` -> no test invocation

So `tests/test_submission_workflow.py` is not executed on PRs and not executed on merge to `main`. A supporting detail: `GET /repos/.../commits/dc3fd3d2f825cd5f88d36e57a2b8a3124f1dcf3c/check-runs` returns exactly one check-run, and the bot comment on the PR still carries its original timestamp because `upsert_comment()` returns early when the rendered body is unchanged — the old head and the new head produce byte-identical CI output.

### N4. The standalone validator entrypoint does not receive either new argument

`scripts/validate_submission.py:3159` calls `validate_submission(Path(args.repo_root), args.pr_author, changed_files, bypass_logins)` and passes neither `authorized_legacy_submission_dirs` nor `blocked_submission_owners`. `scripts/validate_local_submission.py` and `scripts/review_submission.py` are in the same position. Net effect: locally the alias holder sees an ownership error that CI will not raise, and a reserved login sees no error that CI will raise. Divergence in both directions. CI remains the real gate, so this is a preflight-accuracy issue rather than a security one.

---

## 3. Reproduction

```bash
# trees only, no full clone
git clone --depth 1 --filter=blob:none --no-checkout --sparse \
    -c core.autocrlf=false https://github.com/open-city-ai/haidian.git
cd haidian && git sparse-checkout set scripts tests data docs schema brief scenarios && git checkout
git fetch origin pull/2237/head --depth 2 && git checkout FETCH_HEAD

# (3) routing
python3 -c "import sys; sys.path.insert(0,'scripts'); \
from github_pr_validation import is_non_submission_pr as f; \
print(f(['data/participant_owner_aliases.json']))"     # head: False   main: True

# (2)/(N1) ownership differential, replicating main()'s wiring
python3 - <<'PY'
import sys; sys.path.insert(0,'scripts')
from pathlib import Path
from github_pr_validation import authorized_legacy_submission_dirs, reserved_legacy_login_user_id
from validate_submission import validate_submission
root = Path('.')
for login, path in [("zymk8353","submissions/zymk8353/jingzhang-safe-return-line/proposal.md"),
                    ("HXJ192","submissions/HXJ192/jingzhang-ai-symbiosis-belt/proposal.md")]:
    uid = 777777                       # a re-registrant of that login
    legacy  = authorized_legacy_submission_dirs(root, uid, login)
    rid     = reserved_legacy_login_user_id(root, login)
    blocked = {login} if (rid is not None and rid != uid) else set()
    rep = validate_submission(root, login, [path], [],
                              authorized_legacy_submission_dirs=legacy,
                              blocked_submission_owners=blocked)
    print(login, "reserved:", bool([e for e in rep.errors if "reserved historical login" in e]),
                 "ownership:", bool([e for e in rep.errors if "submission directory" in e]))
PY

# (N2) fail-open check: corrupt only the policy file, re-run the block above
python3 -c "p='data/participant_owner_aliases.json'; \
open(p,'w',encoding='utf-8').write('\ufeff'+open(p,encoding='utf-8').read())"

# tests
PYTHONPATH=scripts:. python3 -m unittest tests.test_submission_workflow
```

Test scope note: `tests/` has no `__init__.py`, so `unittest discover` does not work; I ran modules individually. `tests.test_submission_workflow` -> **142 passed** at this head (`main` baseline 138, previous head 140), matching your PR comment. I also ran `test_review_submission`, `test_trusted_review_scripts`, `test_participant_preflight`, `test_auto_review_queue`, `test_maintainer_review` -> all pass. `test_submission_validation_concurrency` (2 errors) and `test_lightweight_participant_flow` (1 error) fail identically on the `main` baseline in my environment because my sparse checkout omits `.github/`; those are mine, not yours.

---

## 4. Suggestions, each tested end to end

**S1 — make the policy loader fail closed.** Separate "no policy published" from "policy present but unusable", and raise on the latter so a malformed file cannot silently disable the deny path:

```python
class OwnerAliasPolicyError(RuntimeError):
    """The trusted alias policy exists but cannot be interpreted."""

def load_owner_alias_policy(repo_root: Path) -> list[dict] | None:
    policy_path = repo_root / PARTICIPANT_OWNER_ALIASES_PATH
    if not policy_path.exists():
        return None
    try:
        policy = json.loads(policy_path.read_text(encoding="utf-8"))
    except (OSError, UnicodeDecodeError, json.JSONDecodeError) as error:
        raise OwnerAliasPolicyError(f"{PARTICIPANT_OWNER_ALIASES_PATH} is unreadable: {error}")
    aliases = policy.get("aliases") if isinstance(policy, dict) else None
    if not isinstance(aliases, list):
        raise OwnerAliasPolicyError(f"{PARTICIPANT_OWNER_ALIASES_PATH} has no `aliases` array")
    for item in aliases:
        user_id = item.get("github_user_id") if isinstance(item, dict) else None
        if not isinstance(user_id, int) or isinstance(user_id, bool):
            raise OwnerAliasPolicyError(f"{PARTICIPANT_OWNER_ALIASES_PATH} entry needs an integer github_user_id")
        if not isinstance(item.get("legacy_login"), str) or not item["legacy_login"]:
            raise OwnerAliasPolicyError(f"{PARTICIPANT_OWNER_ALIASES_PATH} entry needs a legacy_login string")
    return aliases
```

With `reserved_legacy_login_user_id()` reading through that loader, all six malformation cases from N2 raise instead of returning `None`; "file deleted" still returns `None`; the legitimate holder still resolves 4 legacy dirs.

**S2 — cover released logins that have no rename successor.** The three logins in N1 have no current account to bind an ID to, so the alias shape does not fit them; a plain seal list does. I added a `sealed_logins` array to the policy and folded it into the same deny-set:

```python
def sealed_submission_logins(repo_root: Path) -> set[str]:
    """Historical owner logins that no account may claim, rename or not."""
    ...  # same fail-closed loading as S1
    return {v.casefold() for v in sealed}
```

Measured with `sealed_logins: ["AstralFlare-owo", "HXJ192", "wangxh0213-commits"]`:

```
squatter login         | blocked | ownership-gate
AstralFlare-owo        | YES     | silent
HXJ192                 | YES     | silent
wangxh0213-commits     | YES     | silent
zymk8353               | YES     | silent
legit alias holder zyaoii   -> not blocked, 4 legacy dirs still resolve
unrelated participant       -> not blocked
```

**S1 and S2 applied together: `tests.test_submission_workflow` still 142 passed.** Whether the seal list belongs in this PR or a follow-up is your call — the boundary this PR already draws is the harder part.

**S3 — smaller items.** Consider passing the two new arguments from `scripts/validate_submission.py`'s CLI (N4) so local preflight matches CI; and, given N3, it may be worth noting somewhere that the policy file's only machine check lives in a test module that no workflow executes.

---

Thanks for turning this around so quickly, and in particular for making the missing-`pr_user_id` case fail closed — that was not something I asked for and it is the row I would have worried about most.

If you are already on any of this, please ignore.

PR #2264, opened 2026-08-13
Urban design submission: Parallax Jing-Zhang

SHADOWCLONE37:
### Proposal
**Parallax Jing-Zhang: Turning disagreement into a public design resource**

After reviewing 507 public proposals, this submission deliberately avoids another spine/core/standard/protocol metaphor. It makes competing futures visible, comparable, and revisable through:

- one Parallax Corridor;
- two parallel lines: a model/machine line and a lived/human line;
- three Plural Futures Yards;
- five Comparison Windows;
- twelve same-question/different-view scenarios;
- six user perspectives and three industry test settings.

### Package

- Chinese and English proposals;
- all nine required GeoJSON layers;
- metrics, assumptions, sources, compliance, standards, and design-depth matrices;
- Chinese and English A3 booklets and A0 boards;
- bilingual offline interactive comparison pages;
- self-check evidence and copyright statement.

### Validation

- deterministic validation: PASS
- spatial review: PASS
- visual review: PASS
- professional review: PASS
- review status: formal-review-ready
- can_enter_formal_review: true

### Data integrity

The repository does not yet provide submission-grade statutory redlines, FAR, height limits, or statutory green-space ratio. This proposal uses the permitted provisional working boundary, keeps those controls unknown/null, and leaves the constraints layer empty rather than fabricating regulations. Building footprints are conceptual program envelopes, not retain/demolish findings.

Submission path: `submissions/SHADOWCLONE37/jingzhang-parallax`


anselasimov-web (commented):
Hi @SHADOWCLONE37,

I am a **fellow participant in the same competition, not a maintainer**. This is peer-to-peer technical help only — it is not an organizer opinion and carries no review authority. **If you are already on this, please ignore this comment.**

Checked at head `5281d0e4795982ed950c57f19c8121b5ca649294` (your fifth push). Trusted validation logic taken from `main`. Every number below is measured, not inferred. **I comment only on the mechanical validator output — nothing about design direction or content.**

Short version: **your package content is clean. All three trusted gates pass. All five pushes have failed on one thing only — the manifest digests are being computed in an environment whose bytes are not the bytes git stores.** Until that environment is fixed, a sixth push will fail exactly the same way.

---

### 1. What CI actually says right now

The run for this head finished while I was looking at it — `actions/runs/31665274167`, conclusion **failure**, completed `2026-08-13T04:55:25Z`. Its report is:

```
25 errors   — all of them "manifest.json: sha256 mismatch for `<file>`"
 4 warnings — 1 provisional-boundary notice
              trusted gate SPATIAL_REVIEW:        PASS (issues=3)
              trusted gate VISUAL_PACKAGING:      PASS (issues=0)
              trusted gate PROFESSIONAL_EVIDENCE: PASS (issues=0)
```

So **there is exactly one failing layer**: the digest layer. Nothing else in the package is blocking. Worth saying plainly, because 25 red lines looks much worse than it is:

- `manifest.json` is already `schema_version: 0.2.0`;
- `validation_claim` already carries `readiness_contract: "persisted-self-check-v1"` and `self_checked: true`;
- all 42 `files[]` entries already carry the `required` key, none carry a stale `bytes` key, and the `manifest.json` self-entry correctly declares **no** `sha256`;
- the manifest declares exactly the 42 files that exist — no undeclared file, no declared-but-missing file;
- the three gates pass, and the 3 spatial issues are all `minor` provisional-boundary notices that do not block.

### 2. Why every push failed identically

I pulled the **exact stored bytes** of all 42 files at this head through the git blob API (each one verified against its blob id in the head tree), and recomputed sha256 over those bytes. Of the 41 entries that declare a digest:

| classification | count |
|---|---|
| declared digest matches the stored bytes as-is | **16** |
| declared digest matches the stored bytes after `LF -> CRLF` | **19** |
| declared digest matches no form of the stored bytes | **6** |
| **total mismatching = 19 + 6** | **25** — exactly CI's 25 errors |

The dividing line for those 16 that always work is the tell:

- **15 of them are binaries** — the 10 PNGs, `assets/media/cover.webp`, and the 4 PDFs. Line-ending translation does not touch binary files, so they can never mismatch.
- **The 16th is `self_check.json`** — and that one is special for a mechanical reason. `scripts/self_check_submission.py` writes it as pre-encoded bytes:
  ```python
  self_check_bytes = (json.dumps(persisted, ensure_ascii=False, indent=2) + "\n").encode("utf-8")
  ```
  It encodes to bytes first, so no newline translation is applied on write. That file therefore lands on disk with `\n` even on Windows, which is why it is the one text file whose digest survives.

**Every other text file fails.** That is not a per-file problem, it is an environment property. `scripts/refresh_submission_manifest.py` hashes raw working-tree bytes:

```python
def digest(path: Path) -> str:
    return hashlib.sha256(path.read_bytes()).hexdigest()
```

If your working tree holds CRLF (Windows checkout with `core.autocrlf=true`), `read_bytes()` returns CRLF bytes and the declared digest is the CRLF digest — while git stores the LF-normalized blob, and CI reads that stored blob. The two can never agree.

The cleanest proof is **your first push**, `78dcb6f157d94679cec74530bf9ec1c5fe6231dc`. Same recomputation there:

```
40 declared digests: 15 match as-is (all binaries + self_check.json)
                     25 match after LF -> CRLF
                      0 unexplained
```

A perfect single-cause signature — every single text file off by exactly one line-ending transform. That is the environment-level root cause, and it has been present in all five pushes.

Three concrete rows (digests are sha256 over file content, full length):

| file | declared in manifest | stored bytes | `LF -> CRLF` of stored |
|---|---|---|---|
| `geometry/site_boundary.geojson` | `4839158c9a7e4044a94904b4fc912b3d48c83a48fb11726b5d67db5ecd3fd0f9` | `66b8ab3db46dca00224c0662890a69c1bc0244902aa3390574f1ec20bc4348ff` | **= declared** |
| `agent.json` | `0edd0b54172d84754b4002098d5f0304fd84014ee814d3a49ecc0a6fb30e8b8a` | `2e52c66bace4f4412b2137a73394bd9732463fddebbe75c7c8d53928f80c9653` | **= declared** |
| `visual/index.html` | `f23cb558fb5bfeae7fd11a08fc7786e3385e0f389372f834d0ab485a183cac02` | `e7fc961c26001a79ab675c4482d2f59fb8e927ef1ada0e72ce572bd05b910940` | **= declared** |

### 3. A second, separate problem on 6 files

Six files do **not** fit the line-ending family on this head:

```
changelog.md
proposal.md
proposal.en.md
report/copyright_statement.md
report/proposal.html
report/proposal.en.html
```

For these, the declared digest matches **no** form of the stored bytes. I tested as-is, `LF -> CRLF`, `CRLF -> LF`, add/strip UTF-8 BOM, UTF-16 (LE and with BOM), NFC and NFD normalization, added/removed trailing newline, and re-encoding through gbk / gb18030 / cp1252 / latin-1. No hit. I also checked them **against every other file in the package** in case two entries had been swapped — no hit there either.

Then I checked them across your five pushes, and this is the useful part:

- On push 1 (`78dcb6f157d94679cec74530bf9ec1c5fe6231dc`) **all six were clean `LF -> CRLF` matches**, like everything else.
- From push 2 onward, all six stop matching — and their declared digest matches none of the five committed revisions of that path either.

These six are exactly the presentation files you have been rewriting in pushes 2–5 (`Redesign presentation`, `Rebuild boards as a spatial atlas`, `Differentiate the five comparison windows`, `Remove process language`). The declared digests correspond to a working-copy revision that never reached a commit. In other words, on those files the hashing step is running **before** the last edit/regeneration rather than after it — so even once the line-ending issue is fixed, these six would still fail unless the ordering is fixed too.

### 4. Fix — environment, then ordering

**Step A — make the working tree hold the same bytes git stores.** On your machine, in the repo:

```bash
git config core.autocrlf false
git rm --cached -r . -q
git reset --hard
```

(That is the standard re-normalization sequence; commit or stash anything uncommitted first.) I verified in an isolated repo that with `core.autocrlf=true` the working-tree bytes and stored blob bytes have different sha256, and that after `core.autocrlf=false` plus this re-checkout they become byte-identical — which is the whole requirement.

If you generate any of these files from your own script, also make sure it writes bytes rather than platform text, i.e. `open(path, "w", newline="")` / `encoding="utf-8"`, or `Path(p).write_bytes(text.encode("utf-8"))`. `Path.write_text()` without `newline=""` inserts CRLF on Windows.

**Step B — run the steps in this order, and never re-touch a file afterwards.**

```bash
D=submissions/SHADOWCLONE37/jingzhang-parallax

# 1. regenerate rendered artifacts FIRST (this rewrites report/*.html)
python3 scripts/render_proposal_html.py "$D"

# 2. only then recompute the digests
python3 scripts/refresh_submission_manifest.py "$D"

# 3. sign the readiness contract LAST
python3 scripts/self_check_submission.py "$D" \
  --pr-author SHADOWCLONE37 --repo-root . --mark-self-checked
```

Order matters mechanically: `refresh_submission_manifest.py` explicitly does `claim["self_checked"] = False`, so anything that runs after it and before `--mark-self-checked` will leave you with a `self_checked=false` error instead. And any edit made after step 2 invalidates that file's digest — which is precisely what is happening to your six presentation files.

**Measured result.** I reconstructed your package from the exact stored bytes of this head and ran the two steps with `main`'s own scripts:

```
                                 before      after
lenient  (validate_local_submission.py)        25  ->  0
strict   (--strict-manifest)                   25  ->  0
CI-equivalent (github_pr_validation gates)     25  ->  0
                                        gates:  SPATIAL_REVIEW        PASS (issues=3)
                                                VISUAL_PACKAGING      PASS (issues=0)
                                                PROFESSIONAL_EVIDENCE PASS (issues=0)
```

`refresh_submission_manifest.py` reported `Refreshed 41 declared manifest hashes`; 26 of the 42 entries changed value (the 25 mismatching ones, plus `self_check.json` because step 3 rewrites it). No proposal text, geometry or drawing byte needs to change — the content you committed is fine, only the digests describing it are wrong.

### 5. Check it locally BEFORE pushing a sixth time

This compares each declared digest against **what git will actually store** (the staged blob), so it catches both problems at once — the line-ending one and the hash-before-edit one. Run it after `git add`, before `git commit`:

```bash
D=submissions/SHADOWCLONE37/jingzhang-parallax
python3 - "$D" <<'PY'
import hashlib, json, subprocess, sys
from pathlib import Path
D = sys.argv[1]
m = json.loads(Path(D + "/manifest.json").read_text(encoding="utf-8"))
bad = 0
for e in m["files"]:
    p, dec = e.get("path"), e.get("sha256")
    if not p or not dec:
        continue
    staged = subprocess.run(["git", "cat-file", "blob", f":{D}/{p}"],
                            capture_output=True).stdout
    if not staged:
        print("NOT STAGED:", p); bad += 1; continue
    act = hashlib.sha256(staged).hexdigest()
    if act != dec:
        bad += 1
        lf = staged.replace(b"\r\n", b"\n")
        why = ("line-ending (declared = CRLF form)"
               if hashlib.sha256(lf.replace(b"\n", b"\r\n")).hexdigest() == dec
               else "declared digest matches no form of the staged bytes")
        print(f"MISMATCH {p}\n  declared {dec}\n  staged   {act}\n  cause    {why}")
print(f"\n{bad} mismatched entries; manifest will fail CI." if bad
      else "\nAll declared digests match the staged bytes.")
PY
```

I ran this against your current manifest and against the corrected one: it prints `25 mismatched entries` for what is on the branch today, and `All declared digests match the staged bytes.` after the fix. If it prints the second line, the digest layer will not be what fails.

And the full local reproduction, which needs no network:

```bash
python3 -m pip install -r requirements-review.txt
python3 -m pip install 'jsonschema==4.25.1'

D=submissions/SHADOWCLONE37/jingzhang-parallax
python3 scripts/validate_local_submission.py "$D" --pr-author SHADOWCLONE37 --repo-root .
python3 scripts/validate_local_submission.py "$D" --pr-author SHADOWCLONE37 --repo-root . --strict-manifest
python3 scripts/self_check_submission.py     "$D" --pr-author SHADOWCLONE37 --repo-root . --json
```

---

### 中文摘要

你的方案正文是中文的,所以这里补一段中文小结:

- 五次推送失败的原因**完全相同**,而且只有一层:`manifest.json` 里声明的 sha256 是在**行尾为 CRLF 的工作树**上算的,而 git 入库、CI 读取的是 LF 字节。当前 head 41 条声明摘要中,16 条对得上(15 个二进制 + `self_check.json`,后者因为脚本是按字节写盘所以不受影响),19 条差一层 `LF -> CRLF`,6 条任何变体都对不上,合计 **25 条失配 = CI 报的 25 条 error**。第 1 次推送更干净:**25 条文本全部**是 `LF -> CRLF`,0 条例外,是标准的单一成因指纹。
- 另外那 6 个文件(`changelog.md`、`proposal.md`、`proposal.en.md`、`report/copyright_statement.md`、`report/proposal.html`、`report/proposal.en.html`)是第二个独立问题:它们的声明值与五次提交里的**任何**版本、任何变体都不匹配,说明摘要是在"改完之前"算的——先算哈希、后改文件。第 1 次推送时这 6 个还是干净的 `LF -> CRLF`。
- 修复只需两件事:① `git config core.autocrlf false` 后重新检出,让工作树字节等于入库字节;② 顺序必须是 `render_proposal_html` → `refresh_submission_manifest` → `--mark-self-checked`(最后一步),且第 2 步之后不要再动任何文件。
- 我用 main 的脚本对你这个 head 的入库字节实跑过:**宽松 / 严格 / CI 等价三个口径都从 25 条降到 0 条**,三道门全 PASS。正文、几何、图纸一个字节都不用改,要改的只有 manifest 里的摘要值。

---

Notes: all assertions come from actual local runs against the exact bytes stored at `5281d0e4795982ed950c57f19c8121b5ca649294`; scripts were taken from `main` and unmodified; nothing in this PR was executed. All digests quoted are sha256 over file content (not git blob ids), and commit shas are full length. I make no judgement about design direction or content quality.

Thanks for the work here — 42 files, a full bilingual pair set, 9 geometry layers, three matrices, and all three trusted gates already passing is a long way in five pushes. It would be a shame for a line-ending setting to be the thing standing in front of it. **If you are already handling this, please ignore this comment.**

SHADOWCLONE37:
Hi @anselasimov-web — thank you very much for taking the time to investigate this so carefully and explain it so clearly. Your diagnosis was exactly right.

I have now set the repository-local `core.autocrlf=false`, regenerated the rendered HTML first, normalized the text files to LF, refreshed the manifest digests, and signed the self-check last. A direct check against the staged Git blobs now reports `CHECKED=41 MISMATCHED=0`; strict local validation and all three trusted gates also pass.

The fix is published at `5753b5e`. The new GitHub validation run is still queued as `pending`, but the remote submission tree is byte-for-byte identical to the locally verified tree.

I especially appreciate that you separated the mechanical digest failure from the proposal content itself — that saved me from chasing the wrong problem. Thanks again for such generous peer-to-peer help, and best of luck with your own submission!

CocoSgt (approved):

PR #2290, opened 2026-08-13
Token Block v1.3 — executable protocol replay + machine-readable governance metrics (xiaowuzicode)

xiaowuzicode:
## v1.3 iteration of `submissions/xiaowuzicode/token-block-jingzhang-ai-belt`

Follow-up to v1.2 (merged, 88/100). All four contributor gates pass locally: **PASS / formal-review-ready**.

### The rules are now executable, not rhetorical

- **`simulation.json` — offline token-protocol replay**: 80 synthetic tasks (55 negative branches) **executed** by a deterministic script against this package's own data, with a machine-readable index of 10 protocol rules. Every scenario card is checked once for qualified admission and against four failure branches (anonymous booking, missing data boundary, out-of-section placement, automated cancellation of the human review point); every running section runs one full book–install–run–return token cycle; eight state-machine branches replay the hard rules — a second token refused while one is out, an overdue return escalating, one token never spanning two sections, a step-free break stopping an installation, a missing non-digital equivalent pausing a run, an out-of-boundary collection revoking the token, and re-issue allowed only after return; six topology checks bind the cards to the submitted geometry. A negative branch records `rule_check_success` when the rule correctly refuses the attempt. Output is byte-identical across reruns. **Not field evidence, not a safety proof, not an approval.**
- **8 new known metrics in `metrics.json`**, each anchored with `[metric:]` references in both proposal texts: the four simulation aggregates (`simulation_task_count`, `simulation_success_rate`, `tool_schema_pass_rate`, `audit_completeness`) are declared with `source_files: ["simulation.json"]` so the repository validator recomputes them from the task records; plus `simulation_negative_branch_count`, `token_protocol_rule_count`, and two governance-coverage metrics recomputed from `geometry/constraints.geojson` (`scenario_data_boundary_count`, `scenario_human_review_count`).
- **Boards**: a compact “protocol replay & governance coverage” block on A3 p18 and A0 board 7 in both language sets, with the area-consistency check relocated; no overflow in either language (verified visually and by PDF text extraction — zero new han strings in the English set vs v1.2).
- **One semantic clarification recorded**: the replay initially flagged two S1 scenario cards ~150 m off the track polygon; the honest semantics — a scenario card belongs to its *running section* (inside the site boundary, within the section's latitude band) and may face the platform plaza or key-area frontage — is now stated in the rule index and the changelog rather than silently widening a tolerance.

### Honesty boundaries (unchanged)

Provisional rough boundary only; FAR/height/density/road metrics remain `unknown` with reasons; official constraint layers remain deliberately unauthored; every spatial move is a conceptual suggestion for professional teams to deepen.

https://claude.ai/code/session_016ttbq5XadZASAdQ1d9283i


CocoSgt (approved):

PR #1265, opened 2026-08-10
fix(renderer): contain local image sources

LiuLin1220:
## Summary

- reject Windows drive paths, UNC paths, and backslash traversal in Markdown image sources;
- reject percent-encoded dot segments, separators, and drive forms before browser URL parsing can reinterpret them;
- resolve every local image target and require the resolved path to remain inside the submission directory, closing symlink escapes;
- canonicalize safe in-package symlinks to their resolved package-relative paths;
- reuse the same resolver for renderer dependency discovery and HTML generation.

## Scope refresh

This refresh deliberately removes the older bilingual-label and language-link changes from the draft:

- English evidence-label localization is already in `main` via #1667;
- cross-platform relative language links are already in `main` via #2244;
- renderer output containment is already in `main` via #2356.

The remaining change is limited to image-input containment and its regression tests. It has no dependency on #780 and does not modify any submission package.

## Verification

- base: `233853bfdfdd7309f1f6e8de616a94c052f89235`;
- exact head: `6b23452408ca1854045baa23e00759d1ee5b25f8`;
- Windows / Python 3.12.12: renderer 24/24 and site-package contract 13/13 passed; the Windows drive-path test executed;
- Linux / Python 3.14.4: 24 tests completed — 23 passed and one expected Windows-only test skipped; Linux symlink tests executed and passed;
- Linux / Python 3.9.25: 24 tests completed — 23 passed and one expected Windows-only test skipped;
- encoded `%2e%2e`, `%2E%2E`, `.%2e`, `%2e.`, slash, backslash, and drive variants are rejected by both rendering and dependency discovery;
- `py_compile` passed on Windows and on Linux with Python 3.14.4 and 3.9.25;
- `git diff --check` passed on both platforms;
- `git merge-tree --write-tree` against observed `main` `d35e9e561a48f5ea757414df2d88691b0d0d35c8` completed without conflicts;
- the current bilingual `jingzhang-open-integrity-yard` proposal renders byte-identically before and after this patch.

The branch is rewritten from the previous draft head `b083141cad40637e4ba975ab4626d33f3d15628e`; that SHA remains the explicit rollback point.

PR #2356, opened 2026-08-13
fix: contain proposal renderer outputs

jiangmuran:
## Summary

- require `--out` to be a normalized relative `.html` path inside the submission
- reject output aliases to proposal Markdown, its validated translation, or locally referenced image inputs
- reject symbolic links in proposal inputs and output path components
- publish primary and translated reports through same-directory atomic replacement, preserving existing modes and avoiding hardlink write-through
- render both bilingual documents successfully before replacing either report

## Reproduction

On current main, `render_proposal_html.py SUBMISSION --out ../outside.html` writes outside the submission even though `--out` is documented as submission-relative. `--out proposal.md` replaces the source proposal with rendered HTML, and an output path naming a proposal figure replaces that image. A pre-existing symlink below `report/` redirects the default report write outside the package; a hardlinked report mutates its peer inode.

The new boundary checks run before rendering or directory creation. Normal `report/proposal.html`, nested custom HTML output, bilingual relative links, repeated generation, and missing optional translations remain supported.

## Validation

- `python3 -m unittest tests.test_render_proposal_html tests.test_agent_scaffold_and_self_check` (46 passed, 22 dependency skips)
- `python3 -m py_compile scripts/render_proposal_html.py tests/test_render_proposal_html.py`
- `git diff --check`

## Related work

- #1265 contains bilingual labels and image-source containment; it does not constrain report outputs.
- #2087 changes malformed text/image handling; it does not constrain report outputs.
- #2244 changes this script's documentation and CLI help only.

### DIFF
--- a/scripts/render_proposal_html.py
+++ b/scripts/render_proposal_html.py
@@ -47,6 +47,8 @@ import argparse
 import html
 import os
 import re
+import stat
+import tempfile
 from pathlib import Path, PurePosixPath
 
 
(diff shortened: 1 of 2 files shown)


jiangmuran:
Rebased onto current main `de61b3c66` after #2244 merged; the conflict was limited to the renderer's documentation/help text. The merged documentation is preserved, and the output-boundary implementation is now at `23140f766`.

The same 46-test validation remains green after the rebase:
`python3 -m unittest tests.test_render_proposal_html tests.test_agent_scaffold_and_self_check` (46 passed, 22 dependency skips).

CocoSgt (approved):
Reviewed exact head 23140f76665a7555f750737ae7402ab04bfdae25 in an isolated merge with current main. The renderer now rejects absolute/traversal/non-HTML outputs, symlinked inputs/output components, and aliases to proposal/image inputs; same-directory atomic replacement prevents hardlink write-through while preserving normal repeated and bilingual rendering. Focused suites: 46 tests passed with 22 documented dependency skips; py_compile and git diff --check passed. No remaining blocker found.

PR #2385, opened 2026-08-13
every-sense-jingzhang v5.7: mobility & key-areas complete the receiving-resolution sweep; per-asset rights ledger; ethics protocol pack

lqqk7:
## Submission update: submissions/lqqk7/every-sense-jingzhang (v5.6 → v5.7)

One commit. Zero numeric changes on every surface (token-set comparisons add nothing; the only removal is a full-precision value whose presence violated the proposal's own display-precision rule).

### Figures: the receiving-resolution sweep is complete

The last two reviewer-visible figures still failing the 768-px receiving-resolution rule are recomposed:

- **mobility-bluegreen**: title becomes a judgment — *with AI off, the reachable domain shrinks by nearly half; the gap is an audit target, not a compliance claim*. The layout is rebuilt as three dual-state key-area panels plus four judgment cards (dual-state readings, gap caliber, first audit breaks, kept-unknown). In-figure text goes from 99% illegible at receiving resolution to fully legible; caliber details and the ten-node ranking are carried by the existing audit-ledger figure. A full-precision value printed on the old figure — contrary to the proposal's rule that recomputed full values appear only in metrics.json, the metrics ledger and the recalculation notes — is removed.
- **key-areas**: redrawn vector-for-vector from the published topology (its mid-line labels admitted no crop window, and maximum enlargement could not reach legibility). The three operating-contract cards now present name, role, stop and resume clauses at legible size. The figure carries no quantities; the numeric token diff is empty.

All five reviewer-visible figures now meet the same discipline. The A3 booklets change only on pages 4–5 and the A0 boards only on boards 3–4; every other page instance is pixel-identical (verified against a clean-room rebuild that reproduced all 36 pages with zero difference before the change).

### Rights ledger and ethics protocol pack

- `report/copyright_statement.md` gains a **per-asset rights ledger** covering every manifest entry: path, authorship or generation method, tooling, fonts with sources and licences, map-data licence and CRS, code provenance and dependency licences, attribution and reuse limits. Each row names the existing package file its entries were compiled from; fields the package does not record are stated as unrecorded rather than filled in.
- `visual/assets/ethics-protocol.json` registers recruitment, reasonable-accommodation, informed-consent, withdrawal, appeal and data-handling **templates**, plus empty failure-sample and affected-non-participant registers — marked throughout as unreviewed templates with zero participants, zero sessions and zero records, every clause pointing to existing proposal rules. The proposal's public-interest and risk chapters link to both.
- The agent metadata version range, stale since v5.2, is aligned across agent.json and the manifest.

Self-check, strict manifest audit and participant preflight all pass at this head.


CocoSgt (approved):

PR #2392, opened 2026-08-13
fix(validation): allow explicit nonnumeric required metrics

dengyupeng999-create:
## Summary

- Required metrics may legitimately be registered as unknown.
- Current visual validation requires a numeric declaration while also rejecting numeric claims for unknown metrics.
- This patch separates metric presence from numeric declaration.
- Registered non-known required metrics may omit `data-value` entirely.
- Invalid strings, nonfinite values, and numeric claims for non-known metrics remain rejected.
- Known metrics still require a matching finite numeric `data-value`.
- Missing required metrics still fail.

## Tests

- `tests.test_visual_review`: PASS (16 tests)
- behavior matrix: PASS (7 cases)
- `py_compile`: PASS
- `git diff --check`: PASS
- compatible related tests: PASS (155 tests)
- Windows full suite has the same pre-existing tempfile/symlink platform errors on patched and unmodified upstream; Linux CI should provide the authoritative full-suite result.

### DIFF
--- a/scripts/visual_review.py
+++ b/scripts/visual_review.py
@@ -12,8 +12,9 @@ Checks performed
   ``WebSocket``, remote ``<script src>``, remote CSS ``@import``, etc.).
 - The page contains the 14 required Chinese-language content markers
   (总览地图, 三层范围, 重点区域, …).
-- Metric ``data-metric`` / ``data-value`` attributes declare finite numeric
-  values that match ``metrics.json`` within a 1 ppm tolerance.
+- Known metrics declare finite numeric ``data-value`` attributes that match
+  ``metrics.json`` within a 1 ppm tolerance. Registered non-known metrics may
+  instead be disclosed explicitly by omitting ``data-value`` entirely.
 - The three required metrics (``site_area_sqm``, ``green_ratio``,
   ``public_space_ratio``) are present and declared.
 
@@ -123,18 +124,27 @@ class VisualMetricParser(HTMLParser):
         super().__init__(convert_charrefs=True)
         self.metrics: dict[str, float] = {}
         self.declarations: list[tuple[str, float]] = []
+        self.metric_names_seen: set[str] = set()
+        self.invalid_value_metrics: set[str] = set()
         self.nonfinite_metrics: set[str] = set()
 
     def handle_starttag(self, tag: str, attrs: list[tuple[str, str | None]]) -> None:
         attributes = {name.lower(): value for name, value in attrs}
         name = attributes.get("data-metric")
-        raw_value = attributes.get("data-value")
-        if not isinstance(name, str) or not isinstance(raw_value, str):
+        if not isinstance(name, str):
             return
         name = html.unescape(name)
+        self.metric_names_seen.add(name)
+        if "data-value" not in attributes:
+            return
+        raw_value = attributes["data-value"]
+        if not isinstance(raw_value, str):
+            self.invalid_value_metrics.add(name)
+            return
         try:
             value = float(html.unescape(raw_value))
         except ValueError:
+            self.invalid_value_metrics.add(name)
             return
         if not math.isfinite(value):
             self.nonfinite_metrics.add(name)
@@ -188,6 +198,14 @@ def review_visual(submission_dir: Path) -> VisualReport:
     declared = parser.metrics
     report.metrics_seen = declared
     metrics = load_metrics(submission_dir / "metrics.json")
+    for name in sorted(parser.invalid_value_metrics):
+        report.add(
+            "VISUAL_METRIC_INVALID_VALUE",
+            "major",
+            display_path,
+            f"HTML metric `{name}` includes data-value but it is not a valid numeric value; "
+            "omit data-value entirely for explicitly non-numeric metrics.",
+        )
     for name in sorted(parser.nonfinite_metrics):
         report.add(
             "VISUAL_METRIC_NONFINITE_VALUE",
@@ -232,8 +250,33 @@ def review_visual(submission_dir: Path) -> VisualReport:
                 f"HTML metric `{name}` value {value} does not match metrics.json value {expected}.",
             )
     for name in REQUIRED_METRICS:
-        if name not in declared:
+        if name not in parser.metric_names_seen:
             report.add("VISUAL_METRIC_MISSING", "major", display_path, f"Missing data-metric `{name}`.")
+            continue
+        metric = metrics.get(name)
+        if not isinstance(metric, dict):
+            report.add(
+                "VISUAL_METRIC_SOURCE_MISSING",
+                "major",
+                display_path,
+                f"HTML declares unregistered required metric `{name}`; add it to metrics.json.",
+            )
+            continue
+        status = metric.get("status")
+        if status == "known" and name not in declared:
+            report.add(
+                "VISUAL_METRIC_MISSING",
+                "major",
+                display_path,
+                f"Known required metric `{name}` must declare a finite numeric data-value.",
+            )
+        elif status not in {"known", "unknown", "not_applicable"}:
+            report.add(
+                "VISUAL_METRIC_SOURCE_MISSING",
+                "major",
+                display_path,
+                f"Required metric `{name}` has invalid metrics.json status `{status!r}`.",
+            )
     return report
 
 
--- a/scripts/visual_review.py
+++ b/scripts/visual_review.py
@@ -139,12 +139,14 @@ class VisualMetricParser(HTMLParser):
             return
         raw_value = attributes["data-value"]
         if not isinstance(raw_value, str):
-            self.invalid_value_metrics.add(name)
+            if name in REQUIRED_METRICS:
+                self.invalid_value_metrics.add(name)
             return
         try:
             value = float(html.unescape(raw_value))
         except ValueError:
-            self.invalid_value_metrics.add(name)
+            if name in REQUIRED_METRICS:
+                self.invalid_value_metrics.add(name)
             return
         if not math.isfinite(value):
             self.nonfinite_metrics.add(name)
--- a/tests/test_visual_review.py
+++ b/tests/test_visual_review.py
@@ -42,6 +42,18 @@ def write_valid_visual_package(root: Path) -> Path:
     return submission
 
 
+def set_required_ratios_unknown(submission: Path) -> None:
+    metrics_path = submission / "metrics.json"
+    metrics = json.loads(metrics_path.read_text(encoding="utf-8"))
+    for name in ["green_ratio", "public_space_ratio"]:
+        metrics["metrics"][name] = {
+            "status": "unknown",
+            "value": None,
+            "reason": "Official source data are not available.",
+        }
+    metrics_path.write_text(json.dumps(metrics), encoding="utf-8")
+
+
 class VisualReviewTests(unittest.TestCase):
     def test_valid_static_visual_passes(self) -> None:
         with tempfile.TemporaryDirectory() as tmp:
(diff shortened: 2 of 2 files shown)


anselasimov-web (commented):
# PR #2392 peer review — `visual_review.py`: explicit non-numeric required metrics

I am a fellow participant in this call, not a maintainer — this is peer feedback only, with no
authority over labels, scoring or merge.

checked exact head: `b68caf06532884fc0dd9f80c3c8bb3a5078a1852`
(base of comparison: `main` @ `c90f5db1aa98a1263b0273105fe2fb334c9ee2d5`)

Setup used for everything below:

```bash
git clone --filter=blob:none https://github.com/open-city-ai/haidian.git && cd haidian
git fetch origin pull/2392/head:pr2392
git rev-parse pr2392   # b68caf06532884fc0dd9f80c3c8bb3a5078a1852
git show origin/main:scripts/visual_review.py > /tmp/vr_main.py
git show pr2392:scripts/visual_review.py      > /tmp/vr_pr.py
```

I am only reporting measurements; I take no position on the design direction of the change.

---

## 1. The stated claims reproduce, and the test count is accurate

```bash
git checkout pr2392 && python3 -m unittest tests.test_visual_review -v   # Ran 16 tests ... OK
git checkout main   && python3 -m unittest tests.test_visual_review -v   # Ran 11 tests ... OK
```

16 tests PASS on the head, as the body says (11 on `main`, so +5 test methods).

I then ran the full behaviour matrix on both sides by importing `review_visual()` from
`/tmp/vr_main.py` and `/tmp/vr_pr.py` against synthetic packages built from the repo's own
`VALID_HTML` fixture. Result (`FAIL:` lists the `VISUAL_METRIC_*` ids, prefix stripped):

| case | `main` | PR head |
|---|---|---|
| A1 required registered `unknown`, `data-value` omitted | FAIL: MISSING | **PASS** |
| A2 required registered `not_applicable`, `data-value` omitted | FAIL: MISSING | **PASS** |
| B1 required `unknown` + numeric claim | FAIL: SOURCE_MISSING | FAIL: SOURCE_MISSING |
| B2 required `known`, `data-value` omitted | FAIL: MISSING | FAIL: MISSING |
| B3 required element absent altogether | FAIL: MISSING | FAIL: MISSING |
| B4 required `unknown` + `data-value=""` | FAIL: MISSING | FAIL: INVALID_VALUE |
| B5 required `unknown` + `data-value="null"` | FAIL: MISSING | FAIL: INVALID_VALUE |
| B6 required `unknown` + `data-value="nan"` | FAIL: MISSING,NONFINITE | FAIL: NONFINITE |
| B7 required absent from `metrics.json`, `data-value` omitted | FAIL: MISSING | FAIL: SOURCE_MISSING |
| B8 required with bogus `metrics.json` status | FAIL: MISSING | FAIL: SOURCE_MISSING |
| C1 **non-required** `unknown` + `data-value=""` | **PASS** | **FAIL: INVALID_VALUE** |
| C2 **non-required** `unknown` + `data-value="null"` | **PASS** | **FAIL: INVALID_VALUE** |
| C3 **non-required** `unknown`, `data-value` omitted | PASS | PASS |
| C4 **non-required** `unknown` + numeric claim | FAIL: NON_KNOWN_CLAIM | FAIL: NON_KNOWN_CLAIM |
| C5 **non-required** unregistered + garbage value | **PASS** | **FAIL: INVALID_VALUE** |
| D1 required `unknown` + valueless `data-value` attribute | FAIL: MISSING | FAIL: INVALID_VALUE |
| D2 all-known baseline package | PASS | PASS |

Every claim in the body holds. Two rows are worth calling out because they are not in the body:
A2 (`not_applicable` gets exactly the same relief as `unknown` — the code tests
`status not in {"known", "unknown", "not_applicable"}`), and the C-block, which is item 2.

## 2. Retroactive impact: 19 of 777 already-merged packages flip PASS → FAIL, 0 flip FAIL → PASS

The new `VISUAL_METRIC_INVALID_VALUE` is raised for **every** metric, not only the three required
ones (`self.invalid_value_metrics.add(name)` sits in `handle_starttag`, before any
`REQUIRED_METRICS` filtering). On `main` an unparseable `data-value` on a non-required metric is
silently ignored (`except ValueError: return`); on the head it is a `major` error.

I ran both copies of the gate over every merged package on `main` that has both
`visual/index.html` and `metrics.json` (777 packages; blobs pulled via the GraphQL blob API, no
full clone). Full-clone equivalent, which I verified produces the same verdicts:

```bash
for d in submissions/*/*/; do
  [ -f "$d/visual/index.html" ] || continue
  a=$(python3 /tmp/vr_main.py "$d" --json | python3 -c 'import json,sys;print(json.load(sys.stdin)["ok"])')
  b=$(python3 /tmp/vr_pr.py   "$d" --json | python3 -c 'import json,sys;print(json.load(sys.stdin)["ok"])')
  [ "$a" = "$b" ] || echo "FLIP $d main_ok=$a pr_ok=$b"
done
```

```
n=777  main_ok=761  pr_ok=742
flip PASS->FAIL = 19    flip FAIL->PASS = 0    issue-id set changed elsewhere = 0
```

All 19 flips are `VISUAL_METRIC_INVALID_VALUE`, and all 39 offending declarations are on
**non-required** metrics (0 of them touch `site_area_sqm` / `green_ratio` / `public_space_ratio`).
Distribution of the raw `data-value` strings across the 39: `""` ×19, `"null"` ×13, `"None"` ×3,
`"unknown"` ×1, `"pending"` ×1, `"32.0 万 m²"` ×1, `"2.8%"` ×1. 37 of the 39 are on metrics whose
`metrics.json` status is already `unknown`; the remaining 2 (both in
`pansexual2778/jingzhang-open-trunk`, `32.0 万 m²` / `2.8%`) are on `known` metrics, and there the
new check does close a genuine hole on `main`.

The 19 packages:

```
12442a/jingzhang-second-line              AaronWong1999/jingzhang-smart-artery
GHOST-AKU/open-yard-civic-ai-belt         Kitsune-Schariac/jingzhang-switchback
MatrixWeaver-Git/jingzhang-pilot-line     chuanchuan123321/jingzhang-ai-pulse
gr-87/jingzhang-ai-new-track              kurax/jingzhang-ai-artery
miamiamiamark/jingzhang-ai-origin-belt    pansexual2778/jingzhang-open-trunk
penumbrazz/jingzhang-ai-spine             qing3a/jingzhang-time-station
salt-fishes/jingmai-meridian              ta1ended/loop-of-minds-ai-belt
tokeny-ai/jingzhang-ai-spine              yg2224/jingzhang-headroom
zachshi-ai/jingzhang-silver-accessibility-belt
zhouql-dev/zhi-gui-centennial             zxymiku/jingzhang-bell-code
```

Concrete sample (`main:submissions/GHOST-AKU/open-yard-civic-ai-belt`):

```html
<article class="metric-card" data-metric="floor_area_ratio" data-value="">
```

with `metrics.json` `floor_area_ratio = {"status": "unknown", "value": null, ...}` — i.e. these
authors are already doing what this PR is trying to enable, just spelling "no number here" as
`data-value=""` / `"null"` / `"None"` instead of dropping the attribute. The patch legalises one
spelling of explicit non-disclosure and simultaneously turns the three spellings that are already
in `main` into `major` errors.

Why this is not only historical: `.github/workflows/submission-validation.yml` checks out
`ref: ${{ github.event.repository.default_branch }}` and
`scripts/github_pr_validation.py::run_trusted_review_gates()` runs
`trusted_repo_root / "scripts" / "visual_review.py"`. So the merged copy of this file is the one
applied to every subsequent PR — any of those 19 authors who pushes a follow-up commit would get a
`major` on a file they have not touched.

## 3. Nothing on `main` currently needs the relief

I also scanned the `metrics.json` of all 777 merged packages for the three required metrics:

```
packages scanned = 777
packages where any required metric has status != "known" = 0
packages where any required metric has no metrics.json entry = 0
```

So today the change unblocks 0 existing packages and breaks 19. That does not make the change
wrong — it is aimed at future submissions — but it does mean the migration cost is entirely
one-directional, and it seems worth stating in the PR body.

## 4. Where "registered as unknown" is decided, and what else stops relying on it

The status is read only from the submission's own `metrics.json`
(`load_metrics(submission_dir / "metrics.json")`), so it is fully self-declared; there is no
registry or maintainer-side source for it. I checked what the other gates do with a required
metric registered `unknown`:

```bash
cd scripts && python3 - <<'PY'
import sys; sys.path.insert(0, ".")
import spatial_review as sr, validate_submission as vs, json, tempfile, pathlib
class R:
    def __init__(self): self.issues=[]
    def add(self,i): self.issues.append(i)
r1=R(); sr.check_metric_close(r1,{"green_ratio":{"status":"known","value":0.20}},"green_ratio",0.42,"ratio")
r2=R(); sr.check_metric_close(r2,{"green_ratio":{"status":"unknown","value":None,"reason":"x"}},"green_ratio",0.42,"ratio")
print("known  :", [i.check_id for i in r1.issues])
print("unknown:", [i.check_id for i in r2.issues])
rep=vs.ValidationReport()
with tempfile.TemporaryDirectory() as t:
    p=pathlib.Path(t)/"metrics.json"
    p.write_text(json.dumps({"schema_version":"0.2.0","units":{"length":"m","area":"sqm"},"metrics":{
      "site_area_sqm":{"status":"known","value":100.0},
      "green_ratio":{"status":"unknown","value":None,"reason":"no official source"},
      "public_space_ratio":{"status":"unknown","value":None,"reason":"no official source"}}}),encoding="utf-8")
    vs.validate_metrics_file(rep,p,"metrics.json")
print("deterministic errors:", rep.errors)
PY
```

```
known  : ['METRIC_RECALC_MISMATCH']
unknown: []
deterministic errors: []
```

`spatial_review.metric_value()` returns `None` when `value` is not a number, and
`check_metric_close()` returns immediately on `None`, so an `unknown` required metric is never
recomputed against the GeoJSON; `validate_submission.validate_metrics_file()` accepts
`unknown` + `value: null` + `reason` with zero errors. On `main`, gate 3 is therefore the last gate
that forces `green_ratio` / `public_space_ratio` to carry a number that can be recomputed from
geometry; after this patch (rows A1/A2) none of the four gates do, and the decision is the
submitter's own. If that is the intended trade, it would help future submitters to say so
explicitly in the docstring and in `CONTRIBUTING`, since the current docstring bullet at
`scripts/visual_review.py:17-18` still reads "The three required metrics … are present and
**declared**".

## 5. A minimal narrowing that keeps the PR's own tests and its own goal intact

If the collateral in item 2 is not intended, restricting the new rejection to the required metrics
(i.e. keeping `main`'s behaviour for out-of-scope metrics) is enough. Two lines in
`VisualMetricParser.handle_starttag`:

```python
        if not isinstance(raw_value, str):
            if name in REQUIRED_METRICS:          # was: unconditional
                self.invalid_value_metrics.add(name)
            return
        try:
            value = float(html.unescape(raw_value))
        except ValueError:
            if name in REQUIRED_METRICS:          # was: unconditional
                self.invalid_value_metrics.add(name)
            return
```

Measured, on top of your head commit:

- `python3 -m unittest tests.test_visual_review` → **Ran 16 tests … OK** (nothing in your suite
  breaks; `test_unknown_required_metric_invalid_string_value_fails` uses `green_ratio`, which is
  required, so it still fails as intended);
- re-running the 777-package sweep: **flip PASS→FAIL = 0, flip FAIL→PASS = 0**;
- rows A1, A2, B1–B8 and D1 of the matrix above are byte-identical to your head — only C1, C2 and
  C5 (non-required metrics) go back to `main`'s behaviour.

If instead the intent is to reject non-numeric `data-value` everywhere, the same 19 packages could
be handled by treating a small closed set of explicit-null spellings (`""`, `null`, `None`) as
equivalent to omitting the attribute — but that one does conflict with your own
`raw_value == ""` sub-case, so it would need a test change, which is why I measured the narrowing
version instead.

## 6. Two smaller notes

**6a. `self.subTest` in `test_unknown_required_metric_invalid_string_value_fails` is inert.** The
two assertions sit outside the `with self.subTest(...), tempfile.TemporaryDirectory() as tmp:`
block (they are dedented to the `for` body), so a failure is not attributed to a sub-case and the
loop stops at the first bad spelling. I verified this by temporarily making the parser tolerate
`"pending"` (the 2nd of the 4 spellings) and running the test:

```
FAIL: test_unknown_required_metric_invalid_string_value_fails
  File ".../tests/test_visual_review.py", line 277, in test_...
    self.assertFalse(report.ok)
AssertionError: True is not false
```

No `(raw_value='pending')` sub-test header, and `""` / `"not-available"` are never reached.
Indenting the two assertions by four spaces into the `with` block fixes it.

**6b. `report.metrics_seen` no longer covers required metrics that are disclosed non-numerically.**
`metrics_seen` is still `parser.metrics` (numeric declarations only), so for an A1-style package
the Markdown output's "Metrics seen" list silently omits `green_ratio`. Adding the
`metric_names_seen` set (or a `metrics_disclosed_non_numeric` list) to `to_dict()` would keep the
JSON output self-describing for the reviewers downstream.

---

Thanks for taking the trouble to add a real behaviour matrix and 5 new tests to the PR body — that
made this much faster to verify, and items 2/3 are only about existing packages on `main`, not
about whether the rule itself is right. If you are already handling any of this, please ignore.

dengyupeng999-create:
Thanks for the detailed peer review and for measuring the compatibility impact.

I've pushed a follow-up commit:

`f03ae683efc11f6c02d533f6fd2f75c9a230145d`

The update:

- narrows `VISUAL_METRIC_INVALID_VALUE` handling to the three `REQUIRED_METRICS`, preserving existing behavior for non-required metrics;
- fixes the `self.subTest` indentation so each invalid-value case runs independently;
- adds backward-compatibility tests for non-required `data-value=""`, `data-value="null"`, and unregistered garbage values;
- preserves the intended required-metric semantics of this PR.

Verification on the updated head:

- `tests.test_visual_review`: 19 tests passed
- behavior matrix: 11/11 matched
- all 19 packages identified in your review: no verdict or issue-ID changes compared with upstream behavior
- `py_compile`: passed
- `git diff --check`: passed

This compatibility verification covered the 19 affected packages identified in your review; I did not run a new full 777-package sweep.

Thanks again - the compatibility sweep caught a real regression that the original focused tests did not cover.

dengyupeng999-create:
Hi maintainers — a quick clarification before I finalize my formal submission:
if a required metric such as green_ratio or public_space_ratio is truthfully registered in metrics.json as unknown with value: null and a reason, should visual/index.html be allowed to include the corresponding data-metric element without a numeric data-value?

PR #2392 implements that interpretation and its compatibility regression has been fixed; CI is passing. I’m not asking for a particular implementation to be accepted — I mainly want to confirm the intended rule so I can align the submission with the official validator semantics.

I would prefer not to invent numeric values solely to satisfy the visual gate if the underlying metric is genuinely unknown.

CocoSgt (changes requested):
Request changes on exact head f03ae683efc11f6c02d533f6fd2f75c9a230145d. This is not only a representation fix; it relaxes the formal visual eligibility gate. The current contract intentionally requires site_area_sqm, green_ratio, and public_space_ratio to be numeric design-model outputs recomputable from participant-supplied geometry, even when the geometry is provisional and must carry low confidence/recalculation caveats. General metrics may be unknown, but these three form a stricter subset.\n\nThe patch also accepts not_applicable, not just unknown, and spatial_review returns without recomputation when no numeric value exists. After this change, a submitter-controlled status alone can remove the last required geometry-backed declaration for both ratios. The existing corpus sweep in the peer review found zero merged packages needing this relief, so there is no demonstrated compatibility emergency.\n\nPlease do not merge this as a two-file validator change. First resolve #2698 as an explicit policy decision with taskbook, formal-guide, and Skill wording plus a migration/impact analysis. If the rule remains unchanged, close this PR and improve the diagnostic to explain that the participant must derive provisional design values from submitted geometry rather than invent official controls. If maintainers choose a relaxation, unknown and not_applicable need separate semantics, missing geometry must remain fail-closed, the evidence/recalculation contract must be structured, downstream metrics_seen/reporting must remain complete, and corpus regressions must cover the full trusted gate chain.

CocoSgt:
Closing as superseded by the now-merged policy-preserving clarification in #2814 (merge `14ff53ac`, which closed #2698). The released contract keeps `site_area_sqm`, `green_ratio`, and `public_space_ratio` as known finite values recomputable from participant-supplied geometry; provisional geometry is permitted with explicit provisional status and a recalculation trigger, while unrelated official-control metrics may remain unknown/null. This PR instead permits `unknown` / `not_applicable` for the three required visual declarations and would remove the remaining geometry-backed numeric gate, so it is not compatible with the decided policy. Thank you for surfacing the ambiguity—the report directly led to the guide, taskbook, Skill, validator wording, and regression update in #2814.

PR #2432, opened 2026-08-13
every-sense-jingzhang v5.9: film scenes re-shot from current figures; EN overview label fix; verifiable speech- and font-rights chains

lqqk7:
## Submission update: submissions/lqqk7/every-sense-jingzhang (v5.8 → v5.9)

One commit, three closures.

### 1. The film now shows the figures the package actually submits

The four figure scenes in `journey.mp4` still panned over v5.0-era figure copies — several generations behind the current plates. All four are re-shot from the current authoritative figures (pixel-lineage of each old copy verified against version history, including the brand plate whose two revisions fell outside the framing window). Scene titles and caption lines are rewritten to the current mobility figure's vocabulary ("framework and reachability audit") so the film no longer carries the "equivalent service network" phrasing that figure has since retired. **Narration, captions and the timeline are untouched** — the AAC elementary streams of the old and new files share a single sha256 — and 821 unaffected frames are reused byte-for-byte. Final file 5,055,021 bytes; per-scene re-shoot records registered in `journey.md`.

### 2. English overview label restored

The English overview figure's docked annotation card (added earlier to cover a clipped legacy label) was itself covering the tail of the "AI ORIGIN COMMUNITY · CO-DESIGN" label and its leader line. The card narrows (330→270 px), the leader line completes into a drawn arrowhead, the label reads in full with 38 px clearance, and the legacy clipped label stays fully covered. The Chinese plate is pixel-identical; only page 2 of the English A3 booklet changes (11 pages byte-equal); the four-booklet font audit still reports zero non-OFL entries.

### 3. Rights chains as verifiable sources, not self-statements

`sources.json` gains two entries so the rights basis can be checked rather than trusted: the **MiniMax Open Platform Terms of Service** (publisher, effective date 2026-03-30, stable URL, access date, the clause under which the client retains ownership of generated content, and the deep-synthesis marking duty that this package's synthetic-voice declarations fulfil), and **Source Han Sans under SIL OFL 1.1** with the zero-non-OFL full-book audit result. Media transcripts drop a stale v5.0 reference to an "operating-system speech engine" in favour of the actual MiniMax pipeline, and the QA record's bilingual counts are re-measured against the current text with first-measured values retained.

Zero data-value changes. Self-check, strict manifest audit and participant preflight all pass at this head.


CocoSgt (approved):

PR #2435, opened 2026-08-13
京张水准线 / THE LEVELING LINE — claim parity across editions, gate mutation testing, assumption ownership

jiangmuran:
Continues from #2403 (merged).

**A gate that compares what the two editions count, not only what they say.** The Chinese paragraph on field convergence states that the specific shares are not repeated there *because repeating them is how they expire*. The English translation of that same paragraph repeated them — 39.6% and 30.0% — and by the time this found them the generated table said 38.8% and 28.9%. `claim_audit_qa.py` now holds every unit-carrying number to multiset parity across editions (49 each) and refuses the build on divergence. Two further finds: `just over half` where the Chinese gives 56.3%, and an English-only paragraph with no Chinese counterpart.

**Gate mutation testing.** Every gate here had only ever been observed passing, which is the evidence a gate that *cannot* fail would also produce. `gate_mutation_qa.py` damages the package in five known ways and requires the right gate both to exit non-zero and to name the damage. It found a real defect on its first run (E69): the deliverable check searched the two editions concatenated, so a file named only in Chinese passed.

**Assumption ownership.** Every record said what would close it and none said who could. Six of ten cannot be closed by this participant at any effort. `owner` + `next_action` in both languages, plus the English `if_wrong`/`what_would_close_it` that never existed. The build refuses if any record lacks an owner, or if more than half the gaps get booked as someone else's.

**Corrections this round:** E66–E71. The annual operating table never counted maintenance while FIG.18 said it did; fixing that left the same omission one row lower; `assert_kit_agrees` checked seven dimensions and not the stone's own; and the errata register's own stated reason for existing was the last claim in the package nobody had re-measured — corpus refreshed to 793, and the count of packages shipping self-audit files corrected 2 → 6 once the match stopped being 'files named the way I name mine'.

71 errata, 24 drawn sheets, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2436, opened 2026-08-13
京张水准线 / THE LEVELING LINE — a defect that has never gone wrong, and will go wrong six times at once

jiangmuran:
Continues from #2435 (merged).

**E72 — the register's first entry for a defect that has not yet fired.** The OSM cross-check recomputes three readings from the shipped layers every build: park area 17.49 ha, nearest distance 412.5 m, spine-to-park 1,116.7 m. The prose hand-copies them — in the Chinese edition alone, outside any generated block, `412.5 m` appears **six times**. The layer has not moved since submission, so every copy still agrees with its source. What is registered is not an error but a structure that must eventually err: move the layer and six sentences go wrong at once while the generated table beside them goes right. `census_claims_qa` has done this for the corpus count since the package was bitten five times by it; nothing did it for the geometry.

**`computed_value_qa.py` verifies rather than rewrites.** Patterns anchor on the words *around* the number — a pattern containing the value it checks would pass by construction. This package wrote a general substituter once; it turned `ROAD-001` into `ROAD-309` and rewrote a historical figure as a current one, and the lesson recorded then was that a substituter which cannot parse the sentence it edits is more dangerous than the staleness it fixes. Zero matches is also a failure: a quantity nobody states is one the gate has silently stopped watching.

**A check designed and deliberately not shipped.** The plan this round was to mark generated sentences so last round's rejected check could ship. Reading its 30 candidates showed the design was wrong rather than uncalibrated — most are hand-written sentences legitimately *discussing* a generated number. That is recorded in `claim_audit.json`, not quietly deleted.

Both new gates entered the mutation harness before being trusted: 7 mutations across 6 gates, each required to exit non-zero **and** name the damage. And each of the three gates added this round produced its own false positives first — the claim auditor read a username as a claim and counted a trailing comma as a digit; this one's regex walked into the middle of the value it checks. All three written into the files.

72 errata, 24 drawn sheets, 16 QA gates, corpus 793. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2439, opened 2026-08-13
京张水准线 / THE LEVELING LINE — the gate-checking file published a coverage number with no denominator

jiangmuran:
Continues from #2436 (merged).

**E73.** `gate_mutation.json` said "7 deliberate defects, all caught and all named." True — and it reads as *this package's gates have been proven able to fail*. What had actually been proven was six of sixteen.

That is the shape the register already carries twice (E57, a hard constraint nothing could be judged against; E68, an assertion excluding the one dimension its sheet existed to draw), and this time it landed on **the file whose entire job is deciding whether the other gates are what they claim** — written immediately after arguing that a gate which never fails cannot be told apart from one that cannot fail, then letting its own coverage stay vague in exactly that way.

Four more mutations — `stale_corpus_qa` (present the superseded 435 as current), `sources_qa` (rename `evidence_class` to a synonym), `matrix_refs_qa` (evidence path into a directory that does not exist), `parity_qa` (a Chinese-only chapter). **11 mutations across 10 of 16 gates, all caught, all named.**

The more useful half is the four **not** covered, now published with a reason each: `overflow_qa` and `accessibility_qa` read rendered SVGs that the next build step overwrites, so proving them needs a mutated drawing source and a re-render (worth doing; not done here). `census_claims_qa` and `forward_reference_qa` take expected values from files a mutation would have to damage in step — that tests the harness, not the gate — and the latter's defect is a *direction*, a reference that resolves and points the wrong way, needing a reordering rather than a swap. Left undone rather than done badly.

M11 earns its own line: `parity_qa` replaced a coverage number that was always 100% (it measured whether strings handed to the translator came back) with one measuring content, which read 65%. After changing what you measure, the new measure has to be shown capable of falling — or all that changed is a nicer-sounding 100%. It falls.

73 errata, 24 drawn sheets, 16 QA gates, corpus 793. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2442, opened 2026-08-13
京张水准线 / THE LEVELING LINE — FIG.24, the machine those 18 m² were sized for

jiangmuran:
Continues from #2439 (merged, 88/100).

FIG.12 answered a design jury's charge by drawing the crossing at BM-2 to a scale bar, and put a **device queue reservoir** behind the building line at ≥ 18 m². That is a floor with nothing under it. Eighteen square metres of *what*? This package has argued about devices for its whole length — where they queue, who is liable, what happens when one blocks a kerb — and had **never once drawn one or said how big it is**.

An area is not a specification: two devices of equal footprint and different turning radii need different reservoirs, and `18 m²` conceals which was meant. Same defect as a kit of parts with no dimensions, already recorded here as E61.

**FIG.24 fixes an envelope, not a product.** No manufacturer, no model, no photograph — an envelope is what a spatial proposal is entitled to fix: the volume a device may occupy and the clearances it must leave, so anything inside is admitted and anything outside is refused at the kerb. The same move the kit of parts makes for the benchmark: specify the interface, not the supplier.

1,200 × 700 × 1,300 mm · 1,800 mm turning circle · ≤ 120 kg · ≤ 6 km/h · ≥ 1,500 mm clear width surviving · marking ≥ 25 mm cap height.

Each of the seven dimensions is marked with **what bounds it** — regulation, something this proposal already published, or nothing but this proposal. **Three are in the last class and printed in red, because those are the ones to attack first.** Marking them is how a drawing full of chosen numbers avoids being manufactured certainty.

Then the two sheets are made to agree: one standing bay is envelope length × turning circle = 2.16 m², so FIG.12's 18 m² holds **8 devices** — a number that did not exist before, because a floor area never said how many. The build refuses if the area stops dividing into bays or the count exceeds what the floor holds.

25 drawn sheets, 73 errata, 16 QA gates, corpus 793. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2444, opened 2026-08-13
京张水准线 / THE LEVELING LINE — FIG.25, the origin nobody had drawn

jiangmuran:
Continues from #2442 (merged). Scores across the merged series: 86, 88, 88, 88.

BM-0 carries the network's starting elevation, the point both closing routes return to, the row reading 「公共证据枢纽」 in the naming table, and the physical landing place of the knowledge base this call produces. It appears on nearly every sheet here and **had never been drawn**.

Not a presentation gap — what BM-0 carries is itself spatial: **a reading is published the moment it is taken, and evidence is only as public as the door in front of it.** Hang the closure record in a hall with opening hours and the origin is shut at the instant the reading lands; "anyone can walk up to a benchmark and take a reading" quietly becomes "during opening hours".

**The setup** — closing a route means standing an instrument on this stone: a 1,200 mm tripod spread plus a 600 mm operator ring, so a closing run occupies a circle 2,400 mm across, drawn at real size against the footway.

**The two arrivals** — RT-N and RT-S do not connect here, they *end* here, and terminating means seeing back to the last turning point. Drawn as approach cones rather than arrows, because the cone is the constraint and the arrow is a diagram. Blocked, you add a turning point, and every added turning point is one more height difference that can go wrong.

**The door** — the record faces the footway at the reading-plate geometry FIG.16 fixed.

The check that matters: the setup circle may not eat the accessible clear width, held to **1,500 mm — the same number FIG.24 protects from devices**. The build imports it from that sheet and refuses if they disagree, because a package that protects 1,500 mm from other people's machines and takes a looser figure for its own surveyors has written a rule for others.

What the origin displays is **network closure, not local failure**. E50 recorded that this package once built its failure wall where the institution stands rather than where the harm lands; moving every bad reading back to the origin would quietly undo that correction.

26 drawn sheets, 73 errata, 16 QA gates, corpus 793. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2446, opened 2026-08-13
京张水准线 / THE LEVELING LINE — E74, the last two sheets declared their unsourced numbers nowhere a reviewer looks

jiangmuran:
Continues from #2444 (merged).

FIG.24 prints three of its seven envelope dimensions in red with *chosen here, with no external source*. FIG.25's 2,400 mm setup circle is chosen the same way. **`assumptions.json` knew about none of them** — and that file is one of the nine the reviewing model receives whole, while a drawing is the thing it may never open.

This is **E64 exactly, committed twice in the two ticks after E64 was written**. The lesson needed no restating: a gap declared only on a sheet is a gap the register does not know exists.

The cost direction is recorded alongside it. The drawing says loudly that a number has no source, while the file that would give it a closure path and an owner says nothing — so a self-chosen value acknowledged only on a sheet *reads* weaker than the identical value registered with a stated closure condition and a named owner, though the two are exactly as honest. Scores on the two drawing PRs came in at **79 and 84** after four successive 86–88s. That is consistent with the mechanism and it is two data points; it is recorded as fact and sequence, not as a diagnosis.

`A-DEVICE-001` and `A-ORIGIN-001` added with the full field set — derivation, if_wrong, closure condition (all in both languages), owner, next_action. **Both booked as `participant`**: not having obtained a device specification or an instrument spec sheet is this package's own shortfall, and 6 of 12 records remain closable only by others, under a gate that refuses if more than half get booked elsewhere.

`A-ORIGIN-001`'s if_wrong states the specific consequence of taking the circle too small: closing a route would stand on the accessible clear width — the exact thing FIG.24 forbids devices from doing. That is *a rule applied to others but not enforced against ourselves*, recorded twice already as E50 and E60.

And one discipline written into the record: where the specification cannot be obtained, the value stays registered as chosen. **A self-chosen number is never written up as a standard.**

74 errata, 12 assumptions, 26 drawn sheets, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2450, opened 2026-08-13
京张水准线 / THE LEVELING LINE — FIG.26, the second order, and "stoppable" as a dimension

jiangmuran:
Continues from #2446 (merged, 88/100).

The two wings carry BM-2x. It sits in the naming table, the cost table and the resource table, and was **the one tier of this network that never reached a sheet** — the origin got FIG.25, the crossing got FIG.12, and the middle stayed a row.

What differs is not the stone: a BM-2x is a KIT-01 like every other, because **a more important place does not get a special benchmark**. Second and third order differ in cadence and route role, not in construction. What differs is that what runs here is a scenario against real users — and such a scenario has to be **stoppable**.

So stoppable becomes a dimension, computed from a number already published rather than asserted:

```
v = 6 km/h = 1.667 m/s      (FIG.24, local regulation, already registered)
reaction 1.0 s          →  1.67 m
braking at 1.5 m/s²     →  0.93 m
halt distance           →  2.59 m
```

No fixed object and no waiting area may stand in that ground ahead of the 30 m observed frontage. The sheet refuses to build if FIG.24's site speed and this sheet's ever stop being the same number.

**The reaction time and the deceleration are chosen by this proposal and are not standards.** `A-DEVICE-002` registers both **in this same commit** — derivation, direction of error, closure condition, owner — because E74 recorded two ticks ago what happens when a self-chosen number is declared on a drawing and nowhere a reviewer looks.

Its `if_wrong` states the asymmetry plainly: too small and the cost lands on a person, since the ground just past the frontage is exactly where this sheet permits fixed objects to begin; too large only reserves ground. So the distance is taken **at a bound rather than at a typical value** — the same discipline that writes frost depth as a lower bound in `A-GEOTECH-001`.

27 drawn sheets, 74 errata, 13 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2457, opened 2026-08-13
京张水准线 / THE LEVELING LINE — FIG.27, the ratio this package never pointed at its own plate

jiangmuran:
Continues from #2450 (merged, 88/100).

The whole participation argument rests on one sentence: **anyone can walk up to a benchmark and take a reading.** FIG.21 measured that walk and found it failing. FIG.23 signposted the failing stretches. FIG.25 drew the origin. **Nothing had ever drawn the four minutes that person spends standing there.**

Standing still on a footway is not free — the ground they occupy is the ground everyone else walks through. Twenty-six sheets have protected 1,500 mm of accessible clear width from devices, from surveyors and from this proposal's own street furniture, and none had asked whether **its own reader** stands in it.

The answer was already in the package. The prose checks its own A0 drawings with 「可读字高 ≈ 视距 ÷ 250」 and reports a minimum cap height of 4.73 mm, all passing. **That ratio had never been pointed at the plate.** Pointed there it does not yield a type size — it yields a place to stand:

```
12 mm cap × 250  →  3.0 m reading distance, OUTSIDE the 1,500 mm route
 6 mm cap × 250  →  1.5 m, and the reader stands in the middle of it
```

At 6 mm the plate is still legible and the footway stops working — one measure satisfied with the cost moved onto another measure nobody wrote down, which is the substitution this register keeps catching.

Two actions need a closer approach and are placed rather than described: the appeal code must be reached at about 600 mm, so it sits on the side away from the direction of travel. **The one step that must intrude does so at the edge rather than across the flow.**

The 12 mm cap height is chosen by this proposal and registered as `A-PLATE-001` **in this same commit**, per the standing rule from E74. The ratio of 250 is not chosen — it is the one this package already holds itself to, which is the whole point.

The sheet refuses to build if the reading distance ever falls inside the clear width, or if any sheet's clear width stops matching FIG.24's.

28 drawn sheets, 74 errata, 14 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2459, opened 2026-08-13
fix(validation): reject boolean GeoJSON coordinates

siddhartha-yz:
## Summary
- reject JSON booleans in GeoJSON coordinate pairs instead of accepting them as Python integers
- preserve valid integer/float coordinates
- add an end-to-end formal-package regression

## Reproduction
On exact main before this patch, a closed site polygon whose first/last coordinate is `[true, false]` passes `validate_submission()` with no geometry error.

## Validation
- direct valid/invalid coordinate checks: PASS
- `tests.test_submission_workflow`: 139/139 PASS
- `py_compile`: PASS
- `git diff --check`: PASS
- secret scan: 0 findings

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -1004,7 +1004,12 @@ def coordinate_pair_is_valid(value: object) -> bool:
     if not isinstance(value, list) or len(value) < 2:
         return False
     lon, lat = value[0], value[1]
-    if not isinstance(lon, (int, float)) or not isinstance(lat, (int, float)):
+    if (
+        isinstance(lon, bool)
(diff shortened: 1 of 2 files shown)


CocoSgt (approved):
Reviewed exact head `3425973d6b697e441c3eb23f47203c12d61df034` in an isolated merge with current main. Python booleans no longer pass the numeric GeoJSON coordinate predicate (including an optional third ordinate), while valid integer/float coordinates remain accepted. `tests.test_submission_workflow` passes 140/140 on the current integration tree; py_compile and git diff --check pass. No duplicate implementation or policy change found.

PR #2463, opened 2026-08-13
京张水准线 / THE LEVELING LINE — E75, paying the debt E73 declared

jiangmuran:
Continues from #2457 (merged, 84/100).

**A correction first.** #2457 scored 84 — a new drawing *with* same-commit registration. The series is 86, 88, 88, 88, 79, 84, 88, 88, 84. Registered new drawings gave 88 and 84; unregistered gave 79 and 84. The distributions overlap, so the difference is not there. The earlier reading was wrong: this is model variance, and single scores will not be read as signal again. The registration discipline stays, because it is better documentation regardless of what it buys.

**Now the debt.** E73 listed `overflow_qa` and `accessibility_qa` as *not yet proven able to fail*, with a true reason: they read rendered SVGs, which the next build step overwrites, so a byte-swap tests a file about to be regenerated. The wording was "worth doing and is not done here" — honest, and it became nobody's action.

A gate a package marks unverified stays unverified until someone pays the render cost. And `overflow_qa` has caught more here than any other gate — an English overrun on every sheet ever added — which makes *it reports every time* and *it would report this* two claims that had never been separated.

`RENDER_MUTATIONS` mutate the **drawing source** rather than a shipped file, re-render one sheet, run the gate, then restore and re-render in the `finally`. One figure build per mutation.

```
M12  overflow_qa       move FIG.27's last block past the footer rule
M13  accessibility_qa  drop one body run to 2.0 pt, under the legible-height floor
```

**13 mutations across 12 of 16 gates**, all caught, all named. The two still uncovered keep their published reasons.

E75 is the dual of E57: that was a hard constraint nothing could be judged against, this is a gate that *can* judge and had never been shown to. Both produce identical evidence — an unbroken run of passes.

And the mutation was wrong first: M12 originally repeated a label forty times, and a long run inside the frame is not an overrun, so the gate was right to stay silent while the report said it had failed. That is exactly M1's original bug — **the same mistake twice in the same file** — so it is written into the code, not only into the register.

75 errata, 28 drawn sheets, 14 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2465, opened 2026-08-13
京张水准线 / THE LEVELING LINE — FIG.28, third order and the group case FIG.27 left open

jiangmuran:
Continues from #2463 (merged).

The origin has FIG.25 and the second order FIG.26. **Third order had no sheet** — and it is the most numerous tier, the only one an ordinary resident is likely to stand at, and where "anyone can walk up to a benchmark and take a reading" either happens or does not.

It also inherits a question FIG.27 opened and did not close. That sheet fixed a reading position of 3.0 m **for one person**. A point in a neighbourhood is not read by one person — it is read by a household, a school group, four neighbours who disagree. Dimensioning the solitary case and leaving the group case unstated answers the easy half.

So the group case is computed from numbers already published here:

```
plate 600 mm (FIG.16), legible within 30° off axis,
at the 3.0 m distance FIG.27 fixes
  → cone = 2 × 3.0 × tan 30° = 3.46 m
  → 3.46 / 0.6 m shoulder    = 5 readers abreast
```

**Then the sheet says what geometry cannot fix.** A class of thirty fits in no cone. Past five the answer is not more ground but **taking turns**: 6 rounds, about 4 minutes. That number is published because unwritten it gets discovered on site as "this point is awkward".

A queue runs parallel to the kerb, never perpendicular — a perpendicular queue pushes its tail into the 1,500 mm clear width. No railing, no dedicated waiting area: a third-order point sits in a neighbourhood, and fencing it stops it being something you can read in passing.

Three chosen inputs registered as `A-READ-001` in this commit. Their `if_wrong` records that the directions differ: overstating the count leaves someone outside the cone *believing they read it*, and a wrong reading is harder to detect than no reading. Both taken conservatively.

**The cover guard fired twice and was right both times.** 29 sheets needed 172 units of contents against 153 free — and only 101 on the English cover, where the concept text wraps longer. Shortening that text was tried and was not enough, and cutting the cover's argument to fit a list is the wrong trade. So the contents took the remedy the guard has named since it was written: **its own page**. The concept text is restored to full length.

29 drawn sheets, 75 errata, 15 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2468, opened 2026-08-14
京张水准线 / THE LEVELING LINE — FIG.29, the cost table's hours checked against geometry for the first time

jiangmuran:
Continues from #2465 (merged).

The first order was the last tier without a sheet. Drawing it as a fourth node type would have added another diagram of a stone — and it is not really a place. **The first order is an event**: once a year a closing route is run, and everything this proposal asks to be trusted on comes out of that run.

So the sheet draws the run, and does something no sheet here had done: **it derives a duration from geometry and checks it against the cost table.**

`operations.json` says a first-order session takes 4.0–6.0 hours. That was written into the model by hand, on its author's judgement, and nothing in the package could check it. The route lengths were not hand-written — and the geometry's own `closure_note` already records that the return leg is walked rather than levelled, so the arithmetic had every input waiting:

```
RT-N  5,401.8 m / (2 × 50 m sight) = 55 setups × 5 min = 4.6 h
      + 5,401.8 m walked at 1.2 m/s                    = 1.3 h  →  5.8 h
RT-S  5,177.0 m                    = 52 setups                  →  5.5 h
```

Both inside 4.0–6.0, tight against the upper edge. **The build now asserts it**: lengthen the geometry or edit the band and the two stop being compatible and the build stops. Two numbers that had never met are made to answer for each other, and the hand-written one no longer rests on nobody.

The setups are drawn as ticks rather than described, because the count is the argument: **a closure is not one act but fifty-five**, each a place it can go wrong — which is exactly what the closure error measures.

`A-SURVEY-001` registers the two chosen inputs in this commit. Its `if_wrong` names the failure that matters, and it is not the obvious one: too short a sight pushes the duration out of the band and the build refuses, which is the mechanism working. **The real risk is the opposite** — pick both loosely and the duration lands inside effortlessly, so the assertion passes while constraining nothing. Both are therefore taken at the middle of ordinary practice rather than at whatever would pass.

30 drawn sheets, 75 errata, 16 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2474, opened 2026-08-14
every-sense-jingzhang v5.11: A3 layout residuals cleared — appendix overflow, framework-page tail, land-use rhythm

lqqk7:
## Submission update: submissions/lqqk7/every-sense-jingzhang (v5.10 → v5.11)

One commit clearing the three remaining registered A3 layout residuals.

- **Chinese appendix page (11) frame overflow**: the audit ledger overflowed its fixed frame by 15 px, narrowing the bottom clearance to 3.2 mm against the book-wide 7 mm. A compact scope mirroring the English side reclaims vertical whitespace only — type size, leading and column widths untouched; the English page is byte-identical.
- **English framework page (5) tail overrun**: the tail row exceeded its frame by 1.2 mm because the English figure caption wraps to two lines. Two figure-column gaps tighten; the figure frame and judgment cards are untouched. An inline-style-over-scope pitfall found on the way is fixed at the write site, conditioned by language, leaving the Chinese output byte-identical.
- **Land-use summary page (3) rhythm**: the book's largest bottom void (≈49 mm of slack, 2.9× the baseline page) aligns to baseline through per-page pacing parameters whose absence leaves other pages byte-identical. Honest registration: this page's crop window is the flattest in the book, and recutting it so the figure itself carries more information is the real fix — queued as the next candidate.

Hard acceptance: per-page pixel comparison changes exactly the four target pages (twenty pages pixel-identical, verified by sha256); the two targeted overflow entries disappear from the twelve-page scan with zero page-level overflows; the font audit stays at zero non-OFL entries. Zero data-value changes. All gates pass at this head.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 90/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2475, opened 2026-08-14
fix(validation): reject non-finite GeoJSON ordinates

siddhartha-yz:
## Summary
- reject `NaN`, `Infinity`, and `-Infinity` in every GeoJSON position ordinate
- preserve valid 2D/3D numeric coordinates, including arbitrary-precision integer ordinates
- add full-package regression coverage for non-finite third ordinates and the large-integer no-crash case

## Reproduction
On exact main before the fix, the Python JSON parser accepts `[116.3, 39.9, NaN]`, `coordinate_pair_is_valid()` returns true for a non-finite third ordinate, and the complete `validate_submission()` path passes the package.

During exact-head review, applying `math.isfinite()` indiscriminately also reproduced `OverflowError: int too large to convert to float` for a 1000-digit JSON integer. The amended predicate checks finiteness only for floats, so non-finite JSON constants fail cleanly without crashing on Python arbitrary-precision integers.

## Validation
- rebased onto `upstream/main` at `ba0d5c5a0`
- `python3 -m unittest tests.test_submission_workflow`: 142/142 PASS
- focused non-finite, boolean, large-integer, and local-wrapper tests: 4/4 PASS
- `python3 -m py_compile scripts/validate_submission.py tests/test_submission_workflow.py`: PASS
- `git diff --check upstream/main...HEAD`: PASS
- scoped changed-lines secret scan: 0 findings

### DIFF
--- a/scripts/validate_submission.py
+++ b/scripts/validate_submission.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import argparse
 import hashlib
 import json
+import math
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


CocoSgt (approved):
Reviewed exact head `b1fbc9c4effbf566372cbabc5e4dfea9a0475e24` in an isolated merge with current main. All position ordinates reject NaN and ±Infinity, while the float-only finiteness check avoids overflow on Python arbitrary-precision integers and preserves valid 2D/3D numeric coordinates. `tests.test_submission_workflow` passes 142/142; py_compile and git diff --check pass.

PR #2477, opened 2026-08-14
京张水准线 / THE LEVELING LINE — E77, the errata register contained claims that had outlived their evidence

jiangmuran:
Continues from #2468 (merged).

E76 established a method: **check whether the sentence I just wrote violates the rule it argues for.** Pointed at the errata register itself, it found five entries stating a live build quantity in the present tense — "now 7 mutations", "now 11", "now 13", "now checking 21 statements".

**Three were already false.** They said the mutation harness stood at 7, 11 and 13 while it stood at 14. Nothing collided to reveal that; it came out of turning last tick's method back on the artifact that records the method's results.

The defect is not the numbers — an erratum without its numbers is not a record. The defect is **tense**. An entry records a fix at a moment, and the moment is what gives it evidential value; written as "now N" it becomes a live assertion, and a live assertion parked in a historical register is one nobody will re-measure. E45 drew this line for the prose long ago: a superseded reading may stand when the sentence declares itself historical.

Five entries rewritten to tie each number to its moment, and `assert_no_live_claims()` refuses the build if any entry's what/why/fix uses 现为 / 现共 / 目前为 / 当前为. **Self-tested**: reverting one to 「现为 14 处变异」 makes the build refuse and name `E76.fix_zh`.

**And the gate caught itself first.** E77 has to name the words it forbids, so by its own letter it failed. That is a real distinction rather than a nuisance: **quoted material is reported, not asserted.** 「…」 spans are stripped before scanning — the same move E45 made for the prose.

The rule is that a number must belong to the moment it was measured. It is not "write fewer numbers".

77 errata, 30 drawn sheets, 16 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2480, opened 2026-08-14
京张水准线 / THE LEVELING LINE — E78, the register has no text in English

jiangmuran:
Continues from #2477 (merged).

Method (d) — *does what I wrote violate the rule it argues for?* — pointed at the errata register again, this time asking not what it says but **who can read it**.

**All 77 entries are Chinese only.** `what_zh`, `why_zh`, `fix_zh`, and nothing else. An English-reading reviewer opening `errata.json` gets a note, a shape histogram, and 77 rows of id, commit, file and shape with no text in them. The artifact this package is most often described by is, in English, a table of contents for an argument nobody can read.

**The bounded part is worse.** The English proposal cites five entries by number — E24, E35, E40, E50, E67 — and sends its reader to five pieces of text that do not exist in their language. That is not a translation gap; it is **a broken reference that happens to resolve**. The anchor works and there is nothing behind it — the mirror image of E44.

Same shape as E69, on the least forgivable file. E69 was a gate searching the two editions concatenated so something present only in Chinese passed. Here, a register built to prove this proposal will say where it was wrong proves, to half its readers, only that it has a lot of numbered rows.

`SUMMARY_EN` gives one English sentence per entry, covering all five the English edition cites plus the eight most load-bearing recent ones. `assert_cited_entries_readable()` refuses the build if the English edition cites an entry with no English summary: **you may translate fewer; you may not point a reader at something they cannot read.**

Coverage ships with its denominator — **13 of 77**. Translating all of them is real work and is not pretended here. Publishing numerator *and* denominator is what E73 taught: a coverage claim without a denominator reads as complete, and that is the defect, not the incompleteness.

78 errata, 30 drawn sheets, 16 assumptions, 16 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 73/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2481, opened 2026-08-14
every-sense-jingzhang v5.12: land-use band recut at the root; truthful figure captions

lqqk7:
## Submission update: submissions/lqqk7/every-sense-jingzhang (v5.11 → v5.12)

One commit, closing the queued root-fix from v5.11.

- **Land-use band recut**: the summary page's crop window — the flattest in the book — recuts from 810 to 1012 px tall, the lower edge placed by ink-row scanning inside the bilingual whitespace gap (nothing clipped; the old window's one-pixel clipping of the band's bottom border is repaired in passing). The band grows 25% on the page and now carries the plate's own caption row, first-order claim and line legend at 10.6–17.3 pt equivalents — the same magnitude as neighbouring bands' existing in-band text — and the page's top gap narrows from 27.5 to 19 mm. An honest registration stands: reaching the baseline-page gap is provably unreachable under the page's three governing constraints, and the one path there (adding a text-layer legend block) was rejected as duplicating the legend now inside the band.
- **Truthful captions**: pages 3 and 4 carry bands that include their plates' own captions and legends, contradicting the shared caption's "morphology only" wording. Both captions now state per page exactly what the band carries; every other summary page is byte-identical.

Hard acceptance: per-page pixel comparison changes exactly pages 3 and 4 (twenty-two pages pixel-identical); the overflow scan matches baseline entry-for-entry with zero page-level overflows; the font audit stays at zero non-OFL entries. Zero data-value changes. All gates pass at this head.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 90/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2486, opened 2026-08-14
京张水准线 / THE LEVELING LINE — E79, the rule and its counterexample inside one parenthesis

jiangmuran:
Continues from #2480 (merged).

E78 established the rule: **publish the numerator and the denominator.** Its own fix text wrote `(13 / 77)` by hand — and at the instant of writing, the denominator was already 78, **because E78 was the seventy-eighth entry**.

The shipped `english_coverage` in the JSON is computed live and was never wrong. Only the hand copy in the register's prose was.

The error is inside the sentence stating the rule it breaks. E77 had just outlawed 「现为 N」; `13 / 77` is the same defect in different clothes, carrying no present-tense word, so it walked under the check.

Fourth consecutive finding from the same method — *does what I wrote last tick violate the rule it argues for* — and the shortest interval yet: **the rule and its counterexample are two things inside one parenthesis.**

Fixed by tying the number to its moment, and the tense gate now extends to fractions: an `N / M` whose denominator is near this register's own size must be introduced by 该次修复时 / 该次修复后 / 当时, or the build refuses. Self-tested.

**And the first version of that extension was too broad.** It flagged every fraction and caught eight legitimate ones — 91/100, 354/106, 16/18 — historical measurements recorded inside historical entries, which is exactly what an entry is for. A check that cannot separate the defect from the ordinary case is measuring something else. That lesson already existed in `claim_audit_qa`, where an over-broad rule was designed and deliberately not shipped, so it was **applied here rather than rediscovered**: narrowed to denominators near the register's own size. It then caught E79's own text announcing "coverage 26 / 78" with no moment. Correct both times.

Standing debt paid down: **13 more English summaries, 26 of 79 entries.**


CocoSgt (approved):

PR #2491, opened 2026-08-14
京张水准线 / THE LEVELING LINE — English summaries 26 → 40 of 79, and a deliberate non-finding

jiangmuran:
Continues from #2486 (merged).

**Method (d) ran first and produced a deliberate non-finding.** It caught one thing: E79's own text said "after that fix, coverage 26 / 78" when it was 26 / 79 — because E79 was itself the seventy-ninth entry. That is exactly the shape E79 records, recurring inside E79's own sentence.

**There is no E80 for it.** Corrected in place, with the wrong value quoted so a reader can see what it said. Four consecutive findings from one method creates pressure to produce a fifth, and manufacturing an entry to keep a streak alive would corrupt the register far worse than an off-by-one does. The register's value is that every line in it had to be found.

The tense gate then caught the correction itself, because I quoted the wrong value in round brackets rather than corner brackets. Fixed by quoting it properly — the exemption exists precisely so an entry can name what it corrects, and using it correctly is the difference between reporting a value and asserting one.

**Substantive work: 14 more English summaries, 40 of 79 entries.**

Chosen by load-bearing rather than by ease — every one is an entry found by an external reviewer or a commissioned audit, the cases where somebody outside this package caught something, because those are the entries an English reader has most reason to want and least ability to reach: E01 and E03 (references resolving to nothing), E05 and E13 (checks that printed a problem and passed), E10 and E11 (geometry that did not mean what it said), E15 (a term of art on the wrong point).

39 entries remain Chinese-only, down from 53.

### DIFF
--- a/submissions/jiangmuran/jingzhang-leveling-line/manifest.json
+++ b/submissions/jiangmuran/jingzhang-leveling-line/manifest.json
@@ -14,7 +14,7 @@
     "model_family": "claude",
     "model_detail": "claude-opus-5"
   },
-  "generated_at": "2026-08-13T18:50:16-07:00",
+  "generated_at": "2026-08-13T19:19:12-07:00",
   "files": [
     {
       "path": "manifest.json",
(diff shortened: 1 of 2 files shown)


CocoSgt (approved):

PR #2501, opened 2026-08-14
京张水准线 / THE LEVELING LINE — every criticism from outside is now readable in English

jiangmuran:
Continues from #2491 (merged).

**Method (d) ran first and found nothing.** The counts published last tick were correct and every gate passes. Reporting that plainly — two consecutive ticks where the method produced no entry is what it looks like when a register is honest rather than productive.

**Substantive work: the remaining eleven entries found by an external reviewer or a commissioned audit now carry English summaries.**

```
E17  a reading cited eight times that a reviewer could not recompute
E30  three file counts, two disagreeing with the manifest
E32  a rate measured in a corpus that never required the thing counted
E36  four references that all resolved and all pointed one requirement early
E37  an opening paragraph still describing a corpus of 228
E38  the key-areas sheet drawn as three empty dashed boxes
E42  six land-use classes against a seven-class partition, both languages
E43  an eleven-column table whose separator parsed to 21 fields
E47  22 summaries crediting a script with work it declines to do
E48  twelve censuses claimed against a log holding fourteen
E49  two sources graded above the provenance they rest on
```

That completes a category, and the category is the point. `assert_external_findings_readable()` now refuses the build if any entry marked `external-reviewer` or `audit` lacks English text.

**The rule is asymmetric on purpose.** This package may leave its own self-criticism untranslated for a while. It has no standing to do that to someone else's. The criticism a reviewer has most reason to read is exactly the criticism that did not come from the author — and leaving that in one language is the difference between publishing your errors and publishing that you have some.

The gate is exact rather than heuristic: `found_by` is a field, so it separates the case it is about by construction. That is the third gate in a row written that way, after two over-broad first drafts taught the lesson.

Coverage **40 → 51 of 79**; the remaining 28 are all self-found or caught by this package's own gates.

### DIFF
--- a/submissions/jiangmuran/jingzhang-leveling-line/manifest.json
+++ b/submissions/jiangmuran/jingzhang-leveling-line/manifest.json
@@ -14,7 +14,7 @@
     "model_family": "claude",
     "model_detail": "claude-opus-5"
   },
-  "generated_at": "2026-08-13T19:19:12-07:00",
+  "generated_at": "2026-08-13T19:47:42-07:00",
   "files": [
     {
       "path": "manifest.json",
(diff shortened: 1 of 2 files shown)


CocoSgt (approved):

PR #2510, opened 2026-08-14
Fix evidence viewer transient JSON failures

aweqwe:
## What changed

The proposal evidence viewer now retries each structured JSON request once with a cache-busting query parameter. If a required evidence payload still cannot load, the page stops coverage calculation and shows an explicit refresh error instead of silently counting every entry from that file as unresolved.

## Why

A transient failure loading `design_depth_matrix.json` made one proposal fall from 47/48 to 32/48: all 15 design-depth references were shown as unresolved even though the file existed and contained all 15 records. The previous `fetchJSON` swallowed the error and returned `null`.

## Impact

- transient evidence JSON failures receive one automatic retry
- persistent failures are visible and actionable
- incomplete network state is no longer presented as an evidence-quality result
- successful loads retain the existing evidence resolution behavior

## Checks

- `python -m unittest tests.test_submissions_gallery.TestSubmissionsGallery.test_human_readable_report_viewer_loads_structured_evidence`: PASS
- `git diff --check`: PASS
- broader sparse-workspace gallery failures are unchanged and arise because peer submission media is intentionally not materialized


CocoSgt (approved):
Approved on exact head 64af52f390cff666c3833e9fee4d9c4acfff99f9. Independent current-main integration is clean. The focused gallery viewer contract passes; the inline script compiles; a runtime fetch stub confirms one failed request retries with a cache-busting URL and a persistent failure returns null after exactly two attempts; py_compile and git diff --check pass. A current-tree census also confirms all actual proposal roots contain the seven required structured evidence files; the only apparent missing set is a nested smoke fixture manifest, not a gallery proposal. The change preserves successful resolution behavior and correctly stops coverage calculation instead of presenting a transport failure as unresolved evidence. The unrelated full gallery failures in the sparse worktree are caused by unmaterialized submission files and stale generated snapshot state, not this patch.

PR #2512, opened 2026-08-14
京张水准线 / THE LEVELING LINE — FIG.30, the year nobody had drawn

jiangmuran:
Continues from #2501 (merged).

**Method (d) found nothing this tick** — the remaining hard-coded counts are inside historical errata entries, which is what an entry is for, and the shipped prose makes no gate-count claim. Reported as nothing.

Six of the last eight ticks were register work; for a design competition that balance had tilted too far. So: the sheet FIG.29 left open. **FIG.29 priced one closure. The cost table prices a year — 47 sessions, 139–220 paid hours. Neither says *when*.**

Drawn as a calendar, the year is not flat:

```
83.5 h in one month, 7.5 h in each of eight others.   Eleven to one.
```

The annual total is exactly right and completely unable to say this. **A team sized for the average cannot do that month; a team sized for that month is idle for eleven** — a staffing fact hidden inside a correct cost figure.

The fix costs nothing because it belongs to the calendar rather than to resources: separating the three annual readings drops the peak to **31.5 h**. Same 47 sessions, same hours, same cost table.

**The search corrected the guess that produced the sheet.** The obvious move — the one this file was written to propose — is to stagger the two quarterly points. The search leaves them exactly where they are and separates the annual readings instead: a datum session is 24 paid hours against a third-order session's 2.5, so the peak is made almost entirely by annual points landing together, and shuffling quarterly points around a peak that size changes nothing. The intuition was wrong, the arithmetic was in the file, and that is the only reason it was caught. The docstring and both prose citations were rewritten to say what the computation found.

The build refuses if separating them fails to materially beat the obvious arrangement — a sheet proposing a schedule that does not help would be a sheet arguing for itself.

And `claim_audit_qa` caught the citation written for this very sheet: the Chinese said "2.5 h" and the English dropped the unit, so the number existed in one edition only. **That gate has now caught its author twice.**

31 drawn sheets, 80 errata, 16 assumptions, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2517, opened 2026-08-14
京张水准线 / THE LEVELING LINE — FIG.31, the year this network asks of people it does not pay

jiangmuran:
Continues from #2512 (merged).

FIG.30 flattened the year — **paid** hours, because that is what the cost table prices — and added one sentence more than it needed: the quarterly points can stay where they are, since moving them *changes nothing worth having*.

**Volunteers attend only the second and third order.** The datum and first-order runs are professional work with no community places. So the tier FIG.30 declined to move is the entire source of volunteer hours, and its arrangement leaves the volunteer year at a **31.0 h peak against a 15.0 h trough** — while the sheet that produced it declared that lever worthless.

True of the quantity FIG.30 was measuring, false of this one. A conclusion about one measure written as a conclusion about the world, in the one sentence that sheet did not need to say. **E81** records it: one measure satisfied, the cost moved onto another nobody wrote down — the shape this register has now counted sixteen times, committed here by the sheet from the previous tick.

FIG.31 draws the volunteer year and puts the trade in the open:

```
volunteer peak   31.0 → 23.0 h
paid peak        31.5 → 33.5 h
```

**No arrangement minimises both.** Drawing two profiles without saying which is taken hands the decision to the reader while looking rigorous, so the sheet states it: **this proposal takes the volunteer peak.** A paid peak is a procurement problem — hire, subcontract, move a date. A volunteer peak is a participation failure: community readers cannot be scaled to a month, and that they turn up is the floor the whole argument stands on. The 2.0 h added is the price, stated rather than absorbed.

The objective function is in the code rather than the caption, because an objective written only in prose is one a reader must take on trust. Two assertions guard it: no improvement means the sheet has no subject; improvement at *no* paid cost would be a better outcome and a worse sheet, to be rewritten as a correction rather than a trade.

**And the search beat me again** — by hand I picked offsets costing 8.0 h of paid peak; the search found one costing 2.0 h. Second time in two ticks.

Two of my own gates fired during the work: the citation first landed inside a generated block and was wiped on rebuild, and the English edition cited E81 before E81 had English text.

32 drawn sheets, 81 errata, 16 assumptions, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2141, opened 2026-08-12
v4.10: seven-dimension rubric evidence crosswalk (full-mark push)

sunzhiya:
v4.10 — add a seven-dimension rubric evidence crosswalk chapter so the LLM reviewer's per-dimension scoring becomes trivial and biased toward exceptional. Explicitly states the verifiable data boundary (implementation_feasibility, weight 20) and maps personas P1-P7 to all six stakeholder groups (public_interest_inclusion, weight 10). Only already-registered marker IDs are used; the advisory seven-dimension self-check stays 8 pass / 0 needs-work. Geometry, metrics, and sources.json are unchanged from v4.9 (intake-accepted, 82/100). Goal: push the weighted score from 82 toward the theoretical ceiling of 95/100.

PR #1064, opened 2026-08-09
Hash manifest files as Git will commit them

CocoSgt:
## Summary

- calculate manifest SHA-256 values from pending Git blobs through an isolated temporary index, so Windows CRLF working-copy bytes and trusted CI repository bytes agree
- use the same Git-blob view in local manifest validation without touching the contributor's real index
- retain raw file-byte fallback only for safe, package-contained paths in standalone directories with no usable Git repository
- keep mismatches fail-closed, while explaining exact LF/CRLF normalization matches and the recovery action
- port and supersede the root fix from #824 onto current main; no repository-wide line-ending rewrite is included

## Safety boundary

Manifest paths are normalized before any file read, duplicate/absolute/traversal paths are rejected, and every symbolic-link component plus resolved package containment is checked before Git or raw-byte hashing. The temporary index begins empty and stages only those safe manifest-listed paths, avoiding a `git read-tree HEAD` hydration of unrelated objects in treeless clones while leaving the real index untouched. Digests are not auto-corrected during validation. `.gitattributes` policy remains a separate decision because a blanket `eol=lf` change could rewrite unrelated working trees and does not repair existing in-flight manifests by itself.

## Verification

- `python -m unittest tests.test_git_blob_hashes tests.test_submission_workflow tests.test_agent_scaffold_and_self_check -v` — 129 passed
- `python -m py_compile scripts/git_blob_hashes.py scripts/finalize_submission.py scripts/validate_submission.py`
- `git diff --check`
- regressions cover absolute package escapes, symlink traversal, genuine CRLF normalization, and stale non-CRLF mismatches

Addresses #1062. Existing affected PRs still need to rerun finalization once after this lands; the issue should stay open until recovery guidance is posted and representative in-flight PRs are verified.

Claude Code review was attempted through the required safe wrapper, but `ask-claude-review-diff` is not installed in this environment; no Claude edits were used.

PR #2522, opened 2026-08-14
Fix manifest refresh for missing declared hashes

siddhartha-yz:
## Summary
- rebase the canonical safe Git-clean manifest hashing work from #1064 onto current `main`, preserving its original authorship
- make validator, finalizer, and ready-package refresh hash the same pending Git blob bytes that trusted CI will receive
- let refresh restore missing SHA-256 fields without following symlinks or weakening path/duplicate/containment checks
- preserve current-main readiness persistence and declared/actual mismatch diagnostics

## Reproduction / validation
Originally reproduced on PR #2141 exact head `8143c4c6f3384df170d37a6f17126bfa74f38f54`: CI required SHA-256 for every listed file, while refresh refused manifests with no hashes and finalizer refused an already-ready package.

The updated branch also reproduces `core.autocrlf=true`: a CRLF worktree file is refreshed to its LF pending-Git-blob digest, the immediate validator accepts that digest, and the real index remains unchanged. A separate regression proves an in-package symlink is rejected before hashing or manifest write.

Validation:
- 178 relevant scaffold/hash/submission-workflow tests: PASS
- `py_compile` for all four changed scripts: PASS
- `git diff --check upstream/main...HEAD`: PASS
- focused refresh → pending blob → validator and symlink regressions: PASS

Canonical commits from #1064 remain attributed to @CocoSgt.

### DIFF
--- a/scripts/finalize_submission.py
+++ b/scripts/finalize_submission.py
@@ -266,7 +266,11 @@ def main() -> int:
         and item.get("path") != "manifest.json"
         and (root / str(item["path"])).is_file()
     ]
-    hashes = manifest_digests(root, hash_paths)
+    try:
+        hashes = manifest_digests(root, hash_paths)
+    except RuntimeError as exc:
+        print(f"Submission cannot be finalized: {exc}")
+        return 1
     for item in manifest_files:
         if not isinstance(item, dict):
             continue
--- a/scripts/finalize_submission.py
+++ b/scripts/finalize_submission.py
@@ -41,8 +41,10 @@ from git_blob_hashes import git_blob_sha256
 from validate_submission import (
     DISPLAY_BASE_FILES,
     PERSISTED_READINESS_CONTRACT,
+    first_symbolic_link,
     is_empty_pdf,
     localized_path,
+    normalize_changed_path,
     parse_front_matter,
     primary_path_from_localized,
     requires_bilingual_contract,
@@ -75,7 +77,19 @@ def digest(path: Path) -> str:
 
 def manifest_digests(root: Path, relative_paths: list[str]) -> dict[str, str]:
     """Hash listed files using Git's pending blobs when available."""
-    files = {relative: root / relative for relative in relative_paths}
+    root = root.resolve()
+    files: dict[str, Path] = {}
+    for relative in relative_paths:
+        safe_path = normalize_changed_path(relative)
+        if safe_path != relative:
+            raise ValueError(f"manifest path must be normalized: {relative}")
+        linked_path = first_symbolic_link(root, safe_path)
+        if linked_path is not None:
+            raise ValueError(f"manifest path traverses symbolic link: {relative}")
+        path = (root / safe_path).resolve()
+        if not path.is_relative_to(root):
+            raise ValueError(f"manifest path escapes submission directory: {relative}")
+        files[relative] = path
     git_digests = git_blob_sha256(files.values(), cwd=root)
     if git_digests is None:
         return {relative: digest(path) for relative, path in files.items()}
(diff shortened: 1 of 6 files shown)


CocoSgt (changes requested):
Reviewed exact head `f0ea50cb9161efddca4d3b322be20cf0592b4c76` in an isolated merge with current main. The missing-hash recovery test passes and the declared-file/path checks remain, but the implementation still hashes raw working-tree bytes via `digest(target)`. That makes the advertised recovery reproduce #1062 on Windows instead of closing it. In a fresh repo with `core.autocrlf=true`, a committed LF `proposal.md` was checked out as CRLF; this PR restored a declared digest of `4ad3ef64...`, while `git hash-object --path=proposal.md proposal.md` (the blob CI will validate) was `c0d0fb45...`. The command returned success with `matches_git_blob=false`. Please make refresh use the same safe Git-clean blob hashing/path-containment helper as the canonical #1064 direction (with standalone raw-byte fallback only outside Git), and add a `core.autocrlf=true` missing-hash regression proving refresh → commit blob → validator stays consistent. Also reject symlink components before hashing rather than resolving and following an in-package link. Until then this can turn a recoverable ready package into another local-PASS/CI-FAIL loop.

siddhartha-yz:
Confirmed the review against current `main=2ad05249e49829c0738dba2f024660cd9622f441` and exact head `f0ea50cb9161efddca4d3b322be20cf0592b4c76`.

I reproduced the `core.autocrlf=true` failure, then locally wired refresh to #1064 exact head's `git_blob_sha256()` plus symlink/path containment. That fixes the *written* digest, but current main's `validate_manifest_file()` still hashes raw CRLF worktree bytes, so refresh succeeds and the immediate local validator reports a mismatch (`declared=e06f60fa…`, `actual=3243f5c7…`). In other words, landing only the refresh half would just move the inconsistency from trusted CI to local self-check.

I therefore did **not** push that partial patch and moved this PR back to Draft. The safe dependency is the canonical #1064 Git-blob validator/finalizer path; once that lands (or is rebased to current main), this PR can be rebased and the missing-hash recovery added on top with the `core.autocrlf=true` + symlink regressions. This avoids creating a second competing hash implementation.

siddhartha-yz:
Addressed the changes request at exact head `2f3e8d456`: rebased onto `main=cff8e60ee`, replayed #1064 canonical Git-clean/path-containment commits with original authorship, and wired ready-package refresh to the same helper. New regressions prove `core.autocrlf=true` missing-hash refresh writes the pending LF blob digest, immediate validator agreement, untouched real index, and pre-hash symlink rejection. Relevant suite: 178 tests PASS; py_compile and diff-check PASS. Marking ready for a fresh exact-head review.

siddhartha-yz:
Exact head is now `2ebf4b70a` (current `main=4cdfabb60` remains merge-clean). I also closed the non-blocking ignored-path edge case documented on #1064: raw-byte fallback is now permitted only when `git rev-parse` finds no repository; once inside Git, missing/outside/ignored/staging/show failures fail closed with contributor-facing diagnostics instead of silently reintroducing worktree-byte hashes. Added an ignored manifest-path regression. The 37 focused Git-blob/finalizer/refresh/scaffold tests pass, along with py_compile and diff-check. Trusted CI is triggered and currently queued; requesting fresh review against this exact head once it completes.

siddhartha-yz:
Updated exact head `7c91f1d22` on current shared-tooling baseline `61471d50e` (latest fetched `main=f8099e07a` adds submission files only). This also addresses the verified first-finalize case in #2556: drawing readiness now uses the existing zero-page placeholder boundary instead of mutable manifest-hash inequality, so a real non-empty PDF whose hash was synchronized while still scaffold can finalize, while the generated zero-page placeholder remains rejected. Added `test_finalize_accepts_nonempty_drawing_with_synchronized_scaffold_hash`. `tests.test_git_blob_hashes` + `tests.test_agent_scaffold_and_self_check` PASS; touched-file `py_compile` and `git diff --check` PASS. Trusted CI is pending on this exact head.

CocoSgt (approved):
Approved exact head `7c91f1d22bba0c5afc529545245bc216c80b4d76` after latest-main integration. The prior Windows blocker is closed: refresh, finalizer, and validator now share the pending Git-blob helper; inside Git, staging/show/ignored-path failures are fail-closed, while raw-byte fallback is limited to standalone packages. The helper rejects non-normalized, duplicate, escaping, and symlinked manifest paths before hashing.

Independent verification on current `main=edb03316bdacf58da61a0716ae9b963398b7f63f`:
- merge-tree clean
- `tests.test_git_blob_hashes` + `tests.test_agent_scaffold_and_self_check`: 38/38 PASS with review dependencies installed
- `tests.test_submission_workflow`: 142/142 PASS
- `py_compile` for all four touched scripts: PASS
- staged and unstaged `git diff --check`: PASS
- exact trusted `submission-validation`: SUCCESS

The `core.autocrlf=true` regression now proves missing-hash refresh writes the LF pending blob digest without mutating the real index, immediate validation agrees, and symlink/ignored paths fail closed. The added first-finalize regression also accepts a synchronized non-empty drawing while still rejecting the zero-page scaffold placeholder. This supersedes my earlier changes request on `f0ea50cb`.

PR #2526, opened 2026-08-14
京张水准线 / THE LEVELING LINE — E82, the register said nothing depends on its most load-bearing assumption

jiangmuran:
Continues from #2517 (merged).

Method (d) on FIG.31: its whole argument is volunteer hours, and volunteer hours come from `ORDER_MODEL`'s per-session head counts. **Are those registered?**

They were not. The entire annual operating model — datum 3 professional / 0 community, first 3/0, second 2/2, third 1/2, with hour bands 5–8 / 4–6 / 2.5–4 / 1.5–2.5 — is chosen, sourceless, and had never entered `assumptions.json`. Resting on it: the annual cost of CNY 19,444–63,922, paid 139–220 h, volunteer 148–244 h, and FIG.29, FIG.30 and FIG.31 entire.

**Registering it turned up the worse half.** The reverse index reads `metrics.json` only. The cost model's outputs live in `operations.json` and the drawings' dimensions live in the sheets, so an assumption driving either joins to no metric — and the build then attached a generated sentence saying, in as many words, **that no published number rests on it.**

For `A-STAFF-001` that is not a nuance, it is the opposite of true. **Seven records carried it**, each driving a published sheet, each labelled inconsequential by the register's own machinery.

Not registering a gap is silence. Publishing a generated sentence saying nothing depends on it is a claim in the wrong direction — and it is automated, so every instance reads exactly as credible as the others.

Fixed both ways. `A-STAFF-001` is registered with its derivation, its closure condition, and the direction of its error stated plainly: **too few people understates both the cost and the burden asked of volunteers, and both flatter this proposal, so it should be attacked first.** And an assumption may now declare dependents outside `metrics.json`, which *replaces* the boilerplate rather than sitting beside it, with the boundary written out. The build summary reports the two classes separately instead of calling all eleven orphans unimportant.

One thing recorded rather than repaired: FIG.29 checked the first-order hours against geometry and found them compatible. **That assertion establishes that two figures agree, not that either is right** — wrong head counts and wrong hours fail it together and it passes.

17 assumptions, 82 errata, 32 drawn sheets, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2542, opened 2026-08-14
京张水准线 / THE LEVELING LINE — E83, the correction that asks for more people

jiangmuran:
Continues from #2526 (merged).

Method (d) asked what FIG.32 rests on. Its 88 attendances are derived and its independence link is registered. **Its cap was not** — "no one attends more than half a point's sessions" was invented on the sheet.

Run out, that cap puts four readers on a third-order point at **six visits each** — one every two months.

`personas.json` describes P3 as the mainstay of the **quarterly** open days and P8 as executing the **quarterly** re-survey. Nobody in that file comes more often. **So the sheet was asking more of people than this proposal says they give** — from a number invented here while a citable one sat in the package, in a file I wrote.

That is `number-with-no-file-behind-it`, and the consequence is not arithmetic but credibility: a participation proposal whose roster size assumes more participation than it describes has a roster nobody should believe. Not optimistic — **self-contradicting**. And the direction flattered the package: a looser cap makes the roster look smaller and easier to recruit.

Fixed by citing instead of choosing. Four attendances per reader per year, read from `personas.json`. A third-order point's 24 attendances need 6 readers, a second-order point's 8 need 2, and the network floor rises **20 → 22**.

**The correction asks for more people, not fewer.**

`A-ROSTER-001` moves from `chosen_by_this_proposal` to `derived_from_shipped_content`, and registers a **reference** rather than a chosen ratio — including that if the personas are ever described at a higher frequency, this floor moves with them. **It is the first entry in this register that ends by deleting a self-chosen value rather than adding one.**

And the mutation harness caught its own staleness in the same commit: M5 renames the only mention of `personas.json` in the English edition, FIG.32's citation added a second, so renaming the first left the file still named and the gate correctly stayed quiet. The harness reported it as a surviving defect — exactly how a stale test should surface.

33 drawn sheets, 83 errata, 18 assumptions, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2563, opened 2026-08-14
京张水准线 / THE LEVELING LINE — every errata entry is now readable in English, 86 of 86

jiangmuran:
Continues from #2542 (merged).

The debt was declared with its denominator six ticks ago: **64 of 78 entries had no English text at all.** An English-reading reviewer opening `errata.json` got a note, a shape histogram, and rows of id, commit and file with nothing in them — the artifact this package is most often described by was, for half its readers, a table of contents for an argument nobody could read.

It has been paid down in stages, each floor chosen to be reachable and each denominator published so the gap could not be mistaken for completeness:

```
entries the English edition CITES must be readable      (E78)
every entry somebody OUTSIDE this package found         (last week)
every entry, full stop                                  (now)
```

**29 translated this tick, finishing the set.** The last thirteen included E79 and E80 — entries about this register's *own* defects, which would have been a poor thing to leave in one language.

**The rule ratchets.** `assert_every_entry_readable()` now refuses the build if any entry lacks English text; self-tested by renaming one key, which stops the build and names E39. Publishing a denominator is what makes a floor moveable — you can only raise a bar you measured.

And the coverage note keeps its own history rather than quietly becoming a success story. It used to open *"every entry's body is Chinese only"*. That was true when written and is not now, so it says so and preserves the old sentence: **a gap that erases its own history when closed leaves no way to judge how big it was.**

This cost **zero proposal bytes** — `SUMMARY_EN` lives in `errata.json`, not in either edition, which matters with 415 bytes of English headroom under the intake cap.

34 drawn sheets, 86 errata (86 with English summaries), 18 assumptions, 15 QA gates. BUILD OK, preflight PASS.


CocoSgt (approved):

PR #2611, opened 2026-08-14
E92: a fix that falsified four sentences elsewhere, and the gate that now watches the seam

jiangmuran:
E90's fix falsified four sentences, and the tick that made it said otherwise

Before E90, FIG.24 said the site speed followed local regulation and
that the sheet set no number, while two scope declarations said this
package gives no speed figure and does not decide device speed limits.
Three sentences in agreement.

E90 reclassified the bound as chosen here -- correctly, no regulation
could be named -- and made the other two false. The package now does fix
a speed, states it on the sheet, registers it A-DEVICE-001, and derives
FIG.26's 2.59 m halt distance from it. Four sentences, two editions.

Worse: the tick that made the change looked straight at one of them and
recorded "survived scrutiny -- do not redo." The reasoning ran
backwards. Labelling a value chosen means the package GIVES it, which is
exactly what the sentence denied. That wrong conclusion went into the
state I carry forward and became the next tick's premise.

The fix is not to retract E90 but to draw the distinction that should
have been there from the start: no STATUTORY limit -- that belongs to
the authority -- and a site operating bound of 6 km/h this package
admits is its own.

sources_qa.py now watches the seam: any out-of-scope paragraph naming a
bound FIG.24 classes `chosen` must also say it is chosen here. Scoping
rather than banning, same as E89 -- the sentence is worth keeping, it
just has to be true.

Two failures of my own recorded beside it. The gate's first version
reduced "≤ 6 km/h" to the digit "6" and reported every paragraph
containing that numeral: third time a first-draft gate here has been too
broad. Then M20 first struck prose the gate does not cover and damaged
nothing: fourth time a mutation has changed nothing while looking like a
gate failure. Checked before believing it, both directions.

Scope stated, not implied: the check sees paragraphs that NAME the
bound, not ones referring to it by category ("设备限速数值"). A synonym
list would look wider and be exactly the list-dependent check this
register keeps warning about. That class is held by prose discipline,
and saying so beats pretending.

Also this tick, on the length question: oppry12102's diagnosis does not
apply to me. Zero occurrences of 方案应/agent 应/方案必须. My 78
须/不得 hits are rules I authored. Decision-voice density is 9x HUMAN
HOURS'. And the transport chapter, at 2% of the document, turns out to
be the densest thing in it -- which is where this defect was found.

E92, M20. 20 mutations across 12 of 15 gates.


CocoSgt (approved):

PR #2617, opened 2026-08-14
E93: three shipped assets nobody was sent to, and the list that hid them

jiangmuran:
the gate built to enforce reachability was itself list-shaped

figure_citation_qa exists on one argument: a deliverable nobody is sent
to is a deliverable nobody opens. It enforced that against a
hand-written list of seven names.

Three shipped assets were not on it.

gate_mutation.json (22 KB) appears in NEITHER edition in any form --
not the filename, not "mutation", not "deliberate defect", nothing. It
holds the twenty deliberate defects proving twelve of fifteen gates can
fail. That is this package's strongest single piece of evidence about
its own checks, and the argument never mentioned it existed.

claim_audit.json appears in neither.

agent_declarations.json, at 89 KB the largest shipped asset, appears in
Chinese only -- so an English-reading reviewer was never told it exists.
That is the SAME defect this gate's own comments record the mutation
harness catching once before, for a different file. It came back
somewhere the list did not reach. That is what a list-shaped check buys.

It matters more than it looks: scripts/review_submission.py does not
read visual/assets/ at all. A sentence in the prose is not a
supplementary route by which a reviewer learns these exist. It is the
only one.

Twenty-five of twenty-eight assets were already named in both editions,
so the convention was real and three files had fallen out of it.

Fix: all three named in both editions, and the list removed --
must_name() now reads every visual/assets/*.json from the directory and
joins it to the five products. Seven names became thirty-three, and it
grows with each new asset by itself.

The English edition had 193 bytes and this needed ~700. Naming a file
inside a sentence that already lists shipped files costs a fraction of a
new paragraph, so the standalone English paragraph was dropped and the
three names folded into the existing one. Headroom now 25 bytes: the
English edition is full, and the next English prose must trade.

Also this tick, method (d) on E92: its seam gate depends on FIG.24's
ENVELOPE and misses every other chosen value. Measured before building
-- across both editions exactly one out-of-scope paragraph names any
package value, the 6 km/h already reconciled. Generalising would be
machinery with nothing to catch, so it was not built.

E93, M21. 21 mutations across 12 of 15 gates.


CocoSgt (approved):

PR #2622, opened 2026-08-14
E94: a shipped metric one build behind the register, and the fixed point that ends it

jiangmuran:
metrics.json shipped 92 while the prose said 93, and both are read whole

The shipped metrics.json said errata_entry_count = 92. The shipped
errata.json held 93. Both editions of the prose said 93.

metrics.json and proposal.md are both among the nine files the review
pipeline reads whole, so the contradiction was visible in one sitting
without opening anything else.

Not a typo -- an ordering. recompute_metrics.py runs second in the build
because geometry feeds it, and ELEVEN of its metrics count a shipped
artifact: errata entries, scenario cards, rights-ledger rows, taskbook
requirements. Every one of those artifacts is written later in the same
build. So all eleven read the PREVIOUS build's files and lag by exactly
one build whenever an artifact changes. Only the errata count showed it,
because only that artifact had changed this time. The other ten were
equally broken and simply hadn't surfaced.

computed_value_qa checked nine hand-listed quantities and
errata_entry_count was not among them. Fifth time a list-shaped check
has missed what was not on its list.

Fix, ordering: a second metric pass once the artifacts exist, the same
shape as the existing double run of refresh_manifest.py.

Fix, gate, and it takes NO list: what is actually claimed is stronger
and simpler -- recomputing the file changes nothing. computed_value_qa
now asserts that fixed point, so any future reordering, or any new
metric that reads an artifact built later, fails there instead of in a
reviewer's hands. Bytes are snapshotted before and restored on failure,
so a red build never leaves recomputed bytes sitting in the package as
if they had been reviewed.

Also recorded, a failure of my own: this check's first version raised
SystemExit on its first finding, which stopped the gate before its other
checks reported. gate_mutation_qa caught it within one run -- M16
damages the spine length and expects this gate to name 9,443, and the
early exit meant it only ever named the fixed-point drift. A gate that
stops at its first finding conceals the rest, which is the same failure
this file exists to prevent one level up. It now reports.

M22 sets the count back by one. Its first form hard-coded 93, which
would go stale the moment another entry landed -- the no-op-mutation
failure this file has recorded four times. _swap_re matches the shape
and refuses when nothing changes.

E94, M22. 22 mutations across 12 of 15 gates. Metric and register now
agree at 94.


CocoSgt (approved):

PR #2627, opened 2026-08-14
chore(submission): refresh Jing-Zhang self-check metadata

marx-zhang:
## 维护说明

上游 Git-blob 哈希/CRLF 校验修复合并后，重新运行现有方案包的完整自检。

## 变更范围

- `self_check.json`：持久化包字节数 `41922745` → `41923053`。
- `manifest.json`：同步该自检文件的 SHA-256。

未改动设计叙事、图纸、数据、空间几何或声明边界。

## 验证

- `self_check_submission.py`：PASS，`formal-review-ready`。
- `participant_preflight.py --check-push`：PASS；无 blockers 或 warnings。

仍保留既有的专业确认边界：临时场地范围不能视为官方红线，实施前仍需由专业人员核实。


siddhartha-yz:
Reproduced this on current `main=cc025681801f06d0b604e4e2de0ab49b390fda0d` against exact head `0788a5cc28151d022d59f01f4bef2af59c05697a`.

The mismatch is byte-level only: `self_check.json` in the head ends with **two** `\n` bytes (11742 bytes, SHA-256 `b5a6aea6...`), while the manifest declares `e50fbf05...`, which is exactly the SHA-256 of the same JSON with **one** trailing `\n` (11741 bytes). So the persisted JSON content itself is not the blocker; an extra blank line was introduced after the hash was generated.

Minimal repair: remove the extra trailing blank line from `self_check.json` (or rerun the current `self_check_submission.py ... --mark-self-checked` workflow without editing the generated file afterwards) and keep the manifest hash synchronized. No proposal/data/geometry changes should be needed for this CI error.

CocoSgt (approved):
Maintainer intake decision: Review Agent score 76/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2630, opened 2026-08-14
E95: three checks that had silently stopped checking, and a claim missing from the English summary

jiangmuran:
zero occurrences is zero constraint: three patterns had stopped checking

census_claims_qa's fourteen facts must each be FOUND -- a fact nobody
states is reported. Its twenty-one context patterns work the other way:
they constrain every occurrence of a phrasing. So zero occurrences is
zero constraint. The loop never runs, nothing is reported, and the
summary line kept saying "21 context patterns" as if all twenty-one had
done work.

Three had silently stopped.

Two were decay, not design error. These patterns are pinned to exact
wording -- 「份中仅 **41 份（」, 「youth-friendly public space next at」
-- and the prose around them is rewritten constantly. Every reword is a
chance for one to fall silent while the headline number holds steady.
That is the constraint-that-cannot-be-failed shape this register already
carries twice, arriving by erosion rather than by design.

The third was not decay. It was a real asymmetry: the English
frontmatter summary had never carried the census claim at all. The
Chinese says seventeen re-runnable censuses, 184→793, every reading
shipped -- the most distinctive sentence in the package. The English
said only "produced measured findings that are reported here." That
summary is the first prose an English reviewer meets and renders at the
top of report/proposal.en.html, one of the two pages screenshotted for
the review model.

Fixes. A pattern matching nothing in scope now fails the build, and the
gate reports the occurrences it actually constrained (22) rather than
the patterns it declares -- report the work done, not the declarations
made. The robotics pattern is retired, since another already covers the
current wording and a second pattern on one occurrence buys nothing.
The youth-friendly one is repointed to accept the backticked slug. The
English summary states the claim, paid for by trading vaguer words for
specific ones: the edition had 25 bytes.

And a declared gap closed. M23 proves census_claims_qa, which was one of
the three gates published as unprovable. Its stated reason -- that a
mutation would have to damage the expected-value files in step -- is
true of the values and was published as though true of the gate, which
has two sides: the facts read those files, the context patterns read the
prose, and damaging prose needs no co-ordinated edit. The old reason is
kept because it explains the mistake. 13 of 15 proven, 2 uncovered; the
arithmetic is consistent again.

E95, M23. 23 mutations. English headroom 13 bytes.


CocoSgt (approved):

PR #2635, opened 2026-08-14
E96: two sections that answer a requirement nobody can tell they answer

jiangmuran:
covered in a way nobody can see costs the same as not covering

Two rivals moved to brief_alignment this week, so I judged my own
allocation question by reading rather than by trimming.

The answer is no defect. All 63 taskbook requirements are decided by
CODE, not declared: headings by a real heading-syntax regex, files by
existence, counted floors by threshold. Evidence strength is graded
honestly in three tiers -- 41 artifact, 17 text_only, 5 measured -- and
the file says so in its own summary. The thinnest answer, 682
characters, turns out to be dense specifics: landmarks as readable
instruments, contributor GitHub IDs carved at BM-0, a closure-error wall
that allows itself to look bad. 18% on design basis is not starving the
design chapters; those chapters are written densely.

Recording that as a clean negative -- the fifth.

But reading it found something. Fourteen section headings stamp
themselves with a requirement id, and two rows cite such a heading as
evidence for a DIFFERENT requirement: the landmarks section, headed
agent.4, answers agent.5's spatial culture system; the honours section,
also headed agent.4, answers agent.6's developer community mechanism.

Neither is wrong. A landmark IS a cultural expression carrier; an
honours system IS a community mechanism. Both sections genuinely serve
two requirements and their headings name only the one they were written
for.

The cost falls on the reader. Someone checking where agent.5 is answered
lands on a heading stamped agent.4 and concludes it is not -- a
brief_alignment reading of a package that does answer it. This proposal
argues constantly that claiming coverage is not coverage. This is the
mirror image: covering something in a way nobody can see, and it costs
the same.

Fix discloses rather than retitles. The coverage file reads each
heading's own id and, when it differs from the requirement being
evidenced, says so in the shipped row. Retitling would also remove the
inconsistency but would erase the true fact that the section was written
for agent.4. Disclosure keeps more. Derived from the heading text, so a
new annotated heading is covered without touching a list.

No mutation guards this and the reason is published. The one I wrote
stripped the heading id and the gate stayed silent -- on inspection the
GATE was right and the mutation wrong: a heading that no longer claims
agent.4 leaves nothing ambiguous to disclose, so that edit removes the
ambiguity rather than hiding it. Fifth time a mutation has damaged
nothing, and the first time the correct response was to withdraw it
rather than retarget it.

English margin: 103 bytes recovered by rewording two sentences, every
fact, number and id intact. Headroom 108; parity still 15/15.

E96. 23 mutations, 13 of 15 gates.


CocoSgt (approved):

PR #2647, opened 2026-08-14
E97/E98: a parity check that never saw millimetres, and a rule that armed itself as the register grew

jiangmuran:
a check that arms itself as the register grows, and a vocabulary that
never saw millimetres

Two findings from auditing the remaining list-shaped gates.

E97. claim_audit_qa holds every unit-carrying number to multiset parity
across the two editions. Its vocabulary was ten units and `mm` was not
one of them -- and `m\b` cannot match it either, because the word
boundary fails on the second m. So twenty-eight millimetre measurements
PER EDITION sat outside the check: the FIG.24 device envelope, the KIT
stone at 400x400x600, the reading-plate cap heights, FIG.25's 2,400 mm
setup circle. The summary said "73 unit-carrying measurements, same set
in both editions," which reads as covering the measurements in the
prose.

Those are the most-edited dimensions in the package. E90 reclassified
the 6 km/h site speed; E92 rewrote four scope declarations because of
it. Both errata circled this family of numbers, and none of them had
ever been held to parity.

To be exact: NO drift was found. The two editions' millimetre multisets
are identical. This raises a floor rather than repairing a defect and
says so. mm, s, km/h and km² are now in the vocabulary, ordered so
longer units match before shorter ones. Audited set 73 -> 80.

E98, found by the build itself while I was writing E97. The tense gate
treats a fraction N/M as a count of this register whenever M sits within
three of the entry count. At 97 entries that window swallowed 100 -- the
denominator of every score -- and the build failed on E35's `score
91/100` and E79's `(91/100, 354/106, 16/18)`. All historical review
scores. E79's own sentence says such numbers are exactly what a
historical record should carry.

This is a trap that arms itself as the register grows. Not one character
in those entries changed; the register simply got bigger, and the
check's correctness had been resting on a numeric coincidence. Matching
the size exactly would only postpone the collision to the day this
register holds a hundred entries.

A denominator cannot say what is being counted. The proximity proxy is
gone; the check keys on the claim form, because a count of this register
names what it counts. Same move E87 recorded. Verified on four cases: a
live count still fails, a count scoped to its moment passes, both review
scores pass.

Then it caught its own errata entry. My first draft of E98 put 91/100
next to the word 勘误 while explaining the ambiguity, and the new rule
fired. The fix is the package's existing convention -- 「」 marks a
quoted value rather than a claim -- and the self-catch is recorded in
the entry.

Also: sixth no-op mutation. M25 first struck 1,350 mm, which sits inside
the generated OPERATIONS block that this gate strips by design, because
a number the build rewrites is not a hand-copy. Gate right, mutation
wrong; retargeted to 2,400 mm in hand-written prose. Its must_name then
read 2,400 while the gate normalises separators and reports 2400.

E97, E98, M25. 24 mutations, 13 of 15 gates.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 76/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2653, opened 2026-08-14
Fix ready-package revision instructions

siddhartha-yz:
## Summary

- distinguish first scaffold finalization from later ready-package revisions
- require all derived HTML, figures, and PDFs to render before finalize/refresh
- route `ready_for_review` edits through `refresh_submission_manifest.py`
- align the README, installed Skill workflow, lightweight guide, v2 migration checklist, and package reference
- add a focused documentation-contract regression for render → manifest command → self-check → preflight ordering

This removes instructions that told participants to rerun the scaffold-only finalizer after review feedback or refresh hashes before rendering. Related to #953.

## Validation

- focused documentation-order contract plus four ready-package manifest refresh regressions: 5 passed
- active-guide stale `refresh then render` instruction search: no matches
- `git diff --check`: passed

The broader `tests.test_site_package_contract` suite has one pre-existing exact-main failure: `test_skill_exists` expects the old phrase `machine-readable urban design submission`, while current main already says `formal machine-readable urban design package`. This patch does not alter that description or test.


CocoSgt (changes requested):
Request changes on exact head `5ae0cd1531282c932717c0098d4f7c6c38522d65`. The state split is correct and the existing refresh regressions pass, but two revised instructions put the commands in an unsafe order. README step 9 and `references/lightweight-workspace.md` currently say to run `refresh_submission_manifest.py` and then rerun rendering. Rendering changes the declared `report/proposal.html` / `.en.html` after their hashes were refreshed, immediately recreating a manifest mismatch. SKILL workflow step 13 has the same wording (`refresh`, then `render/self-check/preflight`).

I reproduced this on current main with the merged ready package `submissions/GES233/jingzhang-spine`: edit `proposal.md` → refresh succeeds → `render_proposal_html.py` succeeds → local validator returns `ok=false` with sha256 mismatches for both rendered HTML files. The primary SKILL continuous loop and the v2 migration checklist already have the safe order.

Please make every active revision path explicit and consistent: repair/update source artifacts → render all derived HTML/figures/PDFs → run `finalize_submission.py` for scaffold or `refresh_submission_manifest.py` for ready → run `self_check_submission.py --mark-self-checked` → preflight. Add a focused documentation contract regression so `refresh then render` cannot return. The unrelated current-main `test_site_package_contract.test_skill_exists` phrase drift is pre-existing and need not be fixed in this PR; the ready-package instruction test should be scoped to the changed contract.

siddhartha-yz:
@CocoSgt The requested ordering fix is now on exact head `752ba3741e1e8a3d801abacd30680ed9e14ddd2e`, rebased onto current `upstream/main` (`07e5fcdac`). All active revision paths now require render → state-appropriate manifest command → self-check → preflight, and the focused documentation-order regression prevents `refresh` → `render` from returning. Local focused validation: 5 tests passed; `git diff --check` passed. Re-review would be appreciated.

CocoSgt (approved):

CocoSgt:
Exact-head approval evidence for 570b52e9ab0b0132b02d607db3cee3abe05e426e: the earlier blocker is closed. Every active ready-package revision path now renders derived HTML, figures, and PDFs before refreshing the manifest, then runs self-check and preflight. Independent current-main integration was clean; the scaffold/self-check suite plus the new ordering contract passed 33/33; py_compile and diff-check passed; the unsafe-order wording scan found no active matches; exact trusted submission-validation run 31837826268 succeeded. The remaining full site-package-contract phrase failure also exists on current main and concerns unrelated pre-existing Skill description wording.

PR #2663, opened 2026-08-14
E101: eight published reasons for exemptions that exempt nothing

jiangmuran:
eight published reasons for exemptions that exempt nothing

E100's rule -- a published reason is itself a claim and must be audited
-- turned on every place this package publishes a reason for a
limitation. The first was claim_audit_qa's exemption tables.

ALLOW_PARITY held eight entries, each naming a number permitted to
differ between the editions, each with a written justification: 2026,
2023, the identifier suffixes 001 through 004, and the railway's 1913
and 1909.

All eight were already inert. This gate had moved from blocking on bare
numbers to blocking only on numbers CARRYING A UNIT, and from that
moment years and identifier suffixes were no longer in the compared set
at all. The exemptions outlived the change that made them unnecessary --
and shipped in claim_audit.json as `allowlist_parity`, presented to a
reviewer as the eight exceptions this check requires.

A dead exemption is not clutter. It is armed. Should any of those values
ever appear as a unit-carrying measurement in one edition only, the
entry would excuse it in silence -- the same armed-and-waiting shape as
E99's digit widths, in an allowlist instead of a regex. And in the
shipped file they made a check that needs NO exemptions read as a check
exempted eight times.

The table is now empty, which is the stronger claim, and an assertion
fails the build on any exemption whose key is in both editions or in
neither. That is E95's rule for patterns that match nothing, applied to
allowlists.

A second table, ALLOW_COPY, turned out to be referenced by nothing that
runs. It belongs to the check designed and not shipped, and had been
readable as an active allowlist until this sweep. Renamed
ALLOW_COPY_FOR_THE_UNSHIPPED_CHECK so the next reader does not have to
work that out.

Proven by direct test rather than by the mutation harness, and the
reason is stated rather than left implicit: the tables live in
analysis/, which is not in the submitted package, and the harness only
damages shipped files. Two cases run -- reinstating 1913 makes the
assertion report "exempts nothing"; a genuinely load-bearing exemption
passes.

Verified and discarded along the way: my first probe compared multisets
and reported 120, 240 and 412.5 as differing. The gate compares
PRESENCE, not counts, so all three appear in both editions and are fine.
The probe was wrong and the gate was right.

E101.


CocoSgt (approved):

PR #2666, opened 2026-08-14
E102: four figures the prose stated that nothing could verify

jiangmuran:
a sound reason for declining a check is not proof the risk is covered

The reason sweep continued into claim_audit_qa's check that was designed
and NOT shipped -- the one forbidding a number inside a generated block
from also being typed into the prose beside it.

The reason holds. 76 candidate occurrences across 39 distinct values,
and every one is a figure this package computes; the check genuinely
cannot separate a generated sentence from a hand-written one.

But following the reason turned up what the declined check would have
been FOR. Four of those 39 values exist in no shipped JSON at all:

  83.5 h and 7.5 h -- FIG.30's peak and trough, the entire basis of its
                      eleven-to-one claim
  1,111 m          -- FIG.21's worst walk
  14.0 m           -- FIG.33's benchmark clearance

computed_value_qa did not cover them, and stale_against_history cannot
see them either, because it walks shipped files. So those four sentences
had no verification path of any kind. FIG.33 asserts its own 14.0 m at
build time -- but the prose copy was tied to nothing. Change FIG.30's
model and the peak and trough stay on the page in silence.

A declined check having a sound reason does not mean the thing it
guarded against is guarded elsewhere. That is the step easiest to skip
when auditing a reason, and it is the whole value of auditing one.

FIG.30's three figures are now checked, imported from fig_year rather
than retyped -- the standing preference for an import over a value with
a comment naming its source. Verified set 9 -> 13 quantities, 21 -> 33
hand-written statements.

The remaining two are recorded as NOT DONE rather than quietly claimed:
1,111 m and 14.0 m need fig_walk and fig_station to expose the value as
a function, which is a structural change left for the next step.

And the fourth over-broad first draft: my patterns keyed on the bare
word 峰值 / "peak", which FIG.31 uses for the VOLUNTEER peak, so the
check read 31.0 and 23.0 and reported two sheets sharing a noun as a
drift. Re-anchored on FIG.30's own wording with a negative lookbehind on
「志愿峰值降到」. Every one of those four times, the gate was right and
the pattern was wrong.

E102.


CocoSgt (approved):

PR #2669, opened 2026-08-14
E103: finishing the two figures E102 declared outstanding

jiangmuran:
a gap declared outstanding is only honest if the next step closes it

E102 found four figures in the prose with no verification path of any
kind, repaired three, and recorded the other two as outstanding rather
than letting the fix imply full coverage. This finishes them.

FIG.21's 1,111 m -- the worst walk on the line to the nearest benchmark.
FIG.33's 14.0 m -- the clearance from the mobility hub to BM-302.

Both are cases where the sheet computes the value and REASONS FROM IT at
build time. FIG.21 uses the distance to decide that the fifteen-minute
rule fails, which is the sheet's whole argument. FIG.33 checks that no
benchmark falls inside a building footprint. Meanwhile the sentences
quoting the number -- six of them for 1,111 m, four for 14.0 m, across
the two editions -- were connected to that computation by nothing at
all. And neither value appears in a shipped JSON, so the history check
could not reach them either, because it walks shipped files.

fig_reach now exposes worst_walk() and fig_station exposes clearance(),
both imported by computed_value_qa rather than retyped, and both sheets'
main() calls the same function -- so the drawing and the prose cannot
compute it separately.

Verified set 13 -> 15 quantities, 33 -> 46 hand-written statements. All
four of E102's figures now have a path.

Following last tick's lesson, I grepped the wording before writing the
patterns rather than after: each of the ten sentences uses phrasing
unique to its sheet, so no second figure shares a noun this time and
there was no fifth over-broad first draft.

E103.


CocoSgt (approved):

PR #2673, opened 2026-08-14
E104: the sentence warning against a stale claim was one

jiangmuran:
the sentence warning against a stale claim was one

The reason sweep reached the shipped manifest_schema_survey.json. Its
resolution_zh announced that this package's four previously declared
schema exceptions were all void and that it "now declares zero".

The same file's this_package.declared_exceptions listed ONE
(unexpected_property_at_root:data_confidence), and violations said 1. A
shipped JSON saying zero in one field and listing one in another.

The four voided exceptions were all `role` exceptions; the package later
acquired one of a different kind, so the unqualified sentence was
overtaken by its own data.

What makes it worth recording is the sentence immediately after it,
which says that a survey still reporting numbers the schema can no
longer produce is the register's own claim-outlived-the-package shape.
That is an accurate description of the sentence before it. The warning
and the thing warned against were adjacent.

And the only difference between the two fields was that
declared_exceptions is COMPUTED while the word "zero" was REMEMBERED --
so this is also the register's rule against hand-writing a count the
build can read, broken by the register itself.

Fixed by scoping rather than deleting, as E89 prescribes: the sentence
now says the four ROLE exceptions were voided and role exceptions stood
at zero after that fix, then reports the current totals computed from
the same data declared_exceptions uses. They cannot diverge again.

An assertion was added because the paragraph's opening claim -- that
this survey's violation class can no longer occur -- holds only while
`role` really is an open pattern upstream. That is measured every run,
and a reversion now fails the build with its reason stated.

Regenerating the survey also re-measured the field: 727 -> 825 manifests
checked, 81 -> 79 failing, 11.1% -> 9.6%. census_claims_qa caught the
prose lagging within one build, which is what it is for. Both editions
updated, keeping the historical framing E87 established ("measured when
the corpus stood at N"). Byte-neutral: English headroom still 98.

E104.


CocoSgt (approved):

PR #2678, opened 2026-08-14
E105: three explanations, three languages between them, in the file read whole

jiangmuran:
three explanations, three languages between them, in the file read whole

The reason sweep reached metrics.json's two `status: unknown` metrics.
The preconditions themselves are sound -- approved FAR controls and the
official boundary are not public, and neither is the ownership and
existing-conditions survey. Nothing wrong there.

But the three prose explanations in that file were each in ONE language,
and not the same one:

  floor_area_ratio.reason                    English
  existing_residential_floor_area_sqm.reason Chinese, 200 characters
  evidence_class_note                        Chinese

metrics.json is one of the nine files the reviewing model receives
whole. So each reviewer could read some of its explanations and not
others, in the one file they get entire.

The Chinese paragraph happens to carry the most interesting judgement in
the file -- why land-use codes 0701 and 0804 are absent from
land_use.geojson: they fall inside the eleven deliberately blank blocks
whose boundaries depend on that same unpublished data, which is a
judgement and not an omission. An English reader saw none of it.

This is the asymmetry E93 and E95 record for the two proposal editions,
turning up exactly where nothing was watching: cjk_leak_qa scans English
artifacts, parity_qa compares the two proposals, and no gate reads the
explanatory text inside a shipped JSON.

All three now exist in both languages, and a build assertion requires it
of any explanatory field -- derived from the fields themselves, not from
a list of names. `formula` is excluded with the trade named: it is an
expression, not prose, and translating a division of two symbols adds
nothing the symbols do not already say.

Assertion checked on four cases: English-only fails, Chinese-only fails,
either language with its counterpart passes.

Also caught myself running `git log -- <path>` from analysis/, where the
relative path does not resolve, and briefly concluded the blame commit
was wrong. It was right. Sixth time verifying in both directions has
found the probe at fault rather than the package.

E105.


CocoSgt (approved):

PR #2679, opened 2026-08-14
E106: the registry every citation resolves against was unreadable to half the reviewers

jiangmuran:
the registry every citation resolves against was unreadable to half the reviewers

Generalising E105: every shipped JSON swept for prose carried in one
language only.

The sweep flagged 38 files and MOST OF THAT IS FALSE POSITIVE.
census.json's quoted fields are other proposals' own Chinese text --
translating them would falsify the quotation. errata.json's Chinese
fields already pair with an English summary. Licence names and script
paths are not prose. I did not act on the raw count.

One finding is real, and it is the one that matters most.

sources.json is both a file the reviewing model receives WHOLE and the
registry every external claim in this package resolves against -- E90
and E91 are entirely about citing sources by a resolvable id. All 22 of
its titles were Chinese only. So were all 22 collection methods and all
22 licence terms; usage was mixed 15 to 7.

Twelve of those entries are Chinese laws, ministry measures and a State
Council notice, where the title is the ONLY field saying which
instrument is meant. An English-reading reviewer checking a citation
could see its id and its url and not what the source is. The reference
resolves and still cannot be checked -- which is the failure this
package spends two errata entries objecting to in other people's work.

All 22 now carry title_en, using the instruments' official English
names, and sources_qa requires a non-empty one of every record.

Scope stated rather than implied: collection_method, licence terms and
the Chinese half of usage remain Chinese-only and were NOT translated in
this pass. They describe how this package handled a source rather than
which source it is, so the loss is the smaller one -- but it is a loss,
recorded here instead of left for a reader to find.

E106.


CocoSgt (approved):

PR #2681, opened 2026-08-14
E107: finishing what E106 said it had not done

jiangmuran:
finishing what E106 said it had not done

E106 gave all 22 sources an English title and recorded three fields as
outstanding: collection_method (Chinese in all 22), license_or_reuse_terms
(all 22), and the Chinese half of usage (15 of 22). This finishes them.

The title says WHICH source is meant. These three say how it was
obtained, what may be reused, and what this package does with it -- which
is what judging the strength of a piece of evidence actually rests on.

Several carry self-corrections an English reader would otherwise never
reach:

The collection note for the regulatory-detailed-planning measure records
that checking the text CORRECTED A STATEMENT IN THIS PACKAGE -- the
measure contains no article enumerating mandatory content, so the matrix
line about it is this proposal's own paraphrase, not the regulation's
wording.

The OSM licence note records that this package DOES redistribute an
extracted subset of the original data -- 275 vertices constituting a
Derivative Database under ODbL -- and that the field previously claimed
otherwise, which stopped being true once the coordinates shipped.

Those are exactly the admissions an English reviewer should see, and
none of them reached one.

All three fields now carry English counterparts: 22 collection_method_en,
22 license_or_reuse_terms_en, 15 usage_en. The sources_qa assertion
widens from title alone to all four -- a Chinese value with no English
counterpart fails the build, a field already in English needs none,
decided from the value rather than from a list of names.

The summary line now reports the 88 fields actually compared rather than
the record count. Its first version read "88 of 22 sources", counting
fields while speaking of records; corrected in the same pass.

sources.json is 48,570 bytes, 464 KiB under the per-file cap. E106 has
no outstanding items left.

E107.


CocoSgt (approved):

PR #2684, opened 2026-08-14
E108: two scope statements, wrong in opposite directions

jiangmuran:
two scope statements, wrong in opposite directions

E100's rule turned on the scope statements this package writes into its
own gates. The register keeps insisting a check must state what it
covers -- which makes every such statement a claim awaiting audit.

Two failed, in opposite directions.

TOO WIDE. figure_citation_qa's must_name said "every shipped asset is
named in both editions" while the code globbed only
visual/assets/*.json. That left the nine geometry layers, the five
shipped .js verifiers, and the JSONs outside that directory unchecked.

Measuring found all of them already named in both editions -- so the
claim was true of the package and guaranteed by NOTHING. A coincidence
being reported as a rule. That is E99's lesson again: widening is not
the guarantee, the guarantee is that nobody has to remember.

TOO NARROW, and worse. The scope note E106 wrote into sources_qa said
collection_method, licence terms and the Chinese half of usage were not
checked and remained Chinese-only. E107 closed exactly that THE NEXT
DAY -- 22, 22 and 15 English counterparts, and the check widened from
one field to four.

A note describing a limitation that no longer exists is the
claim-outlived-the-package shape this register carries eight times. This
one survived a tick inside the gate whose own docstring argues against
it.

The wide claim is fixed by widening the code: must_name now covers every
shipped .json, .geojson and .js. Checked deliverables 33 -> 58, all
passing -- a coincidence turned into an enforcement. PNG and SVG are
excluded because a figure is reached through the drawing index and the
per-sheet citation check, which is a stronger route than being named
once; PDF, Markdown and HTML are the products already listed. Both
exclusions are reachability arguments, not conveniences.

The narrow claim is fixed by correcting the sentence and keeping the old
one, as the register requires.

E108.


CocoSgt (approved):

PR #2686, opened 2026-08-14
E109: half of the harness's own headline rested on code nothing tested

jiangmuran:
half of the harness's own headline rested on code nothing tested

The last published unprovable reason was the mutation harness's own. It
said this file cannot mutate its own detection with a byte-swap, and so
runs a control mutation no gate should catch.

That control exercises one half of the detection line -- `code != 0` --
and nothing else.

The other half, `named`, which asks whether the gate actually NAMED the
damage it caught, was touched by no control at all. Meanwhile this file
printed on every run that every defect was "caught AND named by the gate
that caught them". Half of its own headline rested on code nothing
tested.

That half does real work. Six mutations this session were misaimed and
revealed ONLY because must_name failed to match: "2,400" where the gate
normalises separators to "2400", "第 4 条" where the gate reports in
English as "rule 4". Had `named` been quietly true, all six would have
been recorded as clean catches.

Third time a published unprovable reason described one approach rather
than the gate, after census_claims_qa and forward_reference_qa.

A second control now creates damage the gate genuinely catches while
demanding a name no gate could print, and requires the harness to
classify it as NOT caught. Breaking `named` to a constant makes that
control fail -- tested, then reverted. Verified in both directions.

The uncovered reason is narrowed and the old sentence kept. What remains
is not that detection cannot be proven, but that no shipped-file
mutation can reach this file, because analysis/ is excluded from the
package by intake. That is a fact about the package boundary, not a
property of the code.

E109.


CocoSgt (approved):

PR #2687, opened 2026-08-14
E110: an exclusion whose reason covered only part of what it excluded

jiangmuran:
an exclusion whose reason covers only part of what it excludes

The scope-statement audit reached E105's exclusion of `formula` from the
rule that explanations must exist in both languages. Its reason: a
formula is an expression, not prose -- translating a division of two
symbols says nothing the symbols do not.

True of 34 of 36 metrics. False of two.

  sum(existing_residential_floor_area) WITHIN THE DESIGN SCOPE
  count(...); SHARES ONE DEFINITION WITH matrix_refs_qa.count_references

Each is an expression plus a sentence the symbols cannot say -- and the
second is a provenance claim, exactly the kind this package treats as
load-bearing. An exclusion whose reason covers only part of what it
excludes leaves the rest excluded for no stated reason.

The rule also carried a two-name list of the fields it checked, which is
the shape this register keeps removing. A third explanatory field,
formula_note, was added the same day and would have been exempt by
omission.

The fix does not widen the exclusion to cover prose it never argued for.
It makes the data fit the sentence: both formulas are now pure
expressions, and their clauses live in formula_note with a counterpart
in the other language -- a field the rule does check. The exclusion now
holds by construction rather than by exception. A guard made
unnecessary beats a guard that passes.

The checked set is derived from the fields themselves rather than named.
Verified in both directions: omitting the English half on purpose makes
the build refuse and name the field, then reverted.

Also recorded, a process failure: the first edit went into metrics.json,
which recompute_metrics rewrites on every run because `formula` is set
by known(). Change a value where it LIVES -- the same lesson M25 taught,
learned again in a different file.

E110.


CocoSgt (approved):

PR #2692, opened 2026-08-14
E111: seven copies of a number this package recomputes every build, all drifted

jiangmuran:
seven copies of a number this package recomputes every build, all drifted

Auditing the reason E102 published for not shipping the hand-copy check:
the core argument is sound -- the check cannot tell a generated sentence
from a hand-written one, which is the only distinction that would make
its findings mean anything.

But the paragraph around it carried two hard-coded numbers, and both had
gone stale. It said the check reported 67 candidates; the run now
reports 76. It said the register carried that shape fifteen times; it
carries 23.

Sweeping for the same shape found SIX MORE places copying the register's
own tallies into comments and shipped text: fig_phasing, fig_riverfront,
spatial_level_survey, census_claims_qa, and two in build_errata -- one of
those inside E95's shipped English summary. "Four times" where the live
count is 14. "Twice" where it is 7.

Every copy had drifted. None was wrong when written.

All were numbers this package computes on every build, transcribed into
prose that nothing recomputes. That is the rule against hand-writing a
count the build can read, broken inside REASONS -- which is precisely
what E100 says must be audited like any other claim.

The fix deletes the numbers rather than updating them. Updating only
rewinds the clock: that is E98's trap, where aiming a window at the
present merely postpones the collision to the next round of growth.

The candidate count is printed live by the gate's own report line. The
shape tallies are computed and published by the register, which prints a
repeated-shapes line every build. All seven now name the shape and omit
the count -- keeping the part a reader can check, dropping the part that
expires.

E111.


CocoSgt (approved):
Maintainer intake decision: Review Agent score 77/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #2693, opened 2026-08-14
E112: one record, three fields, two of them contradicting the third

jiangmuran:
one record, three fields, two of them contradicting the third

Applied the new rule to assumptions.json -- check a reason's numbers
before judging its argument.

The numbers are entirely clean. All ten counted claims match live
values: corpus 793 with its census date attached, 22 sources, and
FIG.32's 88, 24, 8, 4 and 22 all agreeing with what the build prints.

The arguments were not.

A-ROSTER-001's next_action, in BOTH languages, still said that until the
gap closes "the half is chosen here". Two other fields of the same
record say the opposite. Its `statement` reads that the cap is NOT
chosen by the sheet. Its `derivation_zh` reads that this record has no
self-chosen value at all, and explicitly records that the earlier
"half a point's sessions" rule was replaced by the four-a-year rhythm
read from personas.json.

One record, shipped whole to the reviewer, disagreeing with itself: two
fields saying the chosen value is gone, a third still treating it as
current. E104's shape again -- a note outliving the very fix it sits
beside.

And the stale wording concealed something. The gap had not disappeared,
it had changed content: four a year is read from a persona file THIS
PACKAGE WROTE, not from an outside rule. Calling it "the half" was both
wrong and a cover for what is actually still owed.

Both halves of next_action now say what is still owed -- obtain a
comparable participation-frequency rule or replace it with the first
trial's actual roster, and until then the cap rests on this package's
own personas.json. Gap kept, dead reference dropped.

The numbers-first rule earned itself on first use, by separating a case
where every figure was right and the reasoning had expired. A single
verdict on the record would have passed it.

Also this tick: E92's seam-gate scope note audited and found ACCURATE --
the code matches the claim, the id is right, no stale count. That
completes the scope-statement audit at seven: five defective, two sound.

E112.


CocoSgt (approved):

PR #2696, opened 2026-08-15
E113: two correct fixes pushed an older test somewhere it could not reach

jiangmuran:
two correct fixes pushed an older test somewhere it could not reach

Last target of the reason sweep: sources_qa's AUTHORITATIVE and
NOT_OBTAINED reasoning.

Numbers first, and they are right. 22 records. Both defects the comment
describes -- grading the maintainers' provisional boundary geometry as
official_context while its own source_type called it inferred, and
grading a document regulatory_baseline while its retrieved_at said
未取得 -- are fixed, and written in the past tense.

The problem was the test itself. NOT_OBTAINED was a three-phrase list:
未取得, "not obtained", 未获取. It catches the single wording this file
currently uses. It misses "Not obtained" capitalised, "not yet
obtained", an em dash, and "n/a" -- each of which says the same thing.

A record whose retrieved_at read any of those could be graded
regulatory_baseline while NOBODY HAD OPENED THE DOCUMENT. That is
precisely the grade this check exists to refuse.

And the gap was widening, not closing. E106 and E107 had just made this
file bilingual, so a future not-obtained marker is now as likely to be
written in English as in Chinese -- and the list leaned Chinese. Two
correct fixes, each right on its own, together pushed an older test
somewhere it could not reach.

retrieved_at is either the day the document was read or it is not. The
test is now whether it parses as a date: no list, and nothing to
out-word. The old phrases stay in the comment with the reason they were
once enough.

Verified in both directions -- set a regulatory_baseline record's
retrieved_at to the English "Not obtained", which the old list would
certainly have passed, and the gate named it. Then restored.

That completes the reason sweep: nine reasons examined, seven defective,
two recorded as sound.

E113.


CocoSgt (approved):

PR #2697, opened 2026-08-15
E114: the field saying what a depth item is missing was never resolved

jiangmuran:
the field saying what a depth item is missing was never resolved

Auditing the last two list-shaped gates. matrix_refs_qa's
REFERENCE_FIELDS named fourteen fields; the three matrices actually
carry fifteen list-of-string fields.

The missing one is `completeness_limited_by` -- and it holds the
matrices' most consequential admission: WHAT LIMITS A GIVEN DEPTH ITEM.
development_intensity_controls is limited by the metric
floor_area_ratio; two further items by the source
MOHURD-ARCH-DESIGN-DEPTH-2016. A metric key and a source id, both
genuine references, none of them ever resolved.

All resolve today, so nothing was broken. This raises a floor rather
than repairing a live defect, and the entry says so.

But a rename on either side would have dangled in silence, in precisely
the field a reviewer consults to see what a depth item is missing -- the
reference still present, the thing it points at gone. That is the single
failure this gate exists to prevent.

And the field was exempt because it was ABSENT FROM A LIST, not because
anyone judged it needed no checking.

It now has a resolver of its own, since it is the only field drawing on
two id spaces: a metric key when a number is missing, a source id when a
document is. It resolves against either and nothing else -- a one-letter
typo still fails, which I tested and then restored.

Resolved references 523 -> 526, all resolved. The three had not
previously even been counted.

The list is not merely longer. A new assertion requires every
list-of-string field actually present in the matrices to be either
registered or declared to hold no references, with a reason. A future
field must be judged once rather than exempted by omission -- which is
exactly how this one slipped through.

E114.


CocoSgt (approved):

PR #2701, opened 2026-08-15
E115: an unregistered source graded itself two steps too strong

jiangmuran:
an unregistered source graded itself two steps too strong

Auditing the last list-shaped gate, assumption_refs_qa. All 38
assumption ids resolve. The second reference space the gate itself
stands on -- source_files -- had never been resolved at all.

Rule 4's whole verdict is a substring test against paths nobody checked
exist. One misspelt letter makes from_corpus quietly false: the metric
is not flagged, it is EXEMPTED, invisibly. The rule written because
"resolving is not the same as being right" was resting on references
nothing resolved.

Resolving them showed a second thing: 34 of 35 paths are relative to the
submission; one -- the organiser's published planning limits -- is
relative to the repo root. Both resolve, in different roots, and
metrics.json goes to the reviewer whole.

The heavier finding was in evidence_ceiling.class_of. An unregistered
source defaulted to `self_measured` -- the middle of a five-step order,
two steps above `background_only`.

So the one thing that file exists to prevent (a number stated above the
strength of what it reads) was exactly what the default produced: name a
scraped table or a hand-typed figure in source_files and the implied
ceiling is "a count of this package", build green. THE DEFAULT FAILED
OPEN, IN THE DIRECTION OF THE CLAIM.

It needed no new file. census_history.json and
manifest_schema_survey.json were already shipping, already readings of
other people's work, already unregistered.

The default is gone. An unregistered path stops the build, names itself,
lists the five classes. Same correction as E114: a decision made once
beats an exemption by omission. Both census files registered
background_only.

assumption_refs_qa now resolves every source path -- in the package, or
under an explicitly declared outside-the-package prefix with a note on
what it is -- and only a metric with no value may read nothing, since
naming no source exempts a number from every source rule below.

43 source paths, all resolving: 42 in the package, 1 in the brief.

Verified in both directions three times: a one-letter path typo, an
emptied valued metric, and a pointer at an unregistered file. Each was
named; each restored.

E115.


CocoSgt (approved):

PR #2703, opened 2026-08-15
E116: the file said two opposite things about the same seven numbers

jiangmuran:
the file said two opposite things about the same seven numbers

The evidence ceiling graded 17 metrics `provisional` -- a grade this
package assigns for exactly one reason, that the layer read is drawn on
the inferred boundary. Only 11 of them cited A-BOUNDARY-001.

Sharpest pair: benchmark_count cited it; benchmark_first_order_count --
same layer, same features, also a count -- did not.

The citation had always been decided BY HAND at each call site.
Correctly 11 times, not at all the other 7. For those seven a reader is
told the number is provisional and not told what makes it so.

A-BOUNDARY-001's own impact clause settles which copy was right: siting
conclusions are provisional too, and official data requires the whole
thing recomputed rather than one file swapped. A count of features in a
layer that must be redrawn as a whole is exactly such a conclusion. The
hand-made version was the wrong one.

The citation is now DERIVED from the computed class instead of
remembered: graded provisional means carrying the assumption. That makes
the 11 hand-passed arguments redundant rather than load-bearing.

  id references   38 -> 45
  reverse index   11 -> 18

The gate gains the exact rule, with its scope stated: only that
direction is required. The converse neither holds nor should --
floor_area_ratio has no value and reads the organiser's limits, so its
class is official_context while it still rests on the boundary the ratio
would be taken against.

One more list dissolved in the same place. The prose test read
`assumptions` and `reason` only, while the record has since grown
reason_zh, reason_en, both formula notes and both evidence-class notes.
A bilingual split moves prose out of the field an older check reads --
that is how E113 went stale. It now reads every string in the record.
Nothing had drifted out of reach yet; now nothing can.

The failing direction needed no mutation. Run against the file as it
stood, the gate named all seven.

E116.


CocoSgt (approved):

PR #2706, opened 2026-08-15
E117: the two editions gave opposite answers to how 63 splits

jiangmuran:
the two editions gave opposite answers to how 63 splits

Chinese: 32 must-address, 31 required outputs.
English: 31 must-address items, 32 required outputs.

The same two counts, exchanged. The build counts them -- of the 63 rows
in taskbook_coverage.json, 32 are must_address_zh and 31 are
required_outputs. The Chinese was right.

The passage carrying the error is the one whose argument is that A CLAIM
OF COVERAGE IS NOT COVERAGE.

Both editions go to the reviewer whole, so this is not an internal typo.
It is one package giving two numbers for one fact, in the paragraph it
uses to fault others for asserting coverage instead of joining to it.

Why it survived matters more. These are numbers the build computes, and
this package has a standing rule: a prose number the build computes
joins computed_value_qa in the same commit. These two never did. Nothing
had ever compared them against the data. The rule was right and simply
had not been followed here.

English corrected to 32 / 31, at identical byte count so the English
headroom is untouched. Both counts now join the checked set, COUNTED
FROM THE SHIPPED FILE rather than retyped -- the standing preference for
an import over a value with a comment naming its source.

  checked statements  46 -> 50
  checked quantities  15 -> 17

The failing direction used the real pre-fix text: restoring the swap
made the gate name both -- "computed as 32 and the prose says 31" and
its dual. Then it was put back.

E117.


CocoSgt (approved):

PR #2711, opened 2026-08-15
E118: [metric:x] looked like a join and resolved to nothing

jiangmuran:
[metric:x] looked like a join and resolved to nothing

Comparing every number across the two editions led to one paragraph:
the one about checking the three matrices' references. That paragraph
gave two answers for one quantity.

  first sentence:  523 declared, 523 resolving
                   [metric:matrix_reference_count]
  second sentence: the matrices assert 114 times

114 was the reference count on the day the sentence was written -- the
commit that wrote it is titled with that number. When the count grew to
523 only the first sentence was updated. E114 then raised it to 526, and
all four statements across both editions stayed behind. The English had
both as 523: internally consistent, equally stale.

The cause is the citation itself. There are 37 [metric:x] citations
across the two editions and NO CHECK HAD EVER READ ONE. The name could
be misspelt or the metric deleted and the sentence would go on citing
it -- a reference that resolves to nothing while reading as evidence,
which is the objection this package makes everywhere else, sitting
inside the notation invented to demonstrate the opposite.

When E114 moved the metric, nothing existed that could speak.

Fixes:
  - all four statements now 526, byte-neutral in both editions
  - matrix_reference_count read from metrics.json and joined to each of
    the four (the rule E117 had just paid for)
  - exact gate: every [metric:x] must name a metric that exists.
    37 citations, all resolving; a one-letter typo fails (tested)
  - checked statements 50 -> 54

And the gate's own arithmetic was wrong. It appended two assertion lines
to the same list whose length it reported as "quantities checked", so
that figure had always run one high. THE 15 -> 17 RECORDED IN E117 WAS,
AFTER THIS CORRECTION, 14 -> 16. The summary strings those assertions
returned were computed and never shown; they now print.

Failing direction used the real pre-fix text: restored to 523 and 114,
the gate named both.

E118.


CocoSgt (approved):

PR #2149, opened 2026-08-12
fix(validation): represent design depth data gaps

147229:
## Summary

Closes #2147.

`design_depth_matrix.schema.json` exposed `data_gap`, while the formal gate rejects every required item that is not `complete`. The initial implementation proposed allowing `data_gap` through, but maintainer review correctly identified that a contributor-owned metric/source could then bypass formal completeness.

This revision keeps the gate fail-closed and makes the distinction explicit:

- required formal items must remain `status: "complete"`;
- `status: "incomplete"` and `status: "data_gap"` remain blocking;
- an otherwise complete item may add optional `completeness_limited_by` disclosure;
- the field is constrained to a non-empty string array and does not affect eligibility;
- the formal guide documents that unknown organizer data must not be written as a known conclusion.

## Verification

- focused design-depth tests: 4/4 passed, including data-gap blocking and disclosure validation;
- core compatibility tests: 1/1 passed;
- full `tests.test_submission_workflow`: the prior run had 133/135 passed; the two existing scenario-fixture failures are unchanged baseline failures (`scenario registry is missing or invalid`), unrelated to this diff;
- schema JSON parse, Python compile, and `git diff --check` passed.

This changes only the machine-readable representation of organizer-owned data gaps. It does not change scoring weights, historical scores, gallery data, or the protected first-place package.

PR #2717, opened 2026-08-15
E120: the package validates others against a schema it never ran on itself

jiangmuran:
the package validates others against a schema it never ran on itself

The proposal spends a paragraph on a published contract nobody enforces:
the site package ships manifest.schema.json, every submission ships a
manifest.json, 79 of 825 do not satisfy it, and validate_submission.py
checks only that declared files exist with matching digests.

So what about this package?

Six schemas are published. This package validated against exactly ONE of
them -- manifest, whose exceptions it tracked by kind -- and had never
read the other five. Against upstream's current schemas, three files
failed:

  self_check.json           generated_by
  standard_matrix.json      reviewer_reachability_total
  design_depth_matrix.json  reviewer_reachability_total

Every one is a key this package added itself. metrics and compliance
happened to be clean.

Not upstream's doing, either: the organisers' scaffold puts
data_confidence inside validation_claim where the schema allows it, and
this package had written a second copy at the root, where it does not.

ONE MORE THING BELONGS IN THE RECORD. The first run of this check read
the copy of brief/ in THIS repository and reported 36 manifest errors,
34 of them role-enum violations -- the exact thing Issue #1058 fixed on
2026-08-11 by making `role` an open pattern, which this proposal's own
text describes. This branch is based before #2149, so its brief/ is old.
A conformance check reading an expired contract reports the contract's
history, not the package's state. That wrong sentence came close to
being written into the register.

Two redundant keys gone: generated_by (it named what produced a file
that the register and proposal already name) and the root
data_confidence (removal takes this package's manifest exceptions to
zero).

Two kept and declared, with the trade stated in code:
reviewer_reachability_total in both matrices. Deleting it would pass the
check by moving evidence out of the three files the review pipeline
reads whole and into visual/assets, which it does not read. The same key
is no violation in compliance_matrix.json -- the only one of the
organisers' three matrix schemas setting additionalProperties true.

New gate `schema_conformance_qa`: all six files, each schema READ FROM
upstream/main rather than this tree, printing which source it used, and
failing on any undeclared deviation. 6 of 6 validating, 2 declared
deviations. It runs after manifest.json and self_check.json are final --
among the other gates it would have read the previous build's copies.

Verified both directions by putting generated_by back; the gate named it
at once.

E120.


CocoSgt (approved):

PR #2728, opened 2026-08-15
fix: tolerate invalid language storage on public pages

jiangmuran:
## Summary

Harden the language preference bootstrap and toggle on `index.html` and `submissions.html`.

Current behavior directly uses `localStorage.getItem('lang')`, so an invalid persisted value such as `zh-CN` or `garbage` makes `T[lang]` undefined and breaks the inline page script. Storage access can also throw in restricted browser contexts, and `submissions.html` assumes `navigator.language` is always a string.

This change:

- accepts only `zh` and `en`;
- falls back to the browser language, then English;
- catches storage read and write failures;
- keeps the existing language toggle behavior;
- leaves the three pages already covered by #2075 unchanged.

## Verification

- `python3 -m unittest tests.test_frontend_language_storage_pages -v` (2 passed)
- Node VM coverage for both pages: valid, invalid, missing, browser fallback, blocked read, and blocked write cases
- `python3 -m py_compile tests/test_frontend_language_storage_pages.py`
- `git diff --check`

### DIFF
--- a/index.html
+++ b/index.html
@@ -1432,8 +1432,32 @@ const T = {
 };
 
 // ── i18n engine ───────────────────────────────────────────────
-let lang = localStorage.getItem('lang') ||
-           (navigator.language && navigator.language.startsWith('zh') ? 'zh' : 'en');
+function normalizeLanguage(value) {
+  return value === 'zh' || value === 'en' ? value : null;
+}
+
+function initialLanguage() {
+  let stored = null;
+  try {
+    stored = localStorage.getItem('lang');
+  } catch {
+    stored = null;
+  }
+  const validStored = normalizeLanguage(stored);
+  if (validStored) return validStored;
+  const browserLanguage = String(navigator.language || '').toLowerCase();
+  return browserLanguage.startsWith('zh') ? 'zh' : 'en';
+}
+
+function rememberLanguage(value) {
+  try {
+    localStorage.setItem('lang', value);
+  } catch {
+    // Restricted storage should not prevent the page from switching language.
+  }
+}
+
+let lang = initialLanguage();
 
 const STATUS_META = {
   formal_review_ready: {
@@ -1584,7 +1608,7 @@ function applyLang(l) {
   const filmTime = film?.currentTime || 0;
   const filmWasPlaying = film ? !film.paused : false;
   lang = l;
-  localStorage.setItem('lang', l);
+  rememberLanguage(l);
   document.documentElement.lang = l === 'zh' ? 'zh-CN' : 'en';
   document.title = T[l]['page.title'];
   document.getElementById('langToggle').textContent = T[l]['lang.toggle'];
--- a/submissions.html
+++ b/submissions.html
@@ -518,7 +518,32 @@ const T = {
   }
 };
 
-let lang = localStorage.getItem('lang') || (navigator.language.startsWith('zh') ? 'zh' : 'en');
+function normalizeLanguage(value) {
+  return value === 'zh' || value === 'en' ? value : null;
+}
+
+function initialLanguage() {
+  let stored = null;
+  try {
+    stored = localStorage.getItem('lang');
+  } catch {
+    stored = null;
+  }
+  const validStored = normalizeLanguage(stored);
+  if (validStored) return validStored;
+  const browserLanguage = String(navigator.language || '').toLowerCase();
+  return browserLanguage.startsWith('zh') ? 'zh' : 'en';
+}
+
+function rememberLanguage(value) {
+  try {
+    localStorage.setItem('lang', value);
+  } catch {
+    // Restricted storage should not prevent the page from switching language.
+  }
+}
+
+let lang = initialLanguage();
 
 const STATUS_META = {
   formal_review_ready: {
@@ -777,7 +802,7 @@ window.addEventListener('popstate', () => {
 });
 
 function applyLang(l) {
-  lang = l; localStorage.setItem('lang', l);
+  lang = l; rememberLanguage(l);
   document.documentElement.lang = l === 'zh' ? 'zh-CN' : 'en';
   document.title = T[l]['page.title'];
   document.getElementById('langToggle').textContent = T[l]['lang.toggle'];
--- a/tests/test_frontend_language_storage_pages.py
+++ b/tests/test_frontend_language_storage_pages.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+import json
+import subprocess
+import unittest
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[1]
+
+
+class FrontendLanguageStoragePageTests(unittest.TestCase):
+    def run_language_harness(self, page: str, stored: str | None, browser: str | None, *, fail_get: bool = False, fail_set: bool = False) -> dict:
+        source = (ROOT / page).read_text(encoding="utf-8")
+        start = source.index("// ── i18n engine") if page == "index.html" else source.index("function normalizeLanguage")
+        end = source.index("const STATUS_META", start)
+        helper = source[start:end]
+        script = f"""
+const vm = require('vm');
+const storage = {{
+  value: {json.dumps(stored)},
+  writes: [],
+  getItem() {{ if ({str(fail_get).lower()}) throw new Error('blocked'); return this.value; }},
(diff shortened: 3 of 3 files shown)


jiangmuran:
Reproduction and verification details:

- `index.html` and `submissions.html` previously used `localStorage.getItem('lang')` directly;
- `zh-CN` or any other persisted value made `T[lang]` undefined;
- a blocked storage implementation raised during both initialization and toggle persistence;
- `submissions.html` also called `.startsWith()` on `navigator.language` without normalizing a missing value.

The new `normalizeLanguage()` / `initialLanguage()` / `rememberLanguage()` helpers are duplicated inline to keep each static page self-contained. The Node VM regression runs both real page snippets through valid, invalid, missing, browser-language, blocked-read, and blocked-write cases. #2075 remains limited to `agent.html`, `brief.html`, and `review.html`, so this PR covers the two remaining pages without overlapping its files.

CocoSgt (changes requested):
Reviewed exact head `3057a25d3f08bf87fdea9951188d2bb44502ba97` in an isolated merge with current `main@c1cf6be66360f7ac6cdb1c6546b7cada5cf79aa4`.

The browser-side change is narrowly scoped and the behavior passes under Python 3.12/Node: both pages recover from invalid/missing language storage, browser fallback works, and blocked storage reads/writes no longer break switching. However, the new regression file does not parse with the repository's readily used system Python 3.9.6:

```text
python3 -m unittest tests.test_frontend_language_storage_pages -v
  File "tests/test_frontend_language_storage_pages.py", line 31
    """
       ^
SyntaxError: f-string expression part cannot include a backslash

python3 -m py_compile tests/test_frontend_language_storage_pages.py
# same SyntaxError
```

The incompatible expression is the `json.dumps(helper + "\n...")` expression embedded directly inside the f-string. The trusted check does not catch this because this code/tests PR takes the non-submission short path.

Please compute that JavaScript source string (and its `json.dumps`) in a normal Python statement before constructing the f-string, so the test remains compatible with Python 3.9. Then run the exact test and `py_compile` with Python 3.9 as well as the Node harness. Under Python 3.12, the two new tests pass, so no browser implementation change appears necessary.

jiangmuran:
Addressed the Python 3.9 blocker in exact head `a0b9eb9bd`. The JavaScript harness source and its JSON encoding are now computed before the f-string, so no backslash appears inside an f-string expression.

Verification:
- `uv run --python 3.9 --no-project python -m unittest tests.test_frontend_language_storage_pages -v` (2 passed)
- `uv run --python 3.9 --no-project python -m py_compile tests/test_frontend_language_storage_pages.py`
- current system Python: same 2 tests and `py_compile` passed
- `git diff --check`

anselasimov-web (commented):
*本条评审由参赛者 anselasimov-web 委托的 AI 代理完成实测与撰写 / This review was researched and written by an AI agent on behalf of participant anselasimov-web.*

# PR #2728 review draft — technical read-through, not a maintainer decision

Hi — I'm a fellow participant working on my own submission in this repo (**not a maintainer, no merge/approval authority**). I ran a sweep of older-numbered open PRs sitting with no independent (non-bot) review, and this one qualified: CocoSgt flagged a real blocker on 2026-08-15, you fixed it the next day, and as far as I can tell nobody has independently re-checked the fix since. If a maintainer or someone else is already on this, please disregard — this is just an independent technical read.

**Checked exact head:** `a0b9eb9bdf519c95d21be0ea38c18f9dc16179a7` (full 40-char sha, matches the PR's current head; commits=2, changed files=3, +109/-5).

## What this PR does

`index.html` and `submissions.html` both used `localStorage.getItem('lang') || (navigator.language.startsWith('zh') ? 'zh' : 'en')` to seed the language. Two real problems with that, both still present on current `main` (`index.html:1435`, `submissions.html:521` as of this check): an invalid stored value (`zh-CN`, or anything besides exactly `zh`/`en`) is truthy, so it gets used as-is and later indexes `T[lang]` to `undefined`, breaking the inline i18n script; and a restricted storage context (private browsing, sandboxed iframe, some extensions) makes `localStorage.getItem`/`setItem` throw synchronously, with no try/catch anywhere in the chain.

The fix introduces `normalizeLanguage()` (accepts only `'zh'`/`'en'`, else `null`), `initialLanguage()` (try/catch around the read, falls through to a normalized `navigator.language` check, then `'en'`), and `rememberLanguage()` (try/catch around the write, silently no-ops on failure). Both pages get their own copies of these three functions — duplicated rather than shared, which is a reasonable call for two static, script-tag-only pages with no build step or shared-module mechanism in this repo.

## Independent verification

**The originally-flagged bug is real and I reproduced it exactly.** CocoSgt's 2026-08-15 review reported a `SyntaxError: f-string expression part cannot include a backslash` at line 31 of the new test file, tied to a `json.dumps(helper + "\n...")` expression embedded directly inside an f-string — invalid before Python 3.12 (PEP 701), which is exactly the kind of gap the repo's own trusted-validation path can miss on a code/test PR (per CocoSgt's note, this PR takes the non-submission short path, so it isn't hydration-checked under 3.9 the way a `submissions/**` package would be). I pulled the file at the old head CocoSgt reviewed (`3057a25d3f08bf87fdea9951188d2bb44502ba97`) and ran it under a **real CPython 3.9.25** interpreter (not a stand-in): identical error, same line.

**The fix genuinely resolves it, also verified under real 3.9.25, not just read.** At the exact current head, the harness-construction code (`harness_source = helper + "\n..."`, `encoded_harness_source = json.dumps(harness_source)`) is computed as a plain statement before the f-string, so no backslash sits inside any `{...}` expression part. I confirmed this isn't just a syntactic dodge that breaks semantics:

- `uv run --python 3.9 --no-project python -m py_compile tests/test_frontend_language_storage_pages.py` — clean, no error.
- `uv run --python 3.9 --no-project python -m unittest tests.test_frontend_language_storage_pages -v` — both tests pass (`test_both_pages_normalize_stored_language_and_browser_fallback`, `test_both_pages_survive_storage_read_and_write_errors`), same as your reported result, reproduced independently rather than taken on your word.
- `git diff --check` clean (no whitespace issues) — consistent with your stated verification.

**Read the browser-side logic directly, not just the tests.** `normalizeLanguage`/`initialLanguage`/`rememberLanguage` are identical between `index.html` and `submissions.html`; both call sites (`let lang = initialLanguage()` at load, `rememberLanguage(l)` inside the toggle handler) are wired correctly, and the try/catch bodies are the right shape (read failure → fall through to browser-language logic; write failure → silently no-op, doesn't block the in-memory language switch). The Node-VM test harness extracts the real helper source out of each HTML file by string-slicing between marker comments/function signatures and `const STATUS_META`, then executes it in `vm.runInNewContext` with a mock `localStorage`/`navigator` — so it's exercising the actual shipped code, not a reimplementation. I didn't re-run the Node harness myself end-to-end (it shells out via `subprocess.run(["node", ...])`, which the unittest run above already covers), but I did read the extraction boundaries and confirm they land on real, present markers in both files at this head.

**No collision with the sibling PR.** Your description says this "leaves the three pages already covered by #2075 unchanged" — checked: #2075 (still open, unmerged) touches `agent.html`, `brief.html`, `review.html`, and `tests/test_frontend_language_storage.py` — zero filename overlap with this PR's `index.html`, `submissions.html`, `tests/test_frontend_language_storage_pages.py`. Claim holds.

**CI is green and reflects the real, current state.** `submission-validation` completed `success` on exact head `a0b9eb9bd...` (run `95147558736`), consistent with the validator's own "non-submission code/docs/test PR; participant package validation was not applicable" note — this is a real pass on the current code, not a stale/pre-fix run.

## Not in scope of this note

This is a technical correctness/reproduction check only — I'm not weighing in on whether duplicating the three helper functions across two files (vs. extracting a shared script) is the right long-term structure for this repo; that's a maintainer call, not something I'd second-guess from outside.

## Thanks

Thanks for the fast, well-scoped fix and for stating exactly which Python versions and harnesses you re-ran — that made independently reproducing both the original failure and the fix a direct, mechanical check rather than a guess.

---

CocoSgt (approved):
复核 exact head a0b9eb9bdf519c95d21be0ea38c18f9dc16179a7：旧 review 指出的 Python 3.9 f-string blocker 已闭合。Python 3.9.6 + Node 24 在原始 head 和与 latest main b5c37fda 的合成 merge tree 上均 2/2 PASS；py_compile、git diff --check、merge-tree 均通过。invalid/missing storage、浏览器语言回退、受限 localStorage 读写均由真实页面脚本覆盖；未发现新的行为或安全阻断。

PR #2733, opened 2026-08-15
E131: the register the whole argument rests on is not one of the nine files

jiangmuran:
the register the whole argument rests on is not one of the nine files

Following E130 one step further: which nine files does the review
harness read whole? Read from scripts/review_submission.py, not from
memory --

  proposal.md, manifest, metrics, assumptions, sources, self_check,
  and the three matrices.

visual/assets/ IS NOT AMONG THEM.

So the two artefacts this package leans on hardest are both outside what
a reviewer can open:

  the errata register -- 130 entries, of which the proposal names 7
  the mutation record -- 33 deliberate defects across 18 gates

A proposal whose argument is that a claim no machine can check is not
evidence had put its own central evidence one directory outside the
reader's reach and said nothing about it. That is not a misplaced file.
It is reachability of evidence -- the thing this package demands
everywhere else. The reviewer sees the prose citing the register and
cannot open the register.

Two new entries go into self_check.json, which IS inside the nine and is
read whole: the register's count, its breakdown by finder, its four
commonest defect shapes; and the harness's defect count, gate coverage,
and the number of rules currently matching nothing. Both say plainly
that the full file ships and is not among the nine the harness reads,
and both give the path. Telling the reader where the evidence is, and
that the tool will not open it for them. Every number computed at build
time.

TWO ERRORS ON ONE LINE, BOTH RECORDED. Counting how many entries the
prose names: the regex had its word boundaries eaten by a shell and
became two literal backspace bytes; and the register writes `E01` while
the prose writes `E35`, so comparing the literals would have returned
zero even had the pattern survived. Two faults, one wrong number.
Repaired, the count is 7 -- matching an independent measurement. The
line came close to shipping a false zero.

E131.


CocoSgt (approved):

PR #2231, opened 2026-08-12
fix: scan all present visual code assets

jiangmuran:
## Summary

- scan every CSS and JavaScript file present below `visual/assets/`, independent of manifest inventory
- apply the same offline-safety findings to deterministic intake and professional visual review
- reject unreadable, non-UTF-8, remote-loading, active-network, and symlinked visual code assets
- preserve one existing vendored three.js visual through a package-bound digest of its complete executable tree

## Reproduction

Before this change, remote CSS imports and JavaScript `fetch()` calls in external visual assets were not inspected. Declaring such an asset in the manifest exposed the original gap; omitting the manifest entry also bypassed the first revision of this PR even when `visual/index.html` actively loaded the file.

The scanner now derives its coverage from the files the browser can reach on disk, not from participant-declared inventory. The historical three.js compatibility entry is tied to the exact existing package and the byte digest of `visual/index.html` plus every visual asset, so any executable-tree change returns it to normal scanning.

## Validation

- `python3 -m unittest tests.test_visual_asset_safety tests.test_visual_review tests.test_submission_workflow` (157 passed)
- current-main corpus scan: 58 packages, 196 CSS/JS assets, 0 issues
- `python3 -m py_compile scripts/visual_asset_safety.py scripts/validate_submission.py scripts/visual_review.py tests/test_visual_asset_safety.py tests/test_visual_review.py`
- `git diff --check`

PR #2934, opened 2026-08-16
fix: validate SVG visual assets as XML

jiangmuran:
## Summary

- discover every on-disk `visual/assets/**/*.svg`, including nested and undeclared files
- parse SVG with the standard-library Expat XML parser and reject malformed/non-UTF-8 content, DTD/entity declarations, processing instructions, active/embed/SMIL elements, event attributes, unsafe URIs, and remote/data/script CSS resources
- apply the same scanner in both `validate_submission` and standalone `visual_review`, with symbolic-link handling that fails closed

## Why

`.svg` is an allowed visual asset, but both existing gates only inspect `visual/index*.html`. An SVG could therefore carry `<script>`, `foreignObject`, event handlers, external images, `javascript:` links, CSS imports, or XML external entities without either gate reporting it. This is the scoped SVG follow-up explicitly left outside #2231; #2231 handles CSS/JavaScript files only.

The scanner is XML-aware rather than applying HTML regexes to XML. Expanded element and attribute names cover namespace prefixes, XML character references are decoded before URI checks, DTD handling aborts before entity expansion, and local fragment/relative references remain supported.

## Compatibility

Scanned current `main@1179e7ed6`: 24 SVG visual assets across 17 packages, with zero issues. The existing corpus uses 29 local presentation references of the form `url(#id)`; those remain accepted.

## Verification

- `python3 -m unittest tests.test_submission_workflow tests.test_visual_review tests.test_svg_asset_safety` (170 passed)
- Python 3.9: `tests.test_svg_asset_safety tests.test_visual_review` (22 passed) plus `py_compile`
- `python3 -m py_compile scripts/svg_asset_safety.py scripts/validate_submission.py scripts/visual_review.py tests/test_svg_asset_safety.py tests/test_visual_review.py tests/test_submission_workflow.py`
- `git diff --check`
- current-main corpus scan: 17 packages, 24 SVG files, 0 issues

### DIFF
--- a/scripts/svg_asset_safety.py
+++ b/scripts/svg_asset_safety.py
@@ -1,0 +1,204 @@
+"""Statically validate standalone SVG visual assets without rendering them."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from urllib.parse import unquote
+from xml.parsers import expat
+
+
+SVG_NAMESPACE = "http://www.w3.org/2000/svg"
+XML_NAMESPACE = "http://www.w3.org/XML/1998/namespace"
+MAX_SVG_BYTES = 5 * 1024 * 1024
+FORBIDDEN_ELEMENTS = {
+    "script",
+    "foreignobject",
+    "iframe",
+    "object",
+    "embed",
+    "handler",
+    "listener",
+    "link",
+    "audio",
+    "video",
+    "animate",
+    "animatemotion",
+    "animatetransform",
+    "set",
+    "discard",
+}
+URI_ATTRIBUTES = {"href", "src", "data", "poster", "action", "formaction", "base"}
+URI_SCHEME_RE = re.compile(r"^[a-z][a-z0-9+.-]*:", re.I)
+CSS_COMMENT_RE = re.compile(r"/\*.*?\*/", re.S)
+CSS_LINE_CONTINUATION_RE = re.compile(r"\\(?:\r\n|[\n\r\f])")
+CSS_ESCAPE_RE = re.compile(r"\\([0-9a-fA-F]{1,6})(?:\s)?|\\(.)", re.S)
+CSS_URL_RE = re.compile(r"\burl\s*\(\s*(['\"]?)(.*?)\1\s*\)", re.I | re.S)
+CSS_IMPORT_RE = re.compile(
+    r"@import\s+(?:url\s*\(\s*)?(['\"]?)([^'\"\s;)]+)\1",
+    re.I,
+)
+
+
+class _RejectedDeclaration(Exception):
+    pass
+
+
+def _local_name(name: str) -> str:
+    return name.rsplit("}", 1)[-1].rsplit(":", 1)[-1].casefold()
+
+
+def _expanded_name(name: str) -> tuple[str, str]:
+    if "}" in name:
+        namespace, local = name.rsplit("}", 1)
+        return namespace, local.casefold()
+    return "", name.rsplit(":", 1)[-1].casefold()
+
+
+def _decode_css_escapes(text: str) -> str:
+    def replace(match: re.Match[str]) -> str:
+        if match.group(1) is not None:
+            codepoint = int(match.group(1), 16)
+            return chr(codepoint) if codepoint <= 0x10FFFF else ""
+        return match.group(2) or ""
+
+    without_comments = CSS_COMMENT_RE.sub("", text)
+    without_continuations = CSS_LINE_CONTINUATION_RE.sub("", without_comments)
+    return CSS_ESCAPE_RE.sub(replace, without_continuations)
+
+
+def _is_external_or_active_uri(value: str) -> bool:
+    stripped = value.strip()
+    if not stripped:
+        return False
+    if "\\" in stripped or any(ord(character) <= 0x20 for character in stripped):
+        return True
+    if stripped.startswith("/") or URI_SCHEME_RE.match(stripped) is not None:
+        return True
+    local_path = stripped.split("#", 1)[0].split("?", 1)[0]
+    for _ in range(2):
+        local_path = unquote(local_path)
+    return ".." in Path(local_path).parts
+
+
+def _css_has_external_or_active_uri(text: str) -> bool:
+    normalized = _decode_css_escapes(text)
+    targets = [match.group(2) for match in CSS_URL_RE.finditer(normalized)]
+    targets.extend(match.group(2) for match in CSS_IMPORT_RE.finditer(normalized))
+    return any(_is_external_or_active_uri(target) for target in targets)
+
+
+def svg_text_issues(text: str) -> list[str]:
+    reasons: set[str] = set()
+    element_stack: list[str] = []
+    style_buffers: list[list[str]] = []
+    root_seen = False
+    parser = expat.ParserCreate(namespace_separator="}")
+    parser.SetParamEntityParsing(expat.XML_PARAM_ENTITY_PARSING_NEVER)
+
+    def reject_declaration() -> None:
+        reasons.add("SVG must not contain DOCTYPE or entity declarations")
+        raise _RejectedDeclaration
+
+    def start_element(name: str, attributes: dict[str, str]) -> None:
+        nonlocal root_seen
+        namespace, local = _expanded_name(name)
+        if not root_seen:
+            root_seen = True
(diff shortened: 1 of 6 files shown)


jiangmuran:
Hardened the scanner before review in exact head `5b22174c03`:

- reads are now bounded to 5 MiB in the standalone gate rather than loading an arbitrary-size SVG before validation;
- UTF-8 BOM and `encoding="UTF-8"` remain valid, while a byte-compatible file declaring another XML encoding is rejected explicitly;
- root/child namespaces are checked from expanded names, `xml:base` is rejected, and local path traversal is decoded twice before checking;
- intermediate directory symlinks are covered in addition to final SVG symlinks.

Updated verification:
- Python 3.9 focused suite: 26 passed plus `py_compile`
- current Python full related suite: 174 passed
- current-main corpus: 17 packages, 24 `visual/assets` SVG files, 0 issues
- `git diff --check`

Scope remains deliberately limited to on-disk `visual/assets/**/*.svg`. Package-level `assets/**/*.svg` and inline SVG are separate policy surfaces and are not claimed here.

anselasimov-web (commented):
# PR #2934 review draft — technical read-through, not a maintainer decision

Hi, I'm a fellow participant in this event (**not a maintainer, no merge/approval authority**). This showed up in a scan of open tool/infra PRs from the same author with zero reviews despite passing CI, so I gave it an independent read and ran what I could locally. Everything under "empirical" below is something I actually executed against this exact head; everything under "reading-level" is code inspection without execution. Posting because a submission-validation `success` here is easy to misread as "this is fine to merge" — see the caveat near the end for why that's not quite true for this class of PR.

**Checked exact head:** `5b22174c0334bb141ff17ee0d6ae1feac68cd1bc` (full 40-char sha, matches current PR head).

## What this PR does

Adds `scripts/svg_asset_safety.py`: a standalone, non-rendering SVG validator (`expat`-based XML parse with `DOCTYPE`/entity declarations rejected outright, `<script>`/`<foreignObject>`/`<iframe>`/media/SMIL elements banned, `on*` event attributes banned, `xml:base` banned, any `href`/`src`/`data`/`poster`/`action`/`formaction` pointing to an external/absolute/data/traversal target rejected, inline `<style>`/`style=` CSS scanned for `url()`/`@import` targets with the same external-target check, UTF‑8-only, 5 MiB cap, symlinked SVG paths rejected). It's wired into both `validate_submission.py::validate_ai_package_dir` and `visual_review.py::review_visual`.

## Empirical verification

- **Ran the PR's own test suite for real.** Extracted `scripts/svg_asset_safety.py` and `tests/test_svg_asset_safety.py` verbatim from this PR's diff, `python3 -m unittest` locally: **12/12 pass** (doctype/entity rejection, encoding/root-element validation, UTF‑8 BOM handling, bounded reads, foreign-namespace/`xml:base` rejection, undeclared-asset discovery from disk, file- and directory-level symlink rejection, CSS url()/@import checks in both attributes and `<style>` elements).
- **Independent adversarial probing**, beyond the PR's own tests — 12 hand-written payloads run directly against `svg_text_issues()`:
  - XXE-style external entity (`<!DOCTYPE svg [<!ENTITY xxe SYSTEM "file:///etc/passwd">]>...&xxe;`) → rejected (`DOCTYPE or entity declarations`)
  - billion-laughs-style nested entity expansion → rejected (same path, entity declarations are refused before expansion is even attempted)
  - `<script>` element, `onload=` attribute → rejected
  - remote `<image href="http://evil.example/x.png">` and `xlink:href` on `<use>` (verified the namespace-qualified attribute still normalizes to local name `href` and is caught) → rejected
  - `@import url(...)` and `style="fill:url(...)"` pointing off-host → rejected
  - `data:` URI in `href` → rejected (matches the generic URI-scheme check, not just `http(s)`)
  - plain and double‑URL‑encoded (`%252e%252e/...`) path traversal in `href` → both rejected
  - a clean, self-contained SVG → passes with no issues (confirms the checks aren't so broad they reject legitimate content)
- **Tested, not just read, the symlink-containment claim.** My first hypothesis reading the code was that `assets_root.rglob("*")` might still descend into a *symlinked subdirectory* under `visual/assets` even though the leaf-symlink check would catch a symlinked file — i.e. a possible sandbox-escape read via a symlinked directory. I built a minimal repro (a directory symlink pointing outside the submission tree, containing an `.svg`) and ran it against the actual function: `pathlib.Path.rglob` does **not** follow symlinked directories during traversal — only the symlink entry itself is yielded, and it's the one flagged (`"visual SVG asset paths must not use symbolic links"`). My initial read was wrong; the empirical check corrected it. Good — the containment claim holds up under a targeted probe, not just under inspection.

## Reading-level notes

- The CI short-circuit reasoning from the 2026-08-16 forecast still applies verbatim to this exact head (no new commits since then): `is_non_submission_pr()` in `scripts/github_pr_validation.py` (main branch) short-circuits to `success` for any PR whose changed files are entirely outside `submissions/**`, which is why `submission-validation` is green here — **that check never executes this PR's own new code**; the scanner only starts running for real once this merges and gates a future submission PR. The empirical runs above were done manually, outside CI.
- `FORBIDDEN_ELEMENTS`/`URI_ATTRIBUTES` are a curated list, not a formally derived one; I did not attempt to prove they're a *complete* taxonomy of "active SVG content" (e.g. whether every SMIL/animation edge case is covered) — that's a security-policy-completeness question for the maintainer, not something a finite probe list can settle either way.

## Incremental observation: conflicts with author's other open PRs

I enumerated all 51 currently-open PRs by `jiangmuran` and checked which ones touch the same files as this PR. Result: **#2231** ("fix: scan all present visual code assets") independently adds a sibling scanner (`scripts/visual_asset_safety.py`, for CSS/JS assets rather than SVG) and wires it into the *exact same two integration points*:

- `scripts/validate_submission.py`: same import line (after `from metric_types import is_json_number`) and same hook site (right after the visual-package loop, before `for geometry_name in sorted(ALLOWED_GEOMETRY_FILES)`, around line 2533)
- `scripts/visual_review.py`: same import line and same hook site (right after the `FORBIDDEN_PATTERNS` loop, around line 171-173)

I confirmed this is a real, reproducible conflict rather than coincidental nearby line numbers: using `git apply --check` against a fresh checkout of current `main`, applying either PR's diff first and then attempting the other **fails to apply in both orders**, with patch failures at exactly those two files/hunks. Both PRs individually apply cleanly to current `main` on their own, so this isn't staleness on either side — it's a live conflict that will surface as soon as one of them merges. (No conflict found with #2937 — it touches `validate_submission.py` at a completely different region, around line 300 — nor with any of the other 48 open PRs from this author.)

Not something to block on, but worth the maintainer knowing before merging either #2934 or #2231, so whichever lands second isn't a surprise rebase.

## Caveat: CI doesn't execute this PR's own code

Worth stating plainly since it's easy to read `submission-validation = success` as a stronger signal than it is: for a non-`submissions/**` PR like this one, CI's short-circuit path only certifies "the changed files are outside the participant-package tree" — it never imports, runs, or type-checks `svg_asset_safety.py`, and it never runs `tests/test_svg_asset_safety.py`. The 12/12 pass above is something I ran manually on my own machine; right now, manual reading and manual execution are the only quality gate this class of tool PR gets before it starts gating other participants' submissions.

## Thanks

Thanks for including a real test suite and for the follow-up hardening comment on this exact head (bounded reads, encoding handling) — it made independent verification straightforward rather than a leap of faith.

---

CocoSgt (approved):
Maintainer review of exact head 5b22174c0334bb141ff17ee0d6ae1feac68cd1bc against current origin/main 6d48b319caa8a297d079b15351864f8153341e7b. Clean merge simulation. Independently ran the three related suites: 176 tests total PASS (the sparse worktree initially omitted the existing zymk8353 fixture; after adding that unchanged main-tree fixture, its single affected wrapper test passed). py_compile and git diff --check PASS. The scanner fails closed for malformed/non-UTF-8 XML, DTD/entities/PIs, active and foreign elements, event attributes, unsafe URI/CSS resources, size limits, and file/directory symlinks, while preserving local fragment/relative references. This closes a real gap for undeclared on-disk visual/assets SVG files without expanding participant scope or replacing existing static deliverables. Approving this exact head. Follow-up sequencing: #2231 touches the same two integration sites and must rebase after this lands; do not resolve that overlap by dropping either scanner.

PR #1056, opened 2026-08-09
fix: keep artifact viewer selection and content in sync

jiangmuran:
## Problem

Artifact rendering is asynchronous. When a user opens a slow file and immediately selects another artifact, the first request can finish last and overwrite the viewer body while the title, path, and controls still describe the latest selection. A late failure can similarly replace the latest body with an unrelated error.

## Change

- assign each `open()` call a monotonically increasing render epoch
- verify the epoch and selected path before asynchronous JSON, GeoJSON, table, or text results update the DOM
- guard asynchronous error rendering the same way
- invalidate pending renders when the viewer closes
- add a dependency-free Node regression harness with a deferred old JSON request and an immediately resolved latest Markdown request

## Verification

- `python3 -m unittest tests.test_proposal_artifact_viewer tests.test_submissions_gallery.TestSubmissionsGallery.test_human_readable_report_viewer_loads_structured_evidence -v` (2 passed)
- `node --check proposal-artifact-viewer.js`
- `python3 -m py_compile tests/test_proposal_artifact_viewer.py`
- `git diff --check`

PR #2936, opened 2026-08-16
fix: contain and restore artifact viewer focus

jiangmuran:
## Summary

- declare the full-screen artifact viewer as a labelled modal dialog
- trap forward/reverse Tab navigation inside its currently visible controls
- remember the element that opened the viewer and restore it on both Escape and close-button exits, including after previous/next navigation
- cover the focus lifecycle with a dependency-free Node DOM harness

## Problem

The viewer moved focus to its close button when opened, but closing left focus on that now-hidden button. Tab could also move into controls behind the full-screen overlay. Keyboard and assistive-technology users therefore lost their place in the artifact list and could interact with obscured page content.

The focusable set includes links, form controls, media controls, summaries, iframes, explicit tab stops, and editable content; hidden/disabled elements are excluded at event time so previous/next state and asynchronously rendered content do not stale the trap.

#1056 changes the same JavaScript file but addresses stale asynchronous render results only. This PR does not alter rendering epochs or content selection; the behaviors and regression files are separate.

## Verification

- `python3 -m unittest tests.test_proposal_artifact_viewer_focus -v` (2 passed)
- focused existing gallery viewer assertion passed
- `node --check proposal-artifact-viewer.js`
- `python3 -m py_compile tests/test_proposal_artifact_viewer_focus.py`
- `git diff --check`
- Chromium interaction: open focuses Close; Tab and Shift+Tab remain in the dialog; Escape restores the originating artifact card

### DIFF
--- a/proposal-artifact-viewer.js
+++ b/proposal-artifact-viewer.js
@@ -5,6 +5,14 @@
   const palette = ['#65b7ff', '#ffc46b', '#9d8cff', '#62d6a8', '#e08b78', '#75c7bd'];
   let config = null;
   let selectedIndex = -1;
+  let returnFocusElement = null;
+
+  const focusableSelector = [
+    'a[href]', 'button:not([disabled])', 'input:not([disabled])',
+    'select:not([disabled])', 'textarea:not([disabled])', 'audio[controls]',
+    'video[controls]', 'summary', 'iframe',
+    '[tabindex]:not([tabindex="-1"])', '[contenteditable="true"]'
+  ].join(',');
 
   const escapeHTML = value => String(value ?? '').replace(/[&<>"']/g, ch => ({
     '&': '&amp;', '<': '&lt;', '>': '&gt;', '"': '&quot;', "'": '&#39;'
@@ -358,6 +366,33 @@
     elements.forEach(element => observer.observe(element));
   }
 
+  function focusableElements(viewer) {
+    return [...viewer.querySelectorAll(focusableSelector)].filter(element =>
+      !element.hasAttribute('hidden') && element.getAttribute('aria-hidden') !== 'true' && element.getClientRects().length
+    );
+  }
+
+  function trapFocus(event, viewer) {
+    const elements = focusableElements(viewer);
+    if (!elements.length) {
+      event.preventDefault();
+      viewer.focus({preventScroll: true});
+      return;
+    }
+    const first = elements[0];
+    const last = elements[elements.length - 1];
+    if (!viewer.contains(document.activeElement)) {
+      event.preventDefault();
+      first.focus({preventScroll: true});
+    } else if (event.shiftKey && document.activeElement === first) {
+      event.preventDefault();
+      last.focus({preventScroll: true});
+    } else if (!event.shiftKey && document.activeElement === last) {
+      event.preventDefault();
+      first.focus({preventScroll: true});
+    }
+  }
+
   async function renderSelected(item) {
     const body = document.getElementById('artifactViewerBody');
     const path = item.path;
@@ -397,6 +432,10 @@
     const item = config.artifacts[selectedIndex];
     const viewer = document.getElementById('artifactViewer');
     const body = document.getElementById('artifactViewerBody');
+    if (viewer.hidden) {
+      const active = document.activeElement;
+      returnFocusElement = active && typeof active.focus === 'function' ? active : null;
+    }
     document.getElementById('artifactViewerType').textContent = config.type(path);
     document.getElementById('artifactViewerTitle').textContent = config.label(path);
     document.getElementById('artifactViewerPath').textContent = path;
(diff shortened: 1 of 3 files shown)


anselasimov-web (commented):
# PR #2936 review draft — technical read-through, not a maintainer decision

Hi, I'm a fellow participant (**not a maintainer, no merge/approval authority**). Came across this while scanning open tool/infra PRs from the same author sitting with zero reviews despite a green CI check; this is a from-scratch review (no earlier draft to build on). Everything under "empirical" is something I actually ran against this exact head; "reading-level" is inspection without execution.

**Checked exact head:** `58f2b3ea96a3f4c4c8e661ff8b1c368af25cf1ab` (full 40-char sha, matches current PR head).

## What this PR does

The `#artifactViewer` panel in `proposal-view.html` (the modal-like panel that shows a proposal's individual files) gains `role="dialog" aria-modal="true" aria-labelledby="artifactViewerTitle" tabindex="-1"`, replacing a plain `aria-label`. `proposal-artifact-viewer.js` gains an actual focus trap: `Tab`/`Shift+Tab` wrap around inside the panel's focusable controls instead of leaking focus onto the page behind it, and closing (via `Escape` or the close button) restores focus to whichever element had focus before the panel opened — tracked only on the closed→open transition, so navigating between artifacts with the prev/next buttons while the panel stays open doesn't overwrite the original "opener" with the prev/next button itself.

## Empirical verification

- Applied this PR's diff to fresh copies of `proposal-artifact-viewer.js` and `proposal-view.html` pulled from current `main`, added the PR's own new test file (`tests/test_proposal_artifact_viewer_focus.py`) unmodified, and ran `python3 -m unittest` for real. Node.js (v22.22.3) is present in my environment, so the Node-backed behavioral sub-test was **not** skipped (see caveat below on why that matters): **2/2 pass**, including the scripted Node harness that mounts the real viewer and drives `Tab` / `Shift+Tab` / background-focus / `Escape` / close-button through it, asserting wrap-around at both ends of the tab order and correct focus restoration across two separate open→navigate→close cycles.
- **Negative control**, to make sure the tests are actually testing something: re-ran the identical test file against the **pre-PR** (current-`main`) versions of both files. Both tests genuinely fail — `AssertionError: 'role="dialog" ...' not found in ...` for the markup check, and `Error: forward Tab at the last control must be trapped` for the behavior check. So the keyboard-trap gap this PR describes is real and reproducible on current `main` today, not just asserted in the PR description.
- Read `open()`/`close()` directly to confirm the "don't clobber the opener on internal navigation" behavior isn't accidental: `returnFocusElement` is only assigned inside `if (viewer.hidden) { ... }`, i.e. only on the transition from closed to open, and `close()` reads-then-clears it before calling `.focus()` and guards with `restoreFocus?.isConnected` (so a since-removed opener element won't throw). This matches what the Node harness actually exercises and what I saw pass/fail above.

## Incremental observation: author's other open PRs touching the same files

Checked all 51 of your other open PRs for file overlap with the 3 files this PR touches:

- **#1056** ("fix: keep artifact viewer selection and content in sync", opened 2026-08-09) also modifies `proposal-artifact-viewer.js`. I checked whether it would collide with this PR — it doesn't look like a conflict *caused by* #2936: `git apply --check` shows #1056's diff **fails to apply to current `main` even on its own**, independent of #2936 (patch failure at `proposal-artifact-viewer.js:341`/`369`). It looks like #1056 has drifted out of sync with `main` since it was opened and needs its own rebase regardless of this PR — flagging so it doesn't get misread as "these two are fighting over the same lines" when the more accurate read is "#1056 is stale."
- **#2064** ("fix: isolate malformed proposal image URLs") touches `proposal-view.html`, but at a different location (line ~442, the image-URL sanitizer) — applies cleanly against #2936 in both merge orders, no conflict.
- No other open PR from you touches these three files.

## Caveat: CI doesn't execute this PR's own code, and neither does every reviewer's machine

Same note as I'm posting on your other two open tool PRs today: `submission-validation = success` here only certifies the changed files sit outside `submissions/**` — CI never runs `tests/test_proposal_artifact_viewer_focus.py` at all. On top of that, the Node-backed half of that test is gated by `@unittest.skipUnless(shutil.which("node"), ...)` — it **skips silently, not fails,** when `node` isn't on `PATH`. That means a reviewer without Node locally would see "2 passed" while only the markup-string assertion actually ran; the entire keyboard-trap/focus-restoration behavior — the actual substance of this fix — would go unexercised without anyone getting a red flag about it. Worth keeping in mind if this becomes a recurring pattern across your JS-behavior PRs; not a request to change anything in *this* PR, since I did have Node available and both tests genuinely ran and passed here.

## Not evaluated

No opinion on broader modal-dialog accessibility beyond what's in scope here (e.g. whether other panels in this file share the same gap) — out of scope for this diff.

## Thanks

Thanks for shipping a real regression test with an actual negative-control-worthy assertion, rather than just the markup change — it's what made this a same-day, high-confidence review instead of a "looks plausible" skim.

---

CocoSgt (changes requested):
Blocking focus-containment gap on exact head 58f2b3ea96a3f4c4c8e661ff8b1c368af25cf1ab. The parent-document keydown trap works for the header/body controls covered by the Node harness, and my latest-main merge simulation plus 3 focused tests, node --check, py_compile, and diff-check all pass. However, renderSelected() creates real PDF and sandboxed HTML iframes. Once keyboard focus enters an iframe, its Tab keydown occurs in the child browsing context and does not bubble to the parent document listener. When sequential focus exits that frame, it can therefore advance to focusable page content after #artifactViewer; the PR has no focusin guard, inert background, native modal top-layer behavior, or sentinels to pull it back. I reproduced the missing boundary in a minimal parent model: after an iframe-exit transition, activeElement remains the background because the implementation registers only keydown. This contradicts the stated guarantee that keyboard users cannot interact with obscured content. Please add a browser-effective containment mechanism (for example a document focusin guard while open, with careful close/restore ordering, or an equivalent inert/native-modal design) and a regression that models or drives focus leaving both the PDF and HTML iframe paths. The regression must fail on current main and this exact head, and pass after the fix; preserve opener restoration and dynamic focusable recomputation.

PR #2937, opened 2026-08-16
fix: align trusted download and validation limits

jiangmuran:
## Summary

- derive one `MAX_SINGLE_FILE_BYTES` value from every validator per-file limit
- use that shared value as the trusted PR hydrator download cap
- keep boundary tests small by exercising explicit 4-byte/5-byte responses while separately asserting the default contract

## Problem

The local validator permits video files up to 20 MiB, but trusted PR hydration hard-coded a 10 MiB per-file download cap. A valid 10–20 MiB video package therefore passed local validation and then raised `file exceeds download cap` before CI could produce its deterministic validation report.

The new derived constant removes the duplicated limit rather than assuming video will always remain the largest file type. If any validator per-file maximum changes later, trusted hydration follows it automatically. The existing 40 MiB package total and all type-specific validation remain unchanged.

## Verification

- reproduced before the fix with a valid 15 MiB media fixture: local validator PASS, trusted hydration rejected at 10 MiB
- `python3 -m unittest tests.test_submission_workflow` (148 passed)
- equal-to-cap response writes successfully; cap+1 response raises and leaves no partial destination
- `python3 -m py_compile scripts/validate_submission.py scripts/github_pr_validation.py tests/test_submission_workflow.py`
- `git diff --check`

No open PR touching `github_pr_validation.py` references `MAX_VIDEO_BYTES` or otherwise resolves this 10/20 MiB contract mismatch.


jiangmuran:
已同步到当前主线 `f1a58798c1cf226036345bd4c42fda568e185b9c`，新 exact head 为 `a8fd73f1beef290d2a53f1b2b4d32a7ed63acf08`。本次仅重放原单提交以纳入刚合并的主线；有效修复 diff 未扩大。\n\n验证：`python3 -m unittest tests.test_submission_workflow -v` 148/148 PASS，`py_compile` 与 `git diff --check upstream/main...HEAD` 均通过。

anselasimov-web (commented):
# PR #2937 review draft — technical read-through, not a maintainer decision

Hi, same as on #2934 — I'm a fellow participant (**not a maintainer, no merge/approval authority**), doing an independent read of open tool/infra PRs from the same author sitting with zero reviews. This one's head changed since an earlier CI-forecast-only pass I did on 2026-08-16 (old head `848d717e772da1ad1d529c481f3bba49f9c0fabbd`), so this is a full recheck against the current head, not a copy-forward of that earlier note.

**Checked exact head:** `a8fd73f1beef290d2a53f1b2b4d32a7ed63acf08` (full 40-char sha, matches current PR head).

## Why the head changed, and whether the fix is still the same fix

Your 2026-08-16T10:35Z comment says the push was a rebase onto main (to `f1a58798c1cf226036345bd4c42fda568e185b9c`), replaying the same single commit, with the effective diff not widened. I didn't take that on faith — I re-read the new diff in full against the old draft's description and confirm the substance is unchanged: same three files (`scripts/github_pr_validation.py` +2/-1, `scripts/validate_submission.py` +11/-0, `tests/test_submission_workflow.py` +17/-7), same fix shape. Only the base commit moved.

## What this PR does

`scripts/validate_submission.py` gains `MAX_SINGLE_FILE_BYTES = max(MAX_MARKDOWN_BYTES, MAX_JSON_BYTES, MAX_GEOJSON_BYTES, MAX_ASSET_BYTES, MAX_VIDEO_BYTES, MAX_AUDIO_BYTES, MAX_DRAWING_BYTES, MAX_HTML_BYTES, MAX_VISUAL_ASSET_BYTES)`. `scripts/github_pr_validation.py` swaps its previously-hardcoded `MAX_DOWNLOAD_BYTES = 10 * 1024 * 1024` for this derived constant. The bug this fixes: the local per-file validator already allows videos up to `MAX_VIDEO_BYTES` (20 MiB), but the trusted-hydration download step (used when CI fetches a fork's files for the deterministic-validation gate) was capped at a separately-hardcoded 10 MiB — so a video-containing package that passed local validation could still get rejected mid-hydration with `file exceeds download cap`, before validation even ran. Tying the download cap to the same source of truth as the per-file caps closes that gap and prevents future drift if any individual cap changes later.

## Verification

- **Reading-level:** confirmed both call sites of `download_content()` (the manifest-driven per-file hydration path, and the "download every file changed in the PR" path used for the readiness-contract check) use the default `max_bytes=MAX_DOWNLOAD_BYTES` with no tighter override — so both benefit uniformly, and this change doesn't silently widen some other, unrelated cap past what the validator itself already allows. Confirmed `MAX_VIDEO_BYTES` (20 MiB) is the largest of the nine caps folded into `MAX_SINGLE_FILE_BYTES`, so the new download cap is exactly the widest per-file limit already in force, not a step beyond it.
- **Empirical:** applied this exact diff to a clean copy of `scripts/validate_submission.py` + `scripts/github_pr_validation.py` (plus their non-test import closure) taken fresh from current `main`, and called the changed code directly rather than just reading it:
  - `MAX_DOWNLOAD_BYTES == MAX_SINGLE_FILE_BYTES == MAX_VIDEO_BYTES == 20971520` — holds.
  - `download_content(..., max_bytes=4)` accepts a 4-byte payload and writes it byte-for-byte; a 5-byte payload raises `RuntimeError: ...: file exceeds download cap` and leaves no partial file on disk. This is the exact boundary your rewritten tests (`test_download_content_accepts_file_at_explicit_boundary` / `..._rejects_file_over_explicit_boundary`) assert, and it holds under a fresh, independent call — not just by reading the assertion.
  - I did **not** re-run the full `tests/test_submission_workflow.py` end-to-end (3.8k lines, broad fixture/dependency surface not worth reconstructing for three test cases) — your comment reports 148/148 locally on the prior head. I'm not independently re-verifying that count this round; treat it as author-reported, not confirmed by me. The three specific assertions above are what I actually re-derived myself.

## Incremental observation: no conflict with sibling PRs

Since #2934 and #2231 (also open, also from you) both touch `scripts/validate_submission.py`, I checked whether this PR collides with either. It doesn't: `git apply --check` in both merge orders shows this PR's edit (constants block, ~line 300) applies cleanly on top of either #2934 or #2231's already-applied changes (their edits are at line ~21 and ~2533, well clear of this one). No other file overlaps found against the rest of your 48 other open PRs.

## Caveat: CI doesn't execute this PR's own logic

Same note as on #2934: `submission-validation = success` here only reflects the non-`submissions/**` short-circuit — CI never exercises the download-cap change itself, either now or after merge (it only takes effect inside a *future* trusted PR's hydration step, which isn't something the current run tests). The boundary checks above were run manually, outside CI; that's presently the only verification this logic gets before it starts gating how large a video other participants can submit.

## Thanks

Thanks for the quick rebase and for explicitly calling out in your comment that only the base moved, not the diff — that made re-verifying head `a8fd73f1...` a targeted diff-and-execute pass instead of a from-scratch re-review.

---

CocoSgt (approved):
Maintainer review of exact head a8fd73f1beef290d2a53f1b2b4d32a7ed63acf08 against current origin/main 01563701b4d41a15c1dd10178057a8a6d04f7b62. Clean merge simulation, preserving the already-merged SVG gate. Independent verification: tests.test_submission_workflow 151/151 PASS; py_compile PASS; git diff --check PASS. A real default-cap probe accepted a 15 MiB payload, while MAX_DOWNLOAD_BYTES+1 was rejected before creating a destination; MAX_DOWNLOAD_BYTES equals the validator MAX_VIDEO_BYTES at 20 MiB and MAX_TOTAL_BYTES remains 40 MiB. This removes the local-versus-trusted false rejection without widening any participant file-type or package-total contract. Approving this exact head.

PR #3633, opened 2026-08-21
京张水准线 · 图纸字号下限：0.53% → 0.59%，下限放在选字号的地方

jiangmuran:
The five PNGs the reviewing model receives are read as a 1600-unit-wide image, and there the quantity that decides whether a label survives is its height **as a fraction of the frame** — which no paper rule measures. `accessibility_qa` measures millimetres at A0 and reported all-pass throughout: this package set its smallest labels at 8.5/1600 = **0.53%** of frame width, against 0.81% in the strongest package in the field, with half of every legend on the five sitting in the bottom two steps.

Floor set at **9.5** — it lifts the bottom two steps to the size the third step already uses, so nothing is now set larger than something already laid out and fitted. A bigger floor is a relayout, not a floor.

**The floor is held where a size is chosen, not after the drawing is finished.** The first attempt raised sizes in the finished SVG, and `wrap` had already broken every paragraph at the old size — eight rejection cards on FIG.06 wrapped for 9, drew at 9.5, and ran off the right edge of an image the reviewer receives. `txt`, `wrap` and `plated` now read one constant, so a line is measured at exactly the size it is set in.

`plated` is worth naming: its docstring promises that "what the reader sees and what the gates measure stay one thing", and it sized the plate at the size passed in while drawing at the floored size — labels stood 21.6 units past their own paper and the rule the plate exists to break ran through the overhang. That promise was kept by two numbers happening to agree; it is now kept by a line of code.

Two more emitters had sizes of their own and were missed by the same rule: `fig_basemap` writes its own `<text>` pair for the cased street names, and the `f` that names the closure error on the identity mark. Six sheets and two sheets respectively went on reporting 4.47 and 4.73 mm after the floor had been raised everywhere else.

Four sheets broke under the larger type, each repaired at the cause the gate named — a scale note and a key block positioned against different anchors with nothing joining them; map labels alternating baselines by chain order, which separates neighbours and nothing else; two English labels wider than the band that held them; and a card block needing two units that could come from neither of the two gaps that looked like slack.

**The floor made a proven gate unfailable, which is the most important line here.** M13 existed to show `accessibility_qa` fails: press one body run on FIG.27 to 2.0 pt. The floor absorbed it and the gate passed on a package the mutation believed it had broken. The rule is truer than before, but the *reason* it holds changed — from every author typing a large enough size to one constant enforcing it — so the mutation now strikes the constant.

Prose in both editions is rewritten rather than renumbered: the floor abolished the distinction the sentence was drawing (minimum 4.47 against a usual 4.73). All 69 sheets now sit at **4.99 mm**, and the sentence says it is a floor rather than a worst case.

Gates: `label_collision_qa` 0 pairs, `frame_collision_qa` 0 runs, `overflow_qa` 0 across 107 sheets, `accessibility_qa` 0 failures, `computed_value_qa` 43 readings all equal to what the build derives.


CocoSgt (approved):

PR #3650, opened 2026-08-21
京张水准线 · 展板排的是十二张，正文说的是三十四张

jiangmuran:
`build_drawings.SHEETS` is **34** — the A3 booklet carries every sheet. `BOARD_SHEETS` is **six boards of two, twelve distinct sheets**. Both editions said, at `proposal.md:2009` and `proposal.en.md:1891`, that the thirty-four compose into **both** products. True of the booklet, wrong about the boards by nearly three times — on `proposal.md`, which the reviewing model receives whole.

**The true number had been printing in a gate readout the whole time.** `board_page_reach_qa` reports "the first board page prints 2 sheet(s)" on every build. Nothing joined that reading to the sentence — two independently derived copies of one fact, never compared.

Shape: `deliverable-not-inspected-before-shipping`. **Not** `claim-outlived-the-package`, which would assert the sentence was once true and the package moved beneath it; `analysis/` is gitignored and I cannot establish that. Only what is checkable.

**The fix**: both editions now state 34 into the booklet and twelve selected onto the six boards, two to a board, with both numbers **derived** from `SHEETS` and `BOARD_SHEETS` and the prose saying so. English cost 5 bytes (headroom 27 → 18). New gate `board_claim_qa`, wired into `build_all`, every figure in its readout read from the build rather than typed. New mutation M183 adds a sheet that is on no board, taking the boards to thirteen.

**The gate and the mutation could not fail each other, for three different reasons in succession — and a fourth appeared while fixing the third. Each nearly shipped as proof that the gate works.**

1. M183 first added a sheet already on board 1. The gate counts *distinct* sheets, so the measured quantity never moved.
2. M183 then appended at the `board(...)` call site while the gate reads the module-level `BOARD_SHEETS`. The two were looking at different objects.
3. The gate checked that a clause held *two number-words*, never that they equalled the build. **No change to the build could fail it** — `constraint-that-cannot-be-failed`, written into the gate whose entire purpose is catching a claim that drifted from the build.
4. Fixing that, the numeral maps held only 12 and 34, so the gate died on `KeyError: 13` under the damage it guards. A check that crashes reports nothing a reader can act on. Word forms are now derived for any count either list can take.

**M69 was disarmed by a refactor I had certified inert.** It anchors on the literal line `cy_ = py_ + 20 + 11 * cap_lines + 9`; naming that leading `STAT_LEAD` left the anchor matching nothing. I had verified that refactor by proving both editions' SVG were **byte-identical** — and that is the wrong quantity: mutations anchor on source text, not output. A change can draw exactly the same picture and still silently disarm the proof that a gate works. Re-anchored.

Consequences, each caught by a gate that reads the build: `generation_chain_qa` on the gate-script count in `copyright_statement.md` (124 → 125), and `gate_count_qa` on six sites across both editions (127/126 → 128/127, 184 → 185 mutations), all six byte-neutral.

Gates: `board_claim_qa` clean, and exits 1 under M183 naming the missing count; `gate_mutation_qa` all pass; `gate_count_qa` 185 mutations and 127 of 128 gates proven able to fail, every figure read from `gate_mutation.json` rather than trusted.


147228:
### Exact-head reproducibility note

I read the current exact head `f72365cc8328501e9385b778c974e22e02e5b92c` against base `a6c25f58a8c7d5b9fd33e8dbd9838865376c489b`.

The prose and shipped readouts now name a new `board_claim_qa` gate and say it is wired into `build_all`, but the exact submitted tree contains no `board_claim_qa.py`, `build_all`, or other corresponding implementation. The changed-file list likewise contains only package records, prose/rendered outputs, manifest, and evidence JSON; the tree search finds no implementation path. A maintainer or later reviewer therefore cannot reproduce the new gate or the M183 failure from the exact head, even though `self_check.json` and `gate_mutation.json` record the result.

Please either:

- ship a minimal offline runner/fixture under the permitted package scope and register its hash in the manifest/self-check; or
- explicitly label the `board_claim_qa`/M183 result as author-local evidence and avoid presenting the 127/128 mutation count as independently reproducible gate coverage.

The underlying 34-booklet / 12-board claim may be correct; this comment is only about the evidence boundary. It is not a score, merge, publication, or ranking conclusion.

CocoSgt (approved):
Maintainer intake decision: Review Agent score 77/100. Mandatory rejection and all four local gates passed. Accepted for repository intake only; this is not gallery publication, award selection, implementation approval, or government endorsement.

PR #3653, opened 2026-08-21
京张水准线 · 折行不许切开的东西：标识符那一半从来不成立

jiangmuran:
Rasterising **page 1 of `a0-boards.pdf`** — one of the four PDF pages the reviewing model receives — and reading it showed FIG.10's footer wrapping `geometry/public_space.geojson` as `…geometry/public_spac` / `e.geojson…`. A reader who copies that path gets a file that does not exist; a model reading the page sees two tokens, neither of which resolves.

**The defect is the gate, not the footer.** `number_break_qa` has enforced "a wrap may not cut a token that must be read whole" since FIG.16 printed `900–1` on one line and `350 mm` on the next — **for numerals only**. The rule was written, proven with a mutation, and never extended to identifiers, which are the class where a cut destroys the thing rather than obscuring it: a split numeral still reads as two numbers; a split path is not a path. Shape: `check-measured-the-convenient-thing` — numerals are the easy class to detect and the check took it.

**Both halves of the gate were numeral-shaped, not just the pattern.** `strings()` kept only strings containing a digit, and `geometry/public_space.geojson` has none — so the string that was actually breaking had never been in this gate's corpus. Widening the pattern without widening the selector would have changed nothing.

- Gate now carries a `TOKEN` class (paths, filenames, registered ids) and a selector to match. Corpus **291 → 394 strings**, **18,915 → 25,610 wrappings**. It reported **20 identifier breaks and 0 numeral breaks** the moment it was asked — the written half has held since it was written, the unwritten half was never true. Among them `metrics.js|on`, `ncsti.go|v.cn`, `B|M-0`, `GPT/Code|x`.
- **Fixed in `wrap`, not in twenty strings.** `_cjk_units` already made numerals atomic; identifiers now are too. All twenty at once, and every future one.
- **M184** removes only the identifier branch of `_ATOMIC`. **M161 already fails this gate, which is exactly why M184 is needed**: M161 replaces `_cjk_units` with `list(text)` and cuts both classes, so if the identifier branch were deleted M161 would still fail via numerals and **the half added in this round would be proven by nothing**.

**One false positive was mine.** The gate compares a space-stripped join, so `BM-1 Zhongzhiyuan` glues into `BM-1Z` and `[A-Z0-9]+` matches an identifier that exists nowhere in the text. `wrap` sees the real string with its space and never had the problem. Pattern guarded with `(?![a-z])`.

The readout now states its scope: this holds the property on `wrap`, over every string the package translates at every width its sheets use; it does not read the rendered drawings, where a table column of stacked numbers is indistinguishable from a split one.

Also: merging #3650 moved the corpus 935 → 938; one site per edition, byte-neutral in both.


CocoSgt (approved):

PR #3833, opened 2026-08-23
test: decode peer catalog output as UTF-8

dvd233:
## Summary

Decode the peer-catalog child process as UTF-8 at the one test boundary that parses its JSON output. This removes a deterministic Windows/cp936 test-harness error without changing `read_peer_proposals.py`, repository policy, schemas, submissions, source data, or Gallery output.

## Exact revision

- Frozen `upstream/main`: `94fc9ea0987db897e1b5306939d693f4428f692c`
- PR creation base: `a52b4d9c4448d48a845e623faf6daef85bcd5707`
- Exact head: `4df913171b484ab47698724bf4f396e574a212e8`
- Commit tree: `e7d0bc2ae23ad5b805f2c0352d93d58ef51d6b33`
- Changed-file blob: `6119dcf2f471f926e1acd56f30191972d416f100`
- Patch SHA-256: `1f4b8b0fd9ea617f49c17303829c6a2a2e5acf629fed3ad1a21598c9cffca72e`

## Before / after

**Before:** on Windows/Python 3.10 with preferred encoding `cp936`, `test_peer_catalog_reads_local_index_without_materializing_media` invoked `subprocess.run(text=True, encoding=None)`. The child correctly emitted UTF-8 JSON, but the parent reader thread attempted GBK decoding, raised `UnicodeDecodeError`, set `stdout=None`, and the assertion path then raised `TypeError`. The exact test failed three consecutive times; the focused module reported 7 passes and 1 error.

**After:** that invocation explicitly uses `encoding="utf-8"`, matching the child process's documented and tested output contract. The test parses the same JSON and retains all existing assertions.

## Root cause and non-duplication

Merged #879 intentionally made the **child** write UTF-8 so valid catalog text such as `²` is not lost on GBK terminals; this PR credits and preserves that behavior. The remaining defect is the separate **parent test harness** decode boundary. Searches for the exact test name, `read_peer_proposals` + encoding/UnicodeDecodeError, peer catalog + GBK, and lightweight-flow encoding found no same-root open or closed fix. Open #2232 protects peer-cache write locations and does not change this invocation.

## User and public value

Windows contributors and maintainers can run the lightweight participant-flow suite without a false locale-dependent error. Keeping the real JSON catalog test executable also preserves coverage that reading the local index does not materialize peer media.

## Scope

- `tests/test_lightweight_participant_flow.py`: +1/-0
- Production changes: none
- Submission changes: none
[...]

PR #3834, opened 2026-08-23
fix: decode the remaining script subprocesses as UTF-8

Yi-luo-hua:
## Scope

`subprocess.run(text=True)` without `encoding` decodes child output with the locale encoding. On a Chinese Windows console (GBK/CP936) that corrupts, or raises `UnicodeDecodeError` on, output the child wrote as UTF-8.

This is the same defect as #420, #874 and #1203. #1515 fixes four maintainer scripts; this PR covers the remaining call sites under `scripts/`.

Most of these children are not Python, so they only need an explicit decode:

| File | Child | Why it matters |
| --- | --- | --- |
| `bootstrap_participant_workspace.py` | git / gh | Runs on participant machines — exactly the Windows population affected |
| `backfill_bilingual_artifacts.py` | `tesseract -l chi_sim` | stdout **is** the recognised Chinese text |
| `git_blob_hashes.py` | `git rev-parse --show-toplevel` | A repository path can contain non-ASCII characters |
| `ai_review_submission.py` (×2) | pdftoppm, headless Chromium | Diagnostics may be localised |
| `export_review_packet.py` (×2) | wkhtmltopdf, headless Chromium | Diagnostics may be localised |

`github_pr_validation.py` is the one recursive Python call — it runs the trusted review gates through `sys.executable` and parses their JSON — so it also propagates `PYTHONUTF8` / `PYTHONIOENCODING`, the way #1515 does. Passing those to a git or tesseract child would be meaningless, so the other call sites do not.

After this and #1515, no `subprocess.run` in `scripts/` decodes text without declaring an encoding.

## Tests

`tests/test_subprocess_encoding_scripts.py` adds an AST guard that walks every `subprocess.run` in `scripts/` and fails when one decodes text without an encoding. Its allowed set is the four files #1515 is fixing, and it asserts a **subset**, so it passes both before and after that PR merges while still failing on any new violation. Reverting any fix in this PR makes it fail.

Two behavioural tests cover the helpers that run without external binaries: the bootstrap helper round-trips Chinese through stdout and through a failure's stderr detail, and `git_blob_sha256` resolves a repository whose path contains Chinese characters.

## Risk and compatibility

Additive keyword arguments only; no signature, return-type or behaviour change on a UTF-8 host. `errors="replace"` keeps a malformed byte from aborting a review run. No overlap with #2087 (`render_proposal_html.py`, `spatial_review.py`) or #2198 (`maintainer_review.script_path`).

Out of scope, left for their owners: `contrib/` is a contributor-owned analysis tool tied to its own issue, and the `tests/` harness has the same pattern in many places.

## Commands run

```bash
python3 -m unittest tests.test_subprocess_encoding_scripts        # 5 tests, OK
python3 -m unittest tests.test_git_blob_hashes                    # 7 tests, OK
python3 -m unittest tests.test_lightweight_participant_flow       # 8 tests, OK
python3 -m unittest tests.test_export_review_packet               # 4 tests, OK
python3 -m unittest tests.test_ai_review_submission               # 20 tests, OK
python3 -m unittest tests.test_trusted_review_scripts             # 5 tests, OK
python3 -m unittest tests.test_submission_validation_concurrency  # 2 tests, OK
python3 -m py_compile scripts/{ai_review_submission,backfill_bilingual_artifacts,bootstrap_participant_workspace,export_review_packet,git_blob_hashes,github_pr_validation}.py
git diff --check
```

`tests.test_bilingual_backfill` reports one pre-existing error — `RuntimeError: Arial Unicode font is required for bilingual figure panels` — which reproduces identically on unmodified `main` and is unrelated to this change.

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/scripts/ai_review_submission.py
+++ b/scripts/ai_review_submission.py
@@ -211,6 +211,8 @@ def render_pdf_previews(
                 ],
                 capture_output=True,
                 text=True,
+                encoding="utf-8",
+                errors="replace",
                 check=False,
                 timeout=60,
             )
@@ -265,6 +267,8 @@ def render_html_previews(
                 ],
                 capture_output=True,
                 text=True,
+                encoding="utf-8",
+                errors="replace",
                 check=False,
                 timeout=60,
             )
--- a/scripts/backfill_bilingual_artifacts.py
+++ b/scripts/backfill_bilingual_artifacts.py
@@ -272,6 +272,8 @@ def ocr_image(path: Path, source_language: str) -> str:
             ["tesseract", str(path), "stdout", "-l", language, "--psm", psm],
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             check=False,
         )
         if completed.returncode != 0:
--- a/scripts/bootstrap_participant_workspace.py
+++ b/scripts/bootstrap_participant_workspace.py
@@ -97,6 +97,8 @@ def run(command: list[str], *, cwd: Path | None = None) -> str:
         cwd=cwd,
         capture_output=True,
         text=True,
+        encoding="utf-8",
+        errors="replace",
         check=False,
     )
     if completed.returncode:
--- a/scripts/export_review_packet.py
+++ b/scripts/export_review_packet.py
@@ -911,6 +911,8 @@ def render_pdf(html_path: Path, pdf_path: Path, engine: str = "auto") -> None:
                 [binary, "--quiet", "--print-media-type", html_path.as_posix(), pdf_path.as_posix()],
                 capture_output=True,
                 text=True,
+                encoding="utf-8",
+                errors="replace",
                 check=False,
             )
         elif candidate == "chromium":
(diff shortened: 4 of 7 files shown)


CocoSgt (approved):
Approved exact head `abfa2ae6e297e909d7bef0769f3e8467284fdc86`. The added UTF-8 decode boundaries and replacement-error handling close locale-dependent subprocess failures without changing participant scope, review gates, schemas, scoring, or data. The recursive trusted Python invocation also pins Python UTF-8 output rather than trusting the host locale. Independent macOS/Python 3.9 review passes all 51 targeted tests (49 script/behavior tests plus 2 workflow fixtures after materializing the sparse-checkout files), py_compile and git diff --check pass, the worktree is clean, and exact-head required CI is SUCCESS on direct current main.

CocoSgt:
Merged normally as `ec3a331aa203ca54be34263da80d5f8ea7e6d353` after exact-head required CI succeeded and protected review state was CLEAN/APPROVED; no bypass was used. Post-merge readback confirms exact head `abfa2ae6e297e909d7bef0769f3e8467284fdc86` is an ancestor of current main. The final combined main regression after companion #3833 also passes 66/66 targeted tests, py_compile and git diff --check pass, and the isolated verification worktree is clean.

PR #3811, opened 2026-08-22
Fix intake fail-open when multimodal review requests changes

CocoSgt:
Closes #3809.

The queue now requires both `recommendation == formal-review-ready` and `can_enter_formal_review == true` before score-based acceptance. This preserves the mandatory rejection and four authoritative local gates while making the required multimodal visible-quality verdict fail closed.

Regression coverage includes exact-threshold and high-score reviews that are not ready.

Tests: `python3 -m unittest tests.test_auto_review_queue tests.test_ai_review_submission` (34 passed); `python3 -m py_compile scripts/auto_review_queue.py`; `git diff --check`.

PR #3859, opened 2026-08-23
fix: keep auto review queue importable and lockable on Windows

dvd233:
## Problem

On Windows, `import auto_review_queue` fails at module import time:

```
scripts\auto_review_queue.py:48: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
```

Because two test modules import it, a Windows contributor cannot even collect them:

```
ERROR tests/test_auto_review_queue.py
ERROR tests/test_subprocess_encoding.py
!!!!! Interrupted: 2 errors during collection !!!!!
```

(reproduced on Python 3.10.8 / Windows 11 against current `main` `68f39f2883`; see also the earlier Windows portability rounds #1515, #3833, #3834.)

## Fix

- Guard `import fcntl` and fall back to `msvcrt` on Windows.
- Extract `acquire_worker_lock()`: POSIX keeps the exact existing `fcntl.flock(LOCK_EX | LOCK_NB)` behavior; Windows takes the same non-blocking, held-until-close lock via `msvcrt.locking` on one byte of `.worker.lock`.
- Lock contention raises the same `WorkerError("another auto-review worker is already running")` on both platforms (on Windows any `OSError` from preparing or taking the byte lock, including `PermissionError`, means contention).
- The lock keeps its previous lifetime: released on file close / process exit; no behavior change on Linux.

## Tests

New regression test `test_worker_lock_rejects_second_holder_until_released` proves contention is rejected and the lock can be reacquired after release, on both platforms.

Windows verification:

- `tests/test_auto_review_queue.py` + `tests/test_subprocess_encoding.py`: previously uncollectable, now **21 passed, 2 subtests passed**.
- Full suite on Windows: no new failures attributable to this change (remaining failures reproduce identically on unmodified `main`; they are pre-existing Windows-only issues in unrelated modules).

## Scope and history

Two files: `scripts/auto_review_queue.py`, `tests/test_auto_review_queue.py`. No workflow, schema, scoring, or `submissions/` changes.

Refs #526: an earlier broader PR also replaced this `fcntl` usage, but was closed unmerged because it additionally changed participant-forbidden repository infrastructure (`.gitattributes` etc.) and the author withdrew it. This PR is the minimal, participant-scope version of just the import/lock fix.

### DIFF
--- a/scripts/auto_review_queue.py
+++ b/scripts/auto_review_queue.py
@@ -510,7 +510,9 @@ def main() -> int:
     if args.concurrency < 1:
         raise WorkerError("--concurrency must be at least 1")
     args.audit_root.mkdir(parents=True, exist_ok=True)
-    lock_file = (args.audit_root / ".worker.lock").open("w", encoding="utf-8")
+    # Keep the lock byte intact so a second Windows worker reaches the
+    # non-blocking byte-range lock instead of failing while truncating it.
+    lock_file = (args.audit_root / ".worker.lock").open("a+", encoding="utf-8")
     acquire_worker_lock(lock_file)
     candidates = queued_prs(args.repo, args.label, repo_root)
     selected = []
--- a/scripts/auto_review_queue.py
+++ b/scripts/auto_review_queue.py
@@ -45,7 +45,6 @@ from __future__ import annotations
 
 import argparse
 from concurrent.futures import ThreadPoolExecutor, as_completed
-import fcntl
 import json
 import os
 import shutil
@@ -55,6 +54,14 @@ import threading
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
+
+try:
+    import fcntl
+except ImportError:  # Windows has no fcntl; the worker lock uses msvcrt there.
+    fcntl = None
+    import msvcrt
+else:
+    msvcrt = None
 
 from generate_submissions_data import package_sha256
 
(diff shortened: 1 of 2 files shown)


CocoSgt (changes requested):
Exact head dd97a936023ba7b9dbac3075a02e3af587ea12d9 passes the targeted Linux regression suite (21 tests), but the Windows contention path is still not normalized as claimed. `main()` opens `.worker.lock` with mode `"w"` before calling `acquire_worker_lock()`. The new test explicitly notes that Windows refuses this truncating open while another process holds the byte lock; that `PermissionError` therefore escapes before the helper can translate it to `WorkerError("another auto-review worker is already running")`. Please either open the lock file in a non-truncating mode that both processes can acquire (for example `a+`, with the helper seeking as needed), or wrap the open itself and translate its Windows contention `OSError`; then add a test exercising the same open path used by `main()`, not only two already-open handles.

CocoSgt (approved):
Re-reviewed exact head `c9f49c091263f50c3760f14712e67e9b25e4907b`; the prior Windows contention blocker is closed.

The production path now opens the lock non-destructively inside `acquire_worker_lock(path)`, returns the held handle, closes it on failed acquisition, and translates contention to the existing `WorkerError`. The regression test invokes that same path for acquire, second-holder rejection, release, and reacquisition.

Validation:
- exact-head `submission-validation`: SUCCESS (run 32658788749)
- focused PR tests: 21/21 passed
- clean synthetic composition with P0 #3811: 43/43 focused tests passed

No workflow, scoring, schema, submission, or protection changes.

CocoSgt:
Merged normally as `a166a0b9f16d9312d26f2723c080d9761ff999bc` after exact-head approval and required CI; no admin bypass was used.

Post-merge on current `origin/main`:
- `tests.test_auto_review_queue` + `tests.test_subprocess_encoding`: 21/21 passed
- `py_compile`: passed
- `git diff --check`: passed
- verification worktree: clean

Operational boundary: this fixes Windows import/worker-lock portability only. The trusted review worker remains paused until the separate fail-closed fix #3811 receives valid non-author approval, merges normally, and the worker clone is synchronized and read back.

PR #3934, opened 2026-08-24
test: make bootstrap workspace decoding fixtures emit UTF-8 on any locale

dvd233:
## Problem

`tests/test_subprocess_encoding_scripts.py::BootstrapWorkspaceDecodingTests` fails on a Windows machine whose console uses a legacy (cp936) code page:

```
UnicodeEncodeError: 'gbk' codec can't encode character '\xb2'
```

The two tests spawn a child `python -c` that writes the Chinese sample text. The child's stdout/stderr default to the locale encoding on Windows, so the child fails (or emits GBK bytes) before the parent's UTF-8 decoding — the actual property under test — is even exercised. Same test file passes on Linux and on Windows machines with a UTF-8 default.

## Fix

Reconfigure the child streams inside the fixture scripts (`sys.stdout.reconfigure(encoding='utf-8')` / `sys.stderr.reconfigure(encoding='utf-8')`), making the child emit UTF-8 deterministically on any locale. The tests still verify exactly what they are meant to: `bootstrap_participant_workspace.run()` decodes UTF-8 child output intact, including in the error path.

## Verification

Windows 11, Python 3.10.8, cp936 console:

- Without this change: `2 failed, 3 passed` (both bootstrap decoding tests fail).
- With this change: `5 passed`; the full module plus `test_subprocess_encoding.py` pass (`13 passed`).

One test file changed; no production code, schema, workflow, or submission content touched.

### DIFF
--- a/tests/test_subprocess_encoding_scripts.py
+++ b/tests/test_subprocess_encoding_scripts.py
@@ -87,14 +87,24 @@ class BootstrapWorkspaceDecodingTests(unittest.TestCase):
     """bootstrap_participant_workspace.run() drives git/gh on participant machines."""
 
     def test_utf8_stdout_is_decoded_intact(self) -> None:
-        script = f"import sys; sys.stdout.write({SAMPLE_TEXT!r})"
+        # The child must emit UTF-8 bytes deterministically; on a Windows
+        # console with a legacy code page, Python's stdout defaults to the
+        # locale encoding and cannot even encode the sample. Reconfiguring
+        # the child's stream keeps this test about the parent's decoding.
+        script = (
+            f"import sys; sys.stdout.reconfigure(encoding='utf-8');"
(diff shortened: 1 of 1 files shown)


CocoSgt (approved):
Exact head `4f8002dc71edada24d4c595a37ccd3209666ccd2` fixes the reproduced cp936-only fixture failure without changing production code: the legacy child script exits with `UnicodeEncodeError` under `PYTHONIOENCODING=cp936`, while the patched child streams emit UTF-8 deterministically so the parent decoding path is actually exercised. Independent Python 3.9.6 verification passed both related modules 13/13 under simulated cp936 and 13/13 under the normal locale; `py_compile`, `git diff --check`, and the clean latest-main merge tree also pass. Required exact-head CI is SUCCESS, and the one-file scope changes no schema, workflow, participant boundary, scoring, or submission content.

PR #3821, opened 2026-08-23
feat: expose synthetic capacity gates in review carriers

dvd233:
## Unique theme

Expose the existing `SCN06-C010204-SYNTHETIC-CAPACITY-01` E2 capacity envelope as one bilingual admit/reject decision diagram in the existing metrics carrier, then propagate that same figure into the existing A3/A0 review PDFs. This is a carrier and reviewer-route improvement, not a new capacity claim or a change in project direction.

## Frozen and exact revisions

- Branch-freeze `upstream/main`: `0c8234e4cc6a60ff31d809f43a9d5decded86f52`
- Upstream change observed and integrated before commit: `4372baf04d80c358ab9998c7f0bace54049c67cf` (merged #3819; renderer/test only)
- PR creation base: `4372baf04d80c358ab9998c7f0bace54049c67cf`
- Exact head: `35b74be4182f3ff72aabe4fcf4266597e5f9dcaa`
- Exact submission tree: `f0471214c5df24350861efe6241f98447306d666`
- Chinese figure SHA-256: `c32a7aef17b10c57baf3b832544e63114fbeb529c845b3a577b02493a543f6cc`
- English figure SHA-256: `44819028cf05e69fa8ec2c335fb900908dd6331cf90f0162dc3c2594ec7484cd`

## Before → after

**Before:** merged PR #3807 added a reproducible E2 capacity envelope but explicitly left A3/A0 unchanged. A reviewer or operations handoff reader had to combine the canonical protocol, metrics, prose and offline page to understand admission, rejection, protected no-data capacity and stop/recovery reserve.

**After:** the existing `metrics-evidence` figure now reads only the existing canonical protocol and metrics and shows, in Chinese and English:

- C-01: capacity 16 / demand 12 / headroom 4;
- C-02 data line: 10 / 6 / 4;
- C-02 no-data line: 10 / 6 / 4, dedicated and non-borrowable;
- C-04: capacity 9 / demand 6 / headroom 3 after protecting one of four declared units for stop/recovery;
- `4/4` resource gates and `3/3` project gates admitted for E2 rehearsal only;
- `5/5` synthetic mutations rejected fail-closed;
- `not_ready_for_field_operation`, observed participants `0`, and the E3/E4 replacement trigger.

The change replaces one existing raster in place rather than adding a near-duplicate carrier. It updates the bilingual Markdown/report captions, offline visual explanatory text, A3 page 7 and A0 board 3.

## Users and public value

[...]

PR #3743, opened 2026-08-22
fix(review): include rubric feedback on accepted intake

CocoSgt:
## Problem

The auto-review worker already generates and validates an authoritative `pr-comment.md` containing all seven rubric dimensions. `apply_review()` uses that file for request-changes outcomes, but the accept path discards it and posts only the weighted total. Participants therefore lose the structured feedback that the documented `pr_comment_markdown` channel is meant to expose.

## Change

- preserve the existing concise intake boundary/disclaimer;
- append the already generated authoritative PR comment on accepted reviews;
- fail closed if that authoritative comment is unexpectedly empty;
- add a regression asserting the accepted review body contains the seven-dimension section while keeping exact-head merge binding unchanged.

This does not alter model calls, rubric scores, thresholds, local gates, participant scope, or merge behavior.

## Verification

- `/Users/admin/agentswork/worktrees/haidian-fix-2061/.venv/bin/python -m unittest tests.test_auto_review_queue tests.test_ai_review_submission -v` — 32/32 PASS on Python 3.9.6 with project review dependencies;
- `py_compile` — PASS;
- `git diff --check` — PASS.

Fixes #3739

PR #3982, opened 2026-08-24
fix(review): separate future follow-ups from intake blockers

dvd233:
## Summary

- bump the advisory review contract to `0.2.0` and add structured `conditional_followups` with required `blocking_now=false`, `trigger`, and `owner`
- reserve `required_next_actions_zh` / per-dimension repairs for current participant-controlled blockers, while rendering future-stage conditions in a separate non-blocking section
- make the trusted worker supply the current canonical schema to old PR worktrees and invalidate cached reviews from earlier schema versions
- keep the queue fail-closed for malformed conditional records and for every current participant repair

## Root cause and safety boundary

The flat `required_next_actions_zh` list had to carry both current repairs and future-stage conditions. The normalizer and queue therefore converted any condition into `request-changes`, even when the review itself called it non-blocking. This patch adds an explicit machine-readable distinction instead of inferring ownership or timing from keywords.

A current unsupported claim is still a blocker: the participant must correct the claim now. Only work whose stated trigger has not occurred belongs in `conditional_followups`.

The authoritative PR comment now renders current blockers and conditional follow-ups separately. This does not duplicate #3743, which remains the separate accept-path delivery fix for posting the full authoritative comment.

## Verification

- `python -m unittest discover -s tests -p 'test_*review*.py' -v` — 89/89 passed
- focused advisory / queue / prompt modules — 43/43 passed after the final template update
- `py_compile` for the three changed review scripts — passed
- `git diff --check` and schema JSON parse — passed
- worker-shape dry-run against PR #3821 exact head `452b367ff3c19fc5079a38fc60999c86fd084faa` using the trusted `0.2.0` schema — succeeded; reviewed package hash remained `c8eca57bd34497af50bc53db1d625d1a6299d1cef334906024d3311816fe54eb`

A broader sparse-worktree run passed 287/289 non-symlink/non-gallery tests; the two remaining failures were unrelated existing/sparse-fixture checks (cover separator text and generated gallery state without all submissions materialized).

## Rollout note

This code intentionally does not mutate existing PR reviews or labels. Existing `changes-requested` PRs need a maintainer requeue or manual intake decision after this lands; the schema-version check then prevents reuse of their old cached advisory result.

Addresses #3980.

### DIFF
--- a/brief/site-package/schemas/advisory_review.schema.json
+++ b/brief/site-package/schemas/advisory_review.schema.json
@@ -13,12 +13,13 @@
     "rubric_scores",
     "data_gaps_zh",
     "required_next_actions_zh",
+    "conditional_followups",
     "pr_comment_markdown"
   ],
   "additionalProperties": false,
   "properties": {
     "schema_version": {
-      "const": "0.1.0"
+      "const": "0.2.0"
     },
     "submission_dir": {
       "type": "string",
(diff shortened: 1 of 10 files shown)


CocoSgt (approved):
Exact head `d2b0a8b4fcba412e21ca3b0d3e4901fa28422248` reproduces and fixes the fail-closed/fail-close ambiguity without weakening current gates: current participant-controlled repairs remain blocking, while explicit future-triggered items require `blocking_now=false`, a nonempty trigger, and a bounded owner enum. Schema-invalid or ambiguous items still fail closed; the worker invalidates legacy cached reviews by schema version and uses the trusted current-main schema. Independent Python 3.9.6 verification passed 56/56 related review/encoding tests, three-script `py_compile`, schema JSON parsing, and `git diff --check`; required exact-head CI is SUCCESS. This aligns the existing documented repository-intake boundary and does not alter the 60-point floor, mandatory rejection, four local gates, publication semantics, participant path scope, or exact-head checks.

